PR #8, opened 2017-10-05
SIR Comparison

twicki:
Addition of == operators for SIRs and their members


src/dawn/SIR/SIR.cpp
```diff
@@ -174,4 +174,137 @@ std::shared_ptr<sir::StencilCall> sir::StencilCall::clone() const {
...
+      if(GlobalVariableMap->size() == rhs.GlobalVariableMap->size()) {
+        if(GlobalVariableMap->size() > 0) {
+          retval &=
+              pointerMapComparison(*(GlobalVariableMap.get()), *(rhs.GlobalVariableMap.get()));
+        }
+      } else {
+        return false;
+      }
+    } else {
+      return GlobalVariableMap == rhs.GlobalVariableMap;
```
thfabian: Why do we need this? This will simply compare the pointers

afile.txt
```diff
@@ -0,0 +1,2 @@
+afea:x
```
thfabian: Do we need this file?

src/dawn/SIR/SIR.cpp
```diff
@@ -174,4 +174,137 @@ std::shared_ptr<sir::StencilCall> sir::StencilCall::clone() const {
...
+  if(Fields.size() != rhs.Fields.size()) {
+    return false;
+  }
+  retval &= (Name == rhs.Name);
+  retval &= (Attributes == rhs.Attributes);
+  if(StencilDescAst != nullptr && rhs.StencilDescAst != nullptr) {
+    if(StencilDescAst->getRoot() != nullptr && rhs.StencilDescAst->getRoot() != nullptr) {
+      retval &= StencilDescAst->getRoot().get()->equals(rhs.StencilDescAst->getRoot().get());
+    } else {
+      return StencilDescAst->getRoot() == rhs.StencilDescAst->getRoot();
```
thfabian: Remove?

src/dawn/SIR/SIR.cpp
```diff
@@ -174,4 +174,137 @@ std::shared_ptr<sir::StencilCall> sir::StencilCall::clone() const {
...
+  return false;
+}
+bool SIR::operator!=(const SIR& rhs) const { return !(*this == rhs); }
+
+bool sir::Stencil::operator==(const sir::Stencil& rhs) const {
+  bool retval = true;
+  if(Fields.size() != rhs.Fields.size()) {
+    return false;
+  }
+  retval &= (Name == rhs.Name);
```
thfabian: The problem with using
```
retval &= ...
```
is that even *if* it is false, you don't exit the function but do the whole comparison. Maybe it is more efficient to just always return false as soon as you see a mismatch and if you pass all the checks simply return `true`

src/dawn/SIR/SIR.cpp
```diff
@@ -174,4 +174,137 @@ std::shared_ptr<sir::StencilCall> sir::StencilCall::clone() const {
...
+                           sharedpointer_comparison<sir::Interval>);
+    }
+
+    if(Asts.size() > 0) {
+      auto astcomparison = [](std::shared_ptr<dawn::AST> comparate1,
+                              std::shared_ptr<dawn::AST> comparate2) {
+        if(comparate1->getRoot() != nullptr && comparate2->getRoot() != nullptr) {
+          return comparate1->getRoot().get()->equals(comparate2->getRoot().get());
+        } else {
+          return comparate1->getRoot() == comparate2->getRoot();
```
thfabian: You don't need this? It should be just `return false`

src/dawn/SIR/SIR.cpp
```diff
@@ -174,4 +174,137 @@ std::shared_ptr<sir::StencilCall> sir::StencilCall::clone() const {
...
+                               const sir::GlobalVariableMap& map2) const {
+  if(map1.size() != map2.size()) {
+    return false;
+  } else {
+    for(auto& a : map1) {
+      auto finder = map2.find(a.first);
+      if(finder == map2.end()) {
+        return false;
+      } else if(finder->second == nullptr || a.second == nullptr) {
+        return finder->second == a.second;
```
thfabian: `return false`

src/dawn/SIR/SIR.cpp
```diff
@@ -174,4 +174,137 @@ std::shared_ptr<sir::StencilCall> sir::StencilCall::clone() const {
...
+    }
+    return true;
+  }
+}
+
+bool sir::Value::operator==(const sir::Value& rhs) const {
+  if(getType() != rhs.getType()) {
+    return false;
+  }
+  if(getType() == Value::TypeKind::Boolean) {
```
thfabian: Maybe use a `switch` statement to make it less verbose? ;)

src/dawn/SIR/SIR.h
```diff
@@ -363,8 +384,32 @@ struct SIR {
...
+                            const sir::GlobalVariableMap& map2) const;
 };
 
+////////////// Should not be here.....
+/// @brief Compares the content of two shared pointers
+/// @param[in] shared pointer of type T
+/// @param[in] shared pointer of same type T
+/// @return true if contents of the shared pointers match (operator ==)
+template <typename T>
+bool sharedpointer_comparison(const std::shared_ptr<T>& comparate1,
```
thfabian: I think the name is misleading as it does not *actually* compare the pointer but rather the pointee. Maybe something like `pointeeComparison`? Also do you need to expose this function in the header? I think putting it in `SIR.cpp` would be more appropriate.

test/unit-test/dawn/SIR/TestSIRComparison.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "llvm/ADT/STLExtras.h"
+#include <gtest/gtest.h>
+#include <iostream>
+
+using namespace dawn;
+
+namespace {
+
+TEST(SIRComparisonTest, Dummy) {
+  /// Works all the time, initializing the suite
```
thfabian: In general `//` should be used for comments and `///` if you want to write a Doxygen comment.

test/unit-test/dawn/SIR/TestSIRComparison.cpp
```diff
@@ -0,0 +1,129 @@
...
+  EXPECT_TRUE(false);
+}
+
+TEST(SIRComparisonTest, DoubleReference) {
+  std::unique_ptr<dawn::SIR> test01SIR = llvm::make_unique<dawn::SIR>();
+  std::unique_ptr<dawn::SIR> test02SIR = llvm::make_unique<dawn::SIR>();
+
+  // Genereate a stencil
+  test01SIR->Stencils.resize(1);
+  for(auto& a : test01SIR->Stencils) {
```
thfabian: How about 
```
std::generate(test01SIR->Stencils.begin(), test01SIR->Stencils.end(), []() { return std::make_shared<sir::Stencil>(); });
```
to make it a little more condensed? ;)

test/unit-test/dawn/SIR/TestSIRComparison.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "dawn/SIR/AST.h"
+#include "dawn/SIR/ASTUtil.h"
+#include "dawn/SIR/SIR.h"
+#include "dawn/Support/Casting.h"
+#include "dawn/Support/STLExtras.h"
+#include "llvm/ADT/STLExtras.h"
+#include <gtest/gtest.h>
+#include <iostream>
+
+using namespace dawn;
```
thfabian: you can also add `using namespace sir` to avoid typing it each time.

test/unit-test/dawn/SIR/TestSIRComparison.cpp
```diff
@@ -0,0 +1,129 @@
...
+  test02SIR->Stencils[0]->StencilDescAst =
+      std::make_shared<dawn::AST>(std::make_shared<BlockStmt>(std::vector<std::shared_ptr<Stmt>>{
+          std::make_shared<ExprStmt>(std::make_shared<VarAccessExpr>("foo"))}));
+
+  test01SIR->Stencils[0]->Fields.resize(1);
+  test02SIR->Stencils[0]->Fields.resize(1);
+  test01SIR->Stencils[0]->Fields[0] = std::make_shared<sir::Field>("foo");
+  test02SIR->Stencils[0]->Fields[0] = std::make_shared<sir::Field>("foo");
+
+  test01SIR->Stencils[0]->Attributes.set(dawn::sir::Attr::AttrKind::AK_MergeTemporaries);
```
thfabian: `dawn::` is not necessary here

test/unit-test/dawn/SIR/TestSIRComparison.cpp
```diff
@@ -0,0 +1,129 @@
...
+//  // Generate a global Variable Map
+  test02SIR->GlobalVariableMap = std::make_shared<sir::GlobalVariableMap>();
+  test01SIR->GlobalVariableMap = std::make_shared<sir::GlobalVariableMap>();
+  auto val = std::make_shared<sir::Value>();
+  val.get()->setType(sir::Value::TypeKind::Integer);
+  val.get()->setValue(10);
+  test01SIR->GlobalVariableMap->emplace("TestGlobal", val);
+  test02SIR->GlobalVariableMap->emplace("TestGlobal", val);
+
+  EXPECT_TRUE(*test01SIR.get() == *test02SIR.get());
```
thfabian: Looks good! A small suggestion would be to split it in several tests instead of one über test which tests everything. Maybe just test the stencils (no AST), stencils with AST, stencil functions only, fields, only etc.. It will be easier to spot the error in the comparison as now you get basically no feedback if the test fails. 

Btw, you can maybe overload `operator<<(std::ostream& os, const SIR& sir)` to make GTest output it on failure

src/dawn/SIR/SIR.cpp
twicki: We have equality if both Maps are nullptrs.
First [194] we check if both are not null (in oder to ensure sizes do not segfault) but if either is nullptr, we need both to be null and therefore we do this

src/dawn/SIR/SIR.cpp
twicki: I was thinking about this and am not sure what is better. If we expect SIRs to match most of the time (which I think they do), we do not want early exits to better pipeline. 
Of course we lose performance for mismatches. How certain are you that early exits are worth it?

src/dawn/SIR/SIR.cpp
twicki: Again, same argument as for the GlobalVariableMap, if somehow both StencilDescAsts are initialized with a nullptr, we need this for equality (and the if before makes it s.t. get and equals work properly

src/dawn/SIR/SIR.cpp
twicki: why? two ASTs with empy roots are the same?

src/dawn/SIR/SIR.cpp
twicki: why?

test/unit-test/dawn/SIR/TestSIRComparison.cpp
twicki: is way prettier :)

test/unit-test/dawn/SIR/TestSIRComparison.cpp
twicki: I was thinking about that but figured this should be more of a proof of concept kind of thing to show that equality really compares everything. I think what you suggest should be done for the "real" tests case of generating one SIR form file with GTClang and comparing it to an in-memory creation by hand but is probably unnecessary here

src/dawn/SIR/SIR.cpp
thfabian: Point taken.

src/dawn/SIR/SIR.cpp
thfabian: Alright, makes sense.

src/dawn/SIR/SIR.cpp
thfabian: I wouldn't worry much about piplining at this *high* level (this should be done by the compiler anyway - and actually is done). Early exits will definitely improve performance drastically and will not have a noticeable difference if the SIR are equal. See https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

src/dawn/SIR/SIR.cpp
thfabian: Ok

test/unit-test/dawn/SIR/TestSIRComparison.cpp
thfabian: I'm not sure, because if the test fails it's gonna be really painful to find the error. The tests don't need to be elaborate just split it in multiple tests and *unit* test the individual features (for example just copy paste the `GlobalVariableMap` part in it's own test). The problem with SIR from file is that you have another variable component (the parsing) with potential bugs but yes regression testing can be obviously done with SIRs from file.

PR #16, opened 2017-10-12
Proper output for AST comparison

twicki:


src/dawn/SIR/SIR.cpp
```diff
@@ -50,6 +193,259 @@ static bool pointerMapComparison(const sir::GlobalVariableMap& map1,
...
+  if(!(Attributes == rhs.Attributes)) {
+    output += "[Stencil missmatch] Stencil attibutes do not match\n";
+    output += dawn::format("of Stencil %s and %s", Name, rhs.Name);
+    return std::make_pair(output, false);
+  }
+  if(!StencilDescAst->getRoot().get()->equals(rhs.StencilDescAst->getRoot().get())) {
+    output += "[Stencil missmatch] Stencil ASTs Do not match\n";
+    auto a = compareAst(StencilDescAst, rhs.StencilDescAst);
+    if(a.second)
+      dawn_unreachable("Stencils should not match");
```
thfabian: Better in this case use `DAWN_ASSERT_MSG(a.second, "stencils should not match");`. Unreachables are more a compiler hint in switch statements to indicate impossible cases.

src/dawn/SIR/SIR.cpp
```diff
@@ -50,6 +193,259 @@ static bool pointerMapComparison(const sir::GlobalVariableMap& map1,
...
+std::pair<std::string, bool> sir::Value::comparison(const sir::Value& rhs) const {
+
+  std::string output;
+  auto type = getType();
+  if(type != rhs.getType()) {
+    output += "[Value missmatch] Values are not of the same type\n";
+    return std::make_pair(output, false);
+  }
+  switch(type) {
+  case sir::Value::TypeKind::Boolean:
```
thfabian: I would write a function which does this kind of stuff (to reduce code duplication).
```
template <class T>
std::pair<std::string, bool> isEqualImpl(const sir::Value& a,  const sir::Value& b, const char* name) {
  if(a.getValue<T> != b.getValue<T>) 
    std::make_pair(dawn::format("[Value mismatch] %s values are not equal\n
                                                   "  left = %s\n"
                                                   "  right = %s\n", name, internal::StdToString<T>()(a), internal::StdToString<T>()(b)), false)
  return std::make_pair(std::string(), true); 
}
```

It might be worth to indenting all the extra information about the members. For example:
```
[Value mismatch] Boolean values are not equal
  left  = true
  right = false
```
seems to be more readable than
```
[Value mismatch] Boolean values are not equal
left  = true
right = false
```
especially if there are multiple reports, you can simply spot the brackets (this may also applies to all the other comparisons).

src/dawn/SIR/SIR.h
```diff
@@ -114,6 +115,31 @@ struct Interval {
            LowerOffset == other.LowerOffset && UpperOffset == other.UpperOffset;
   }
   bool operator!=(const Interval& other) const { return !(*this == other); }
+
+  std::pair<std::string, bool> comparison(const Interval& rhs) const {
+    std::string output;
+    if(LowerLevel != rhs.LowerLevel) {
+      output += dawn::format("[Inverval missmatch] Expected LowerLevel %i, received %i", LowerLevel,
```
thfabian: I would be consistent in the formating. So far you employed the strategy 
```
[type of error] description
  expected = ...
  recieved = ...
```

Meaning I would expect an error message like
```
[Interval mismatch] LowerLevels don't match
  left  = X
  right = Y
```

PR #26, opened 2017-11-07
Serialization support for SIR via Protobuf

thfabian:
This is still a WIP.

Notable additions:
 - Extend CMake to smoothly support protobuf
 - Add SIR description in Protobuf (`SIR.proto file`)
 - Add python support for Dawn (`python/dawn/`)
    - Building of in-memory SIR and serialization to JSON (done)
    - Add a direct API to call the Dawn compiler from within python
 - Add serialization/deserialization support in C++ for the SIR (work in progress)
 - Add proper [RPATH](https://cmake.org/Wiki/CMake_RPATH_handling) support in CMake


cmake/AddProtobuf.cmake
```diff
@@ -37,4 +37,6 @@ dawn_export_package(
   INCLUDE_DIRS "${Protobuf_INCLUDE_DIRS}"
   LIBRARIES "protobuf::libprotobuf"
   VERSION "${Protobuf_VERSION}"
+  # https://stackoverflow.com/questions/34474091/protobuf-inline-not-in-headers
```
cosunae: In line 15 of this file, how does cmake ensures that it finds that version of protobuf? protobuf does not include a version file in cmake...

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+
+syntax = "proto3";
+
+package dawn.sir.proto;
+
+// @brief Source information
+//
+// `(-1,-1)` indicates an invalid location.
+// @ingroup sir_proto
+message SourceLocation {
```
cosunae: do we need also a "file"?

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+  int32 Line = 1;   // Source line
+  int32 Column = 2; // Column in the `Line`
+}
+
+// @brief Stencil attributes
+//
+// Stencil attributes can be used to adjust optimization and code generation for individual stencil
+// and stencil_functions.
+// @ingroup sir_proto
+message Attributes {
```
cosunae: as discussed, this should be options to the compiler and not SIR

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+  bool merge_stages = 2;      // Merge the Stages of this stencil
+  bool merge_do_methods = 3;  // Merge the Do-Methods of this stencil
+  bool merge_temporaries = 4; // Merge the temporaries of this stencil
+  bool use_k_caches = 5;      // Use K-Caches
+}
+
+// @brief Description of a field argument of a Stencil or StencilFunction
+// @ingroup sir_proto
+message Field {
+  string name = 1;        // Name of the field
```
cosunae: we will need to code in the SIR somehow the dimensionality

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+  bool is_temporary = 3;  // Is the field a temporary?
+}
+
+// @brief Directional argument of a StencilFunction
+//
+// Note that this message is merely a placeholder as the actual arguments is encoded as an AST node.
+// @ingroup sir_proto
+message Direction {
+  string name = 1;        // Name of the directional argument
+  SourceLocation loc = 2; // Source location
```
cosunae: I dont understand yet how we reference this and the Offset, dont we need an unique ID or something in order to dereference it? Or should later the compiler go to the SourceLocation in order to evaluate a particular argument?

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+    End = 1;   // The last level
+  }
+
+  // Lower level
+  //
+  // A level can be any integer between `[Start, End]` while the `Start` and `End` levels are
+  // treated specially.
+  oneof LowerLevel {
+    SpecialLevel special_lower_level = 1;
+    int32 lower_level = 2;
```
cosunae: Maybe would be good to have a Level message, instead of simply an integer. 
```
message Level {
  string name = 0;
  int32 position = 1;
}
```
Like that an interval would always refer to an already defined level, rather an unsafe integer value. 
Also I think we need a way to identify if the level will be provided at compile time (optionally) or runtime. 
If position would be = -1, then the compiler would need to generate an API where the values of all levels are passed by runtime arguments

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+message BuiltinType {
+  enum TypeID { None = 0; Auto = 1; Boolean = 2; Integer = 3; Float = 4; }
+
+  TypeID type_id = 1; // Type identifier
+}
+
+// @brief Supported builtin types
+// @ingroup sir_proto
+message Dimension {
+  enum Direction { I = 0; J = 1; K = 2; Invalid = 3; }
```
cosunae: even if gtclang does not support this, we need to generalize the dimensionality in the SIR.  Keep the one as a canonical dimension and add a custom
```
message CustomDimension {
  string name = 1;
  int32 index;
}
message Dimension {
  anyof dim { CanonicalDimension ;
     CustomDimension;
  };
```

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+  }
+
+  bool is_const = 3;    // Is the type const qualified?
+  bool is_volatile = 4; // Is the type volatile qualified?
+}
+
+// @brief Declaration of a vertical region
+// @ingroup sir_proto
+message VerticalRegion {
+  enum LoopOrder { Forward = 0; Backward = 1; }
```
cosunae: do we need an "any" as well?

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+  Interval interval = 3;    // Vertical interval
+  LoopOrder loop_order = 4; // Loop order of execution
+}
+
+// @brief Stencil call declaration
+// @ingroup sir_proto
+message StencilCall {
+  SourceLocation loc = 1;       // Source location of the stencil call
+  string callee = 2;            // Name of the called stencil (i.e callee)
+  repeated Field arguments = 3; // Fields passed as arguments during the stencil call
```
cosunae: it does not have to be "Field". Shouldnt this be StencilFunctionArg?

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+// @ingroup sir_proto
+message Dimension {
+  enum Direction { I = 0; J = 1; K = 2; Invalid = 3; }
+
+  Direction dimension = 1; // Dimension identifier
+}
+
+// @brief Supported builtin types
+// @ingroup sir_proto
+message Type {
```
cosunae: I think I argued that we dont need other than builtin types, but I think I changed my opinion. Having possibility to have POD types might be good idea

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,627 @@
...
+//
+// @ingroup sir_proto
+message VarDeclStmt {
+  Type type = 1;       // Type of the variable
+  string name = 2;     // Name of the variable
+  int32 dimension = 3; // Dimension of the array or 0 for variables
+  string op = 4;       // Operation used for initialization
+
+  // List of expression used for array initialization or just 1 element for variable initialization
+  repeated Expr init_list = 5;
```
cosunae: how do we ensure consistency of the SIR? 
In this particular case, we need to ensure that the repetition number equals the dimension. Is it possible to enforce this in protobuf? If not how to we do it, do we need to develop validity scripts? (would be nice not to have to...)

src/dawn-c/Compiler.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ *  This file is distributed under the MIT License (MIT).
+ *  See LICENSE.txt for details.
+ *
+\*===------------------------------------------------------------------------------------------===*/
+
+#ifndef DAWN_C_DAWNCOMPILER_H
+#define DAWN_C_DAWNCOMPILER_H
+
+#ifdef __cplusplus
```
cosunae: empty file?

thfabian: This is a work in progress ;)

cmake/AddProtobuf.cmake
thfabian: Protobuf includes a file `protobuf-config-version.cmake` alongside their `protobuf-config.cmake` which contains a version definitions
```
set(PACKAGE_VERSION "3.4.1")
```
This is in fact the same way we propagate the Dawn version and seems to be the suggested way by CMake

src/dawn/SIR/SIR.proto
thfabian: At moment we are under the assumption that everything is in the same file (thus only the line and col is required). If we ever drop this requirement we may have to store a file index in the `SourceLocation` and keep a map of these file indices to the actual filenames in the SIR.

src/dawn/SIR/SIR.proto
thfabian: Thses objects, `Direction`, `Offset` and `Field`, are merely *placeholders* used in the `Stencil` and `StencilFunction` argument lists. The Stencil only needs to know the names of it's input fields thus a simple `name` member suffices. For stencil functions, the actual data is encoded in the AST via the `StencilFunctionArgExpr` node.

src/dawn/SIR/SIR.proto
thfabian: Agreed, this requires lots of changes to the SIR and I suggest we tackle this problem in a separate PR as this one is getting quite big already.

src/dawn/SIR/SIR.proto
thfabian: The way the MultiStage splitting passes work is that these loop orders serve as the fallback value of the loop order if a parallel loop order is not possible, meaning we *always* assume a parallel loop order and adjust it to the user provided one if necessary.

src/dawn/SIR/SIR.proto
thfabian: It has to be `Field` as this a **Stencil**Call (not to be confused with **StencilFunction**Call). `Stencils` cannot be parametritzed on directions or offsets. I'm not sure if this a feature we should go for as I can foresee this will increase the code complexity by at least an order of magnitude.

src/dawn/SIR/SIR.proto
thfabian: Agreed, I've also double checked this recently and came to a similar conclusion.

src/dawn/SIR/SIR.proto
thfabian: Consistency checking has to be done on the client side. Meaning you either make these checks *before* you write the SIR or directly *after* parsing. I would argue that the consumer of the SIR (for example Dawn) needs to perform the consistency checks. Protobuf itself does not support any checks afaik besides adhering to the provided scheme. In this particular case however it doesn't need to match the dimensionality as one could write
```
doube a[5] = {0};
``` 
which initializes everything to 0.

src/dawn/SIR/SIR.proto
cosunae: alright, got it. We should distinguish the two in the doc

src/dawn/SIR/SIR.proto
cosunae: yes, it does not make sense in dawn, but as a user description of the stencil, I wouldnt know what to generate for a horizontal stencil, having to specify forward or backward is ambiguous. So I would still add the keyword in the SIR, and simply not use it internally by gtclang and dawn

src/dawn/SIR/SIR.proto
cosunae: so what happens if in the VarDeclStmt message one has a `dimension = 3` and a `init_lint = {0,0,0,0}`

src/dawn/SIR/SIR.proto
thfabian: As discussed I will integrate it.

src/dawn/SIR/SIR.proto
thfabian: This is illformed C++ code and will be rejected by clang in gtclang. I think this should be checked by the one who builds the SIR i.e the frontend DSL.

src/dawn/SIR/SIRSerializer.h
```diff
@@ -23,20 +23,30 @@ struct SIR;
...
+  /// @brief Serialize the SIR as a JSON formatted string to `file`
   ///
-  /// @param file   File to serialize
-  /// @param sir    SIR to serialize
   /// @throws std::excetpion    Failed to open `file`
   static void serialize(const std::string& file, const SIR* sir);
+
+  /// @brief Serialize the SIR to a JSON string
+  /// @returns JSON formatted strong of `sir`
+  static std::string serializeToString(const SIR* sir);
```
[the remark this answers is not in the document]
thfabian: Not sure if necessary. Usually retuning by value from a funtion is subject to RVO (Return Value Optimization) and even garanteed in this case in C++17. If we want to return a pointer I would suggest a `unique_ptr` here.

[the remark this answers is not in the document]
cosunae: sorry, I was not clear. I totally agree with you with the RVO. I meant the argument SIR*

cosunae:
There is something kind of broken here in the serialization of the offsets, 
they are always 0. See here:
https://github.com/MeteoSwiss-APN/gtclang/pull/15/files#diff-c47058ae20a30d0a193869414298ad15R92

python/dawn/sir.py
```diff
@@ -0,0 +1,594 @@
...
+    """ Create an ReturnStmt
+
+    :param cond_part:   Condition part.
+    :param then_part:   Then part.
+    :param else_part:   Else part.
+    """
+    stmt = IfStmt()
+    stmt.expr.CopyFrom(makeStmt(cond_part))
+    stmt.expr.CopyFrom(makeStmt(then_part))
+    if cond_part:
```
twicki: should this not be
if else_part:

python/dawn/sir.py
```diff
@@ -0,0 +1,594 @@
...
+    return t
+
+
+def makeAST(root: StmtType) -> AST:
+    """ Create an AST
+
+    :param root:    Root node of the AST (needs to be of type BlockStmt)
+    """
+    ast = AST()
+    if isinstance(root, BlockStmt) or isinstance(root, Stmt):
```
twicki: Why is any kind of Statement acceptable?

python/dawn/sir.py
```diff
@@ -0,0 +1,594 @@
...
+    """
+    msg = message_type()
+    try:
+        json_format.Parse(text, msg)
+    except json_format.ParseError as e:
+        raise ParseError(str(e))
+    return msg
+
+
+def makeType(builtin_type_or_name, is_const: bool = False, is_volatile: bool = False):
```
twicki: What is the convention for function annotations here? Are all the function-annotated functions exposed in an interface to C?

src/dawn/SIR/CMakeLists.txt
```diff
@@ -12,25 +12,51 @@
...
+          SIRSerializer.h
+          SIRSerializer.cpp
           Statement.h
+          ${sir_proto_cpp_files}
   OBJECT
 )
+
+# Genreate python proto files
+dawn_protobuf_generate(
+  OUT_FILES sir_proto_python_files
```
twicki: Where do we get this from?

src/dawn/SIR/SIR.h
```diff
@@ -149,10 +149,7 @@ struct StencilFunctionArg {
   ArgumentKind Kind;  ///< Type of argument
   SourceLocation Loc; ///< Source location
 
-  ///@brief comparison of Name and Kind (omitting location)
   bool operator==(const StencilFunctionArg& rhs) const;
-
-  ///@brief comparison of Name and Kind (omitting location) with output if the comparison fails
   CompareResult comparison(const sir::StencilFunctionArg& rhs) const;
```
twicki: Why do we remove this?

src/dawn/SIR/SIR.h
```diff
@@ -149,10 +149,7 @@ struct StencilFunctionArg {
   ArgumentKind Kind;  ///< Type of argument
   SourceLocation Loc; ///< Source location
 
-  ///@brief comparison of Name and Kind (omitting location)
   bool operator==(const StencilFunctionArg& rhs) const;
-
-  ///@brief comparison of Name and Kind (omitting location) with output if the comparison fails
```
twicki: why do we remove this?

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,608 @@
...
+}
+
+// @brief Assignment operator
+//
+// Synopsis
+// @code{.cpp}
+//  left = right
+// @endcode
+// @ingroup sir_proto
+message AssignmentExpr {
```
twicki: Assignment should support an operator as is currently in the c++ code to be able to capture assign and operation operators

src/dawn/SIR/SIR.proto
```diff
@@ -0,0 +1,608 @@
...
+  //
+  // The `argumentMap` stores an index to the argument list of the stencil function with -1
+  // indicating this argument is unused. The `argumentOffset` holds the parsed offsets of the
+  // direction (or offset).
+  //
+  // @b Example:
+  // Consider the following example which implements an average stencil function :
+  //
+  // @code{.cpp}
+  // foo(storage in, direction dir) {
```
twicki: call the function avg if we refer to it as avg later on.

And if we do so, should we have the ()/2 there for consistency?

src/dawn/SIR/SIRSerializer.h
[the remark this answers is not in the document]
thfabian: Yes I figured latter that you meant this. The reason we use a raw pointer here is that is is usuable with value semantics 
```
SIR sir = ...
serializeToString(&sir);
```
as well as with smart pointers
```
std::shared_ptr<SIR> sir = ...
serializeToString(sir.get());
```

python/dawn/sir.py
thfabian: Correct!

python/dawn/sir.py
thfabian: Technically we only accept `BlockStmt` but we also need to accept `Stmt` which is simply a wrapper around the `BlockStmt`. I strengthened the requirement to check which kind of statement `Stmt` wraps.

python/dawn/sir.py
thfabian: Functions annotations are just for static type checking by linter tools and IDEs (it is completely optional in Python)

src/dawn/SIR/SIR.proto
thfabian: Fixed

src/dawn/SIR/SIR.proto
thfabian: Done

src/dawn/SIR/CMakeLists.txt
thfabian: This is generated by `protoc` compiler from `SIR.proto`

thfabian:
Implemented comments in https://github.com/thfabian/dawn/pull/2

PR #27, opened 2017-11-09
Qt project

cosunae:
# Technical Description

Some files, like the integration tests .cpp of gtclang are not included in Qt projects since they are not part of any cmake target. 
This creates functionality to add cpp files in a custom target (that does nothing) with the sole purpose that qt creator can add them to the project

This goes together with Meteoswiss-APN/gtclang#13

### DIFF
--- a/cmake/modules/DawnAddCustomDummyTarget.cmake
+++ b/cmake/modules/DawnAddCustomDummyTarget.cmake
@@ -1,0 +1,48 @@
+##===------------------------------------------------------------------------------*- CMake -*-===##
+##                          _                      
+##                         | |                     
+##                       __| | __ ___      ___ ___  
+##                      / _` |/ _` \ \ /\ / / '_  | 
+##                     | (_| | (_| |\ V  V /| | | |
+##                      \__,_|\__,_| \_/\_/ |_| |_| - Compiler Toolchain
+##
+##
+##  This file is distributed under the MIT License (MIT). 
+##  See LICENSE.txt for details.
+##
+##===------------------------------------------------------------------------------------------===##
+
+include(DawnIncludeGuard)
+dawn_include_guard()
+
+include(CMakeParseArguments)
+
+## gtclang_add_integrationtest
+## ---------------------------
+##
+## Creates a dummy target with list of C++ files contained in the list of provided directories, with 
+##   the purpose that qt creator can load those files into the project, which does not happen for those
+##   files that are not included in at least one target.
+##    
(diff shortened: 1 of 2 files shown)


cmake/modules/DawnAddCustomDummyTarget.cmake
```diff
@@ -0,0 +1,48 @@
...
+##   the purpose that qt creator can load those files into the project, which does not happen for those
+##   files that are not included in at least one target.
+##    
+##    NAME:STRING=<>             - Name of the dummy custom target
+##    DIRECTORIES:STRING=<>      - List of directories that contain the source files
+##
+function(dawn_add_custom_dummy_target)
+  # Parse arguments
+  set(options)
+  set(one_value_args GTCLANG NAME)
```
thfabian: This should be `set(one_value_args NAME)`

cmake/modules/DawnAddCustomDummyTarget.cmake
```diff
@@ -0,0 +1,48 @@
...
+  set(multi_value_args DIRECTORIES GRIDTOOLS_FLAGS)
```
thfabian: This should be `set(multi_value_args DIRECTORIES )`

cmake/modules/DawnAddCustomDummyTarget.cmake
```diff
@@ -0,0 +1,48 @@
...
+  cmake_parse_arguments(ARG "${options}" "${one_value_args}" "${multi_value_args}" ${ARGN})
+
+  set(all_files "")
+  foreach(dir ${ARG_DIRECTORIES})
+
+    file(GLOB test_sources "${dir}/*.cpp" )
+    file(GLOB test_headers "${dir}/*.hpp" )
```
thfabian: maybe also glob headers which end with `.h`?

cmake/modules/DawnAddCustomDummyTarget.cmake
cosunae: done

cmake/modules/DawnAddCustomDummyTarget.cmake
cosunae: done

cmake/modules/DawnAddCustomDummyTarget.cmake
cosunae: done

cosunae:
@thfabian I implemented your changes, are you fine? can I merge?

thfabian:
lgtm

PR #28, opened 2017-11-13
Fix clang format

thfabian:
## Technical Description

This PR makes the `find_package` module of clang-format more robust by allowing to search for specific versions. In addition, we fix the version of clang-format to 3.8 when using it.


cosunae:
I tested in my system that has a default clang-format-3.7 and it works. So I am merging it

PR #29, opened 2017-11-14
add a find gridtools

cosunae:
# Technical Description: 

add a find griddtools cmake module


thfabian:
Looks good to me if you add the documentation. 

This is the template I usually use for header-only libraries https://github.com/thfabian/sequoia/blob/master/sequoia-cmake/cmake/Findglm.cmake

cosunae:
ok, since we decided to install GT we dont need this anymore... I will close this one

PR #32, opened 2017-11-16
Issue #28, Caching of Non-Temporaries in the IJ-Plane

twicki:
## Brief

Addition of PassSetNonTempCaches that decides, which non-temporary variables we should write to a scarchpad based on the metric implemented in  PassDataLocalityMetric . It adds filler / flush stages if need be at the beginning and the end of any given multistage the pass is run on. 
In order to run with this pass enabled, the flag `-fcache-non-temp-fields` can be set.

Small adaptation to the PassDataLocalityMetric to support the updated needs.

### DIFF
--- a/src/dawn/Optimizer/MultiStage.h
+++ b/src/dawn/Optimizer/MultiStage.h
@@ -118,6 +118,9 @@ public:
   /// @brief Get the caches
   std::unordered_map<int, Cache>& getCaches() { return caches_; }
   const std::unordered_map<int, Cache>& getCaches() const { return caches_; }
+
+  /// @brief Rename all the occurances in the multi-stage
+  void renameAllOccurrences(int oldAccessID, int newAccessID);
 };
 
 } // namespace dawn
--- a/src/dawn/Optimizer/PassDataLocalityMetric.cpp
+++ b/src/dawn/Optimizer/PassDataLocalityMetric.cpp
@@ -54,23 +54,20 @@ class ReadWriteCounter : public ASTVisitorForwarding {
   /// Current stencil function call
   std::stack<StencilFunctionInstantiation*> stencilFunCalls_;
 
-  /// Hardware configuration
-  const HardwareConfig& config_;
 
   /// Map of ID's to their respective number of reads / writes
   std::unordered_map<int, ReadWriteAccumulator> individualReadWrites_;
 
 public:
-  ReadWriteCounter(StencilInstantiation* instantiation, const MultiStage& multiStage,
-                   const HardwareConfig& config)
+  ReadWriteCounter(StencilInstantiation* instantiation, const MultiStage& multiStage)
       : instantiation_(instantiation), numReads_(0), numWrites_(0), multiStage_(multiStage),
-        fields_(multiStage_.getFields()), config_(config) {}
+        fields_(multiStage_.getFields()) {}
 
   std::size_t getNumReads() const { return numReads_; }
   std::size_t getNumWrites() const { return numWrites_; }
 
   void updateTextureCache(int AccessID, int kOffset) {
-    if(textureCache_.size() < config_.TexCacheMaxFields)
+    if(textureCache_.size() < instantiation_->getOptimizerContext()->getHardwareConfiguration().TexCacheMaxFields)
       textureCache_.emplace_front(AccessID, kOffset);
     else {
       auto it = std::find_if(
@@ -128,23 +125,13 @@ public:
     if(cache.getCacheType() == Cache::K) {
       if(cache.getCacheIOPolicy() == Cache::fill ||
          cache.getCacheIOPolicy() == Cache::fill_and_flush) {
-//        std::cout << "   Fill: " << getNameFromAccessID(AccessID) << std::endl;
         numReads_++;
-        auto iter = individualReadWrites_.find(AccessID);
-        if(iter != individualReadWrites_.end())
-          (*iter).second.numReads++;
-        else
-          individualReadWrites_.emplace(AccessID, ReadWriteAccumulator{1, 0});
+        individualReadWrites_[AccessID].numReads++;
       }
       if(cache.getCacheIOPolicy() == Cache::flush ||
          cache.getCacheIOPolicy() == Cache::fill_and_flush) {
-//        std::cout << "   Flush: " << getNameFromAccessID(AccessID) << std::endl;
         numWrites_++;
-        auto iter = individualReadWrites_.find(AccessID);
-        if(iter != individualReadWrites_.end())
-          (*iter).second.numWrites++;
-        else
-          individualReadWrites_.emplace(AccessID, ReadWriteAccumulator{0, 1});
+        individualReadWrites_[AccessID].numWrites++;
       }
     }
     kCacheLoaded_.insert(AccessID);
@@ -155,13 +142,8 @@ public:
 
     // Is field stored in cache?
     if(!multiStage_.getCaches().count(AccessID)) {
-//      std::cout << "   Write: " << getNameFromAccessID(AccessID) << std::endl;
       numWrites_++;
-      auto iter = individualReadWrites_.find(AccessID);
-      if(iter != individualReadWrites_.end())
-        (*iter).second.numWrites++;
-      else
-        individualReadWrites_.emplace(AccessID, ReadWriteAccumulator{0, 1});
+      individualReadWrites_[AccessID].numWrites++;
 
       // The written value is stored in a register
       register_.insert(AccessID);
@@ -188,13 +170,8 @@ public:
 
         // Check if the field is either cached or stored in the texture cache
         if(!(multiStage_.isCached(AccessID) || isTextureCached(AccessID, kOffset))) {
-//          std::cout << "   Read: " << getNameFromAccessID(AccessID) << std::endl;
           numReads_++;
-          auto iter = individualReadWrites_.find(AccessID);
-          if(iter != individualReadWrites_.end())
-            (*iter).second.numReads++;
-          else
-            individualReadWrites_.emplace(AccessID, ReadWriteAccumulator{1, 0});
+          individualReadWrites_[AccessID].numReads++;
         } else {
           updateKCache(AccessID);
         }
@@ -204,13 +181,8 @@ public:
 
         // Check if the center is stored in a register
         if(!(register_.count(AccessID) && getOffset(fieldExpr) == Array3i{{0, 0, 0}})) {
-//          std::cout << "   Read: " << getNameFromAccessID(AccessID) << std::endl;
           numReads_++;
-          auto iter = individualReadWrites_.find(AccessID);
-          if(iter != individualReadWrites_.end())
-            (*iter).second.numReads++;
-          else
-            individualReadWrites_.emplace(AccessID, ReadWriteAccumulator{1, 0});
+          individualReadWrites_[AccessID].numReads++;
         }
 
       } else {
@@ -257,7 +229,7 @@ public:
 /// every output at most 1 store.
 DAWN_ATTRIBUTE_UNUSED static std::pair<int, int>
 computeReadWriteAccessesLowerBound(StencilInstantiation* instantiation,
-                                   const MultiStage& multiStage, const HardwareConfig& config) {
+                                   const MultiStage& multiStage) {
   std::size_t numReads = 0, numWrites = 0;
   std::unordered_map<int, Field> fields = multiStage.getFields();
 
@@ -305,9 +277,8 @@ computeReadWriteAccessesLowerBound(StencilInstantiation* instantiation,
 /// @brief Approximate the reads and writes individually for each ID
 std::unordered_map<int, ReadWriteAccumulator>
 computeReadWriteAccessesMetricPerAccessID(StencilInstantiation* instantiation,
-                                          const MultiStage& multiStage,
-                                          const HardwareConfig& config) {
-  ReadWriteCounter readWriteCounter(instantiation, multiStage, config);
+                                          const MultiStage& multiStage) {
+  ReadWriteCounter readWriteCounter(instantiation, multiStage);
 
   for(const auto& stage : multiStage.getStages())
     for(const auto& DoMethod : stage->getDoMethods())
@@ -318,12 +289,10 @@ computeReadWriteAccessesMetricPerAccessID(StencilInstantiation* instantiation,
   return readWriteCounter.getIndividualReadWrites();
 }
 
-
 /// @brief Approximate the reads and writes accoding to our data locality metric
 std::pair<int, int> computeReadWriteAccessesMetric(StencilInstantiation* instantiation,
-                                                   const MultiStage& multiStage,
-                                                   const HardwareConfig& config) {
-  ReadWriteCounter readWriteCounter(instantiation, multiStage, config);
+                                                   const MultiStage& multiStage) {
+  ReadWriteCounter readWriteCounter(instantiation, multiStage);
 
   for(const auto& stage : multiStage.getStages())
     for(const auto& DoMethod : stage->getDoMethods())
@@ -359,10 +328,10 @@ bool PassDataLocalityMetric::run(StencilInstantiation* stencilInstantiation) {
         std::cout << "  MultiStage " << multiStageIdx << ":\n";
 
         auto readAndWrite =
-            computeReadWriteAccessesMetric(stencilInstantiation, multiStage, config_);
+            computeReadWriteAccessesMetric(stencilInstantiation, multiStage);
 
         //        auto readAndWrite =
-        //            computeReadWriteAccessesLowerBound(stencilInstantiation, multiStage, config_);
+        //            computeReadWriteAccessesLowerBound(stencilInstantiation, multiStage);
 
         std::size_t numReads = readAndWrite.first, numWrites = readAndWrite.second;
 
--- a/src/dawn/Optimizer/PassDataLocalityMetric.cpp
+++ b/src/dawn/Optimizer/PassDataLocalityMetric.cpp
@@ -54,7 +54,6 @@ class ReadWriteCounter : public ASTVisitorForwarding {
   /// Current stencil function call
   std::stack<StencilFunctionInstantiation*> stencilFunCalls_;
 
-
   /// Map of ID's to their respective number of reads / writes
   std::unordered_map<int, ReadWriteAccumulator> individualReadWrites_;
 
@@ -67,7 +66,8 @@ public:
   std::size_t getNumWrites() const { return numWrites_; }
 
   void updateTextureCache(int AccessID, int kOffset) {
-    if(textureCache_.size() < instantiation_->getOptimizerContext()->getHardwareConfiguration().TexCacheMaxFields)
+    if(textureCache_.size() <
+       instantiation_->getOptimizerContext()->getHardwareConfiguration().TexCacheMaxFields)
       textureCache_.emplace_front(AccessID, kOffset);
     else {
       auto it = std::find_if(
@@ -327,8 +327,7 @@ bool PassDataLocalityMetric::run(StencilInstantiation* stencilInstantiation) {
 
         std::cout << "  MultiStage " << multiStageIdx << ":\n";
 
-        auto readAndWrite =
-            computeReadWriteAccessesMetric(stencilInstantiation, multiStage);
+        auto readAndWrite = computeReadWriteAccessesMetric(stencilInstantiation, multiStage);
 
         //        auto readAndWrite =
         //            computeReadWriteAccessesLowerBound(stencilInstantiation, multiStage);
--- a/src/dawn/Optimizer/PassDataLocalityMetric.cpp
+++ b/src/dawn/Optimizer/PassDataLocalityMetric.cpp
@@ -58,7 +58,7 @@ class ReadWriteCounter : public ASTVisitorForwarding {
   const HardwareConfig& config_;
 
   /// Map of ID's to their respective number of reads / writes
-  std::unordered_map<int, ReadWriteIDspecific> individualReadWrites_;
+  std::unordered_map<int, ReadWriteAccumulator> individualReadWrites_;
 
 public:
   ReadWriteCounter(StencilInstantiation* instantiation, const MultiStage& multiStage,
@@ -134,7 +134,7 @@ public:
         if(iter != individualReadWrites_.end())
           (*iter).second.numReads++;
         else
-          individualReadWrites_.emplace(AccessID, ReadWriteIDspecific(1, 0));
+          individualReadWrites_.emplace(AccessID, ReadWriteAccumulator(1, 0));
       }
       if(cache.getCacheIOPolicy() == Cache::flush ||
          cache.getCacheIOPolicy() == Cache::fill_and_flush) {
@@ -144,7 +144,7 @@ public:
         if(iter != individualReadWrites_.end())
           (*iter).second.numWrites++;
         else
-          individualReadWrites_.emplace(AccessID, ReadWriteIDspecific(0, 1));
+          individualReadWrites_.emplace(AccessID, ReadWriteAccumulator(0, 1));
       }
     }
     kCacheLoaded_.insert(AccessID);
@@ -161,7 +161,7 @@ public:
       if(iter != individualReadWrites_.end())
         (*iter).second.numWrites++;
       else
-        individualReadWrites_.emplace(AccessID, ReadWriteIDspecific(0, 1));
+        individualReadWrites_.emplace(AccessID, ReadWriteAccumulator(0, 1));
 
       // The written value is stored in a register
       register_.insert(AccessID);
@@ -194,7 +194,7 @@ public:
           if(iter != individualReadWrites_.end())
             (*iter).second.numReads++;
           else
-            individualReadWrites_.emplace(AccessID, ReadWriteIDspecific(1, 0));
+            individualReadWrites_.emplace(AccessID, ReadWriteAccumulator(1, 0));
         } else {
           updateKCache(AccessID);
         }
@@ -210,7 +210,7 @@ public:
           if(iter != individualReadWrites_.end())
             (*iter).second.numReads++;
           else
-            individualReadWrites_.emplace(AccessID, ReadWriteIDspecific(1, 0));
+            individualReadWrites_.emplace(AccessID, ReadWriteAccumulator(1, 0));
         }
 
       } else {
@@ -246,7 +246,7 @@ public:
   }
 
   void visit(const std::shared_ptr<FieldAccessExpr>& expr) override { processReadAccess(expr); }
-  std::unordered_map<int, ReadWriteIDspecific> getIndividualReadWrites() const {
+  const std::unordered_map<int, ReadWriteAccumulator>& getIndividualReadWrites() const {
     return individualReadWrites_;
   }
 };
@@ -303,9 +303,10 @@ computeReadWriteAccessesLowerBound(StencilInstantiation* instantiation,
 } // anonymous namespace
 
 /// @brief Approximate the reads and writes individually for each ID
-std::unordered_map<int, ReadWriteIDspecific> computeReadWriteAccessesMetricIndividually(StencilInstantiation* instantiation,
-                                                   const MultiStage& multiStage,
-                                                   const HardwareConfig& config) {
+std::unordered_map<int, ReadWriteAccumulator>
+computeReadWriteAccessesMetricPerAccessID(StencilInstantiation* instantiation,
+                                          const MultiStage& multiStage,
+                                          const HardwareConfig& config) {
   ReadWriteCounter readWriteCounter(instantiation, multiStage, config);
 
   for(const auto& stage : multiStage.getStages())
--- a/src/dawn/Optimizer/PassDataLocalityMetric.cpp
+++ b/src/dawn/Optimizer/PassDataLocalityMetric.cpp
@@ -57,6 +57,9 @@ class ReadWriteCounter : public ASTVisitorForwarding {
   /// Hardware configuration
   const HardwareConfig& config_;
 
+  /// Map of ID's to their respective number of reads / writes
+  std::unordered_map<int, ReadWriteIDspecific> individualReadWrites_;
+
 public:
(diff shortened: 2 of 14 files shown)


src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+  accessIDaccessMetric(int id, int gain) : accessID(id), dataLocalityGain(gain) {}
+  int accessID;
+  int dataLocalityGain;
+  bool operator<(const accessIDaccessMetric& rhs) {
+    return dataLocalityGain < rhs.dataLocalityGain;
+  }
+};
+
+struct nameToImprovementMetric {
+  nameToImprovementMetric(std::string name_, Cache cache_, int metric)
```
thfabian: This is not needed in C++11 see http://en.cppreference.com/w/cpp/language/aggregate_initialization

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+        }
+      }
+      return true;
+  }
+
+  MultiStage* multiStagePrt_;
+  HardwareConfig config_;
+  StencilInstantiation* instantiation_;
+
+  std::unordered_map<int, int> accessIDtoDataLocality_;
```
thfabian: accessID**T**oDataLocality_

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+  std::unordered_map<int, int> oldIDtoNewID;
```
thfabian: which IDs do you refer to here? Maybe call it `oldAccessIDToNewAccessID_`

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+            stmtAccessesPair->getCallerAccesses()->getWriteAccesses().find(id);
+        if(wirteAccessIterator != stmtAccessesPair->getCallerAccesses()->getWriteAccesses().end()) {
+          return false;
+        }
+      }
+    }
+    return false;
+  }
+
+  bool checkReadOnlyAccess(int id) {
```
thfabian: It would be better to stick to the convention of calling it `AccessID` instead of just `id`.

The formatting is really broken in this function

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+  bool checkReadBeforeWrite(int id) {
+    for(auto stageItGlob = multiStagePrt_->getStages().begin();
+        stageItGlob != multiStagePrt_->getStages().end(); ++stageItGlob) {
+      DoMethod& doMethod = (**stageItGlob).getSingleDoMethod();
+      for(int stmtIndex = 0; stmtIndex < doMethod.getStatementAccessesPairs().size(); ++stmtIndex) {
+        const std::shared_ptr<StatementAccessesPair>& stmtAccessesPair =
+            doMethod.getStatementAccessesPairs()[stmtIndex];
+
+        // Find first if this statement has a read
+        auto readAccessIterator = stmtAccessesPair->getCallerAccesses()->getReadAccesses().find(id);
```
thfabian: It think for sake of readability it is worth storing the read accesses in a local variable to avoid the long chains `stmtAccessesPair->getCallerAccesses()->getReadAccesses()` (you access it several times)

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+    assignmentStage->update();
+
+    return assignmentStage;
+  }
+
+  ///@brief add the assignment operator of two unique id's to a given domethod
+  void addAssignmentToDoMethod(std::unique_ptr<DoMethod>& domethod, int assignmentID,
+                               int assigneeID) {
+    // Create the StatementAccessPair of the assignment with the new and old variables
+    std::shared_ptr<FieldAccessExpr> fa_assignee =
```
thfabian: It is very hard to follow these few lines...  It would strongly recommend using `auto` in this case as the syntactic noise from `shared_ptr` here is crazy.

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+  /// We always fill to the extent of the given multistage and we only add one stage to fill all the
+  /// variables to reduce synchronisation overhead
+  void addFillerStages() {
+    for(int i = 0; i < std::min((int)sortedAccesses_.size(), config_.SMemMaxFields); ++i) {
+      int oldID = sortedAccesses_[i].accessID;
+
+      // Create new temporary field and register in the instantiation
+      int newID = instantiation_->nextUID();
+
+      instantiation_->setAccessIDNamePairOfField(newID, dawn::format("__tmp_cache_%i", i), true);
```
thfabian: `instantiation_->setAccessIDNamePairOfField(newID, "__tmp_cache_" + std::to_string(i), true);` [this an order of magnitude faster ;)]

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  /// @brief for each chached variable, check if we need a filler-stage (if we find a
+  /// read-before-write) and add it if necessary, renames all occurances and sets the cache for the
+  /// temporary variable newly created
+  ///
+  /// We always fill to the extent of the given multistage and we only add one stage to fill all the
+  /// variables to reduce synchronisation overhead
+  void addFillerStages() {
+    for(int i = 0; i < std::min((int)sortedAccesses_.size(), config_.SMemMaxFields); ++i) {
```
thfabian: It would be better to precompute the upper bound of the loop as a general rule of thumb.

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+      multiStagePrt_->renameAllOccurrences(oldID, newID);
+
+      oldIDtoNewID.emplace(oldID, newID);
+      Cache& cache = multiStagePrt_->setCache(Cache::IJ, Cache::local, newID);
+      originalNameToCache_.emplace_back(instantiation_->getOriginalNameFromAccessID(oldID), cache,
+                                        accessIDtoDataLocality_.find(oldID)->second);
+    }
+
+    // Create the cache-filler stage
+    std::vector<int> oldIDs;
```
thfabian: It would advocate to use `oldAccessIDs`

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+  void process() {
+    computeOptimalFields();
+    addFillerStages();
+    addFlushStages();
+  }
+
+  std::vector<nameToImprovementMetric> getoriginalNameToCache() { return originalNameToCache_; }
+
+private:
+  /// @brief use the data locality metric to rank the fields in a stencil based on how much they
```
thfabian: For user of the documentation it would be good to stick to one convention and always start the brief statement of doxygen with a capital letter

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+    std::vector<int> oldIDs;
+    std::vector<int> newIDs;
+    for(auto& idPair : oldIDtoNewID) {
+      if(!checkReadOnlyAccess(idPair.second)) {
+        oldIDs.push_back(idPair.first);
+        newIDs.push_back(idPair.second);
+      }
+    }
+
+    if(oldIDs.size() > 0) {
```
thfabian: How about `if(!oldIDs.empty())`. I think this a relays the intent slightly better.

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+#include "dawn/Optimizer/StencilInstantiation.h"
+#include "dawn/SIR/ASTExpr.h"
+#include "dawn/Support/Unreachable.h"
+#include <iostream>
+#include <set>
+#include <vector>
+
+namespace dawn {
+
+struct accessIDaccessMetric {
```
thfabian: Classes should start with a capital letter.

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+  /// @brief entry method for the pass: processes a given multistage and applies all changes
+  /// required
+  void process() {
+    computeOptimalFields();
+    addFillerStages();
+    addFlushStages();
+  }
+
+  std::vector<nameToImprovementMetric> getoriginalNameToCache() { return originalNameToCache_; }
```
thfabian: You should almost never return by **value** a member of a class. It is almost always beneficial to return a constant reference and you call this function as an in (line 298) a loop which could be a disaster (In this case you only waste one copy as containers in range based for-loops are guaranteed to be evaluated only once - https://stackoverflow.com/questions/10821756/is-the-ranged-based-for-loop-beneficial-to-performance)

Also the name scheme is not correct it should be get**O**riginalNameToCache

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+struct nameToImprovementMetric {
+  nameToImprovementMetric(std::string name_, Cache cache_, int metric)
+      : name(name_), cache(cache_), dataLocalityImprovement(metric) {}
+  std::string name;
+  Cache cache;
+  int dataLocalityImprovement;
+};
+
+/// @brief The GlobalFieldCacher class handles the caching for a given Multistage
+/// @param[in, out]  msprt   Pointer to the multistage to handle
```
thfabian: This comment should be part of the constructor?

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  std::vector<nameToImprovementMetric> getoriginalNameToCache() { return originalNameToCache_; }
+
+private:
+  /// @brief use the data locality metric to rank the fields in a stencil based on how much they
+  /// would benefit from caching
+  void computeOptimalFields() {
+    auto dataLocality =
+        computeReadWriteAccessesMetricIndividually(instantiation_, *multiStagePrt_, config_);
```
thfabian: How about naming this function `computeReadWriteAccessesMetricPerAccessID`? It is not immediately clear what individually means in this context (does it refer to the metric or does it mean individually for each field).

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+    for(const auto& stagePtr : multiStagePrt_->getStages()) {
+      for(const Field& field : stagePtr->getFields()) {
+        auto findField = accessIDtoDataLocality_.find(field.AccessID);
```
thfabian: `findField` is not a very good name here, I don't think this is even a field? Given that this is actually an `int` I would assume from the name of the map this is the value of the metric, no?

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+        // We already checked this field
+        if(findField != accessIDtoDataLocality_.end())
+          continue;
+
+        // Only ij-cache fields that are not horizontally pointwise and are vertically pointwise
+        if(!(field.Extent.isVerticalPointwise()) || (field.Extent.isHorizontalPointwise()))
```
thfabian: parenthesis are not needed here operator **!** takes precedence over **||**

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+          continue;
+
+        // This is caching non-temporary fields
+        if(instantiation_->isTemporaryField(field.AccessID))
+          continue;
+
+        int cachedreadwrites = dataLocality.find(field.AccessID)->second.total();
```
thfabian: cached**R**ead**And**Writes or something like that

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+        // We need one more write to flush the cache back to the variable
+        if(!checkReadOnlyAccess(field.AccessID))
+          cachedreadwrites--;
+
+        if(cachedreadwrites > 0)
+          accessIDtoDataLocality_.emplace(field.AccessID, cachedreadwrites);
+      }
+    }
+
+    for(auto& performanceMesure : accessIDtoDataLocality_) {
```
thfabian: What does `performanceMesure` mean here? Isn't this just a pair of AccessIDs to the value of the metric? How about naming this `AccessIDMetricPair`?

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,325 @@
...
+          if(wirteAccessIterator != stmtAccessesPair->getCallerAccesses()->getWriteAccesses().end()) {
+            return false;
+          }
+        }
+      }
+      return true;
+  }
+
+  MultiStage* multiStagePrt_;
+  HardwareConfig config_;
```
thfabian: This is never initialized and just relies on default initialization. We should open an issue that the `HardwareConfig` gets properly passed around in the Passes. One way would be to extend the `Pass::run()` to take an additional argument i.e the `HardwareConfig`

src/dawn/Optimizer/PassSetNonTempCaches.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef DAWN_OPTIMIZER_PASSNONTEMPCACHES_H
+#define DAWN_OPTIMIZER_PASSNONTEMPCACHES_H
+
+#include "dawn/Optimizer/Pass.h"
+
+namespace dawn {
+
+/// @brief This Pass optimizes data locality on a per-multistage basis
+///
+/// For all the non-temporary fields that we access in any given multistage, we check how high the
```
thfabian: For all the non-temporary, vertically-pointwise fields ... Then you can skip the next sentence

src/dawn/Optimizer/PassSetNonTempCaches.h
```diff
@@ -0,0 +1,47 @@
...
+/// performance gain of using a local ij-cache woud be. This is only beneficial if the fields are
+/// horizontally non-pointwise but vertically pointwise.
+/// If is is beneficial, we cache the fieldswith the most performance gain up to HardwareConfig's
```
thfabian: fields with

src/dawn/Optimizer/PassSetNonTempCaches.h
```diff
@@ -0,0 +1,47 @@
...
+/// Maximum number of fields concurrently in shared memory (defined in PassDataLocalityMetric.h)
+///
+/// The caching invokes a renaming of all the occurances of any given field we cache and adds
+/// potentially filler stages and definitly flush stages to the multistage thus changing the ISIR
```
thfabian: definitely

src/dawn/Optimizer/MultiStage.h
```diff
@@ -118,6 +118,9 @@ class MultiStage {
   /// @brief Get the caches
   std::unordered_map<int, Cache>& getCaches() { return caches_; }
   const std::unordered_map<int, Cache>& getCaches() const { return caches_; }
+
+  /// @brief Rename all the occurances in the multi-stage
+  void renameAllOccurrences(int oldAccessID, int newAccessID);
```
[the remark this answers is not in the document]
twicki: done

src/dawn/Optimizer/PassDataLocalityMetric.cpp
```diff
@@ -125,13 +128,23 @@ class ReadWriteCounter : public ASTVisitorForwarding {
...
+          (*iter).second.numReads++;
+        else
+          individualReadWrites_.emplace(AccessID, ReadWriteIDspecific(1, 0));
       }
       if(cache.getCacheIOPolicy() == Cache::flush ||
          cache.getCacheIOPolicy() == Cache::fill_and_flush) {
-        std::cout << "   Flush: " << getNameFromAccessID(AccessID) << std::endl;
+//        std::cout << "   Flush: " << getNameFromAccessID(AccessID) << std::endl;
         numWrites_++;
+        auto iter = individualReadWrites_.find(AccessID);
```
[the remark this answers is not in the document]
twicki: the problem is that you actually need the iterator and do something different with it dependent on the situation. 
We could possibly write a function 
`std::pair<bool, typename T::iterator> accessIDisInContainer(int id, T container)`
but I don't think this really improves readabilty?

src/dawn/Optimizer/PassDataLocalityMetric.cpp
```diff
@@ -218,23 +246,11 @@ class ReadWriteCounter : public ASTVisitorForwarding {
   }
 
   void visit(const std::shared_ptr<FieldAccessExpr>& expr) override { processReadAccess(expr); }
+  std::unordered_map<int, ReadWriteIDspecific> getIndividualReadWrites() const {
```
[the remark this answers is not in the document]
twicki: oversight, fixed

src/dawn/Optimizer/PassDataLocalityMetric.h
```diff
@@ -27,6 +28,16 @@ struct HardwareConfig {
   int TexCacheMaxFields = 3;
 };
 
+struct ReadWriteIDspecific {
```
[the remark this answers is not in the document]
twicki: If you want to use emplace to emplace a key-value pair of int and struct in a map, is an explicit ctor not needed?

src/dawn/Optimizer/PassDataLocalityMetric.h
```diff
@@ -41,6 +52,14 @@ class PassDataLocalityMetric : public Pass {
   HardwareConfig config_;
 };
 
+std::pair<int, int> computeReadWriteAccessesMetric(StencilInstantiation* instantiation,
```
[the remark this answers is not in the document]
twicki: This was here previously... I think it returns total number of Reads and total number of Writes

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: sure

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: again, emplace does not work if we do not have this. Is there a better way?

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: fixed

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: fixed

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: sounds good.

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: We can call it `iter` since it actually is an iterator in the map that checks if we already checked this field

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: I was thinking of having them there for readability but can see the argument for removing them

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: we can do this

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: sure

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: after changing it I am also not super pleased with how auto looks. In either case this is a noisy part of the code as we manually create a stage in memory. Whatever works better

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: it actually is only 2 consecutive calls. Is it really worth it for that?

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: Should it be moved around though? Should this not in the end be a singleton that is in a place where it is globally accessible for all passes?

src/dawn/Optimizer/PassDataLocalityMetric.h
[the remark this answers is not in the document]
thfabian: Possibly true, but you can always write `map.emplace(ReadWriteAccumulator{X,Y})` which is equally efficient.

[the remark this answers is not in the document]
twicki: then you would have to remove the default values. A trade off I am willing to do

src/dawn/Optimizer/PassDataLocalityMetric.cpp
```diff
@@ -125,13 +128,23 @@ class ReadWriteCounter : public ASTVisitorForwarding {
     if(cache.getCacheType() == Cache::K) {
       if(cache.getCacheIOPolicy() == Cache::fill ||
          cache.getCacheIOPolicy() == Cache::fill_and_flush) {
-        std::cout << "   Fill: " << getNameFromAccessID(AccessID) << std::endl;
+//        std::cout << "   Fill: " << getNameFromAccessID(AccessID) << std::endl;
```
cosunae: this is dead code, can we remove it?

twicki: Well, the whole pass could be refactored. These were just cout's in the middle of the computation. We can go one of two ways: Either we let the pass return a struct that holds all the info that is cout'ed and store it somewere (probably in the same place where the hardware-config singleton should be) or we make it s.t the pass only keeps the data locally and add a proper if verbose-stmt to it.

For now I did not want to kill all the output that the pass had for good so I left in in this stage until the decision is made

src/dawn/Optimizer/PassDataLocalityMetric.cpp
```diff
@@ -125,13 +128,23 @@ class ReadWriteCounter : public ASTVisitorForwarding {
...
         numReads_++;
+        auto iter = individualReadWrites_.find(AccessID);
```
cosunae: just wondering, would the one line
```
individualReadWrites_[AccessID].numReads++;
```
do the whole magic? individualReadWrites_[AccessID] would default construct an element if key is not found...

src/dawn/Optimizer/PassDataLocalityMetric.cpp
```diff
@@ -125,13 +128,23 @@ class ReadWriteCounter : public ASTVisitorForwarding {
...
+        if(iter != individualReadWrites_.end())
+          (*iter).second.numReads++;
+        else
+          individualReadWrites_.emplace(AccessID, ReadWriteIDspecific(1, 0));
       }
       if(cache.getCacheIOPolicy() == Cache::flush ||
          cache.getCacheIOPolicy() == Cache::fill_and_flush) {
-        std::cout << "   Flush: " << getNameFromAccessID(AccessID) << std::endl;
+//        std::cout << "   Flush: " << getNameFromAccessID(AccessID) << std::endl;
```
cosunae: remove?

src/dawn/Optimizer/PassDataLocalityMetric.cpp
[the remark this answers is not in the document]
cosunae: I made a same comment above on the reads...

src/dawn/Optimizer/PassDataLocalityMetric.cpp
```diff
@@ -142,8 +155,13 @@ class ReadWriteCounter : public ASTVisitorForwarding {
 
     // Is field stored in cache?
     if(!multiStage_.getCaches().count(AccessID)) {
-      std::cout << "   Write: " << getNameFromAccessID(AccessID) << std::endl;
+//      std::cout << "   Write: " << getNameFromAccessID(AccessID) << std::endl;
```
cosunae: remove?

src/dawn/Optimizer/PassDataLocalityMetric.cpp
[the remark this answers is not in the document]
cosunae: is not fixed yet, isnt? Actually the method is const, so shouldnt it return  a const & ?

src/dawn/Optimizer/PassDataLocalityMetric.cpp
twicki: This is actually really smart :+1:

src/dawn/Optimizer/PassDataLocalityMetric.cpp
```diff
@@ -286,6 +302,38 @@ computeReadWriteAccessesLowerBound(StencilInstantiation* instantiation,
...
+
+
+/// @brief Approximate the reads and writes accoding to our data locality metric
+std::pair<int, int> computeReadWriteAccessesMetric(StencilInstantiation* instantiation,
+                                                   const MultiStage& multiStage,
+                                                   const HardwareConfig& config) {
+  ReadWriteCounter readWriteCounter(instantiation, multiStage, config);
+
+  for(const auto& stage : multiStage.getStages())
+    for(const auto& DoMethod : stage->getDoMethods())
```
cosunae: question about convention: do we have a convention for if, ...scopes? I would always  {} (as opposed to single line) since it can avoid mistakes when we add more statements. But if we have a convention for single line without {} already is fine.... as far as we are consistent

src/dawn/Optimizer/PassDataLocalityMetric.h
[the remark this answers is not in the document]
cosunae: sorry, I didnt get it. Why not using here ReadWriteAccumulator better than std::pair? like that we know what each member of the pair means, so is more readable.
Also why do we pass StencilInstantiation* instead of reference of shared_ptr ?
And my final comment here, can we add documentation (doxygen) what the function does?

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+struct NameToImprovementMetric {
+  std::string name;
+  Cache cache;
+  int dataLocalityImprovement;
+};
+
+class GlobalFieldCacher {
+public:
+  /// @brief The GlobalFieldCacher class handles the caching for a given Multistage
```
cosunae: shouldnt this documentation be above the class declaration (rather than ctr)?

src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+  /// @brief entry method for the pass: processes a given multistage and applies all changes
+  /// required
+  void process() {
+    computeOptimalFields();
+    addFillerStages();
+    addFlushStages();
+  }
+
+  const std::vector<NameToImprovementMetric>& getOriginalNameToCache() {
```
cosunae: should the method be const?

src/dawn/Optimizer/PassDataLocalityMetric.cpp
thfabian: I never use the {} for a single line (I think it is also stated by the LLVM coding style). Imho it just pollutes the code with unnecessary braces.

src/dawn/Optimizer/PassSetNonTempCaches.cpp
thfabian: The `@brief` statement yes, but not the `@param`. I think I wasn't entirely clear in my review on this.

src/dawn/Optimizer/PassDataLocalityMetric.h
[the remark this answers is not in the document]
thfabian: I don't get it, why would you need to remove the default values? 
```
ReadWriteAccumulator foo{}; // <-- calls default constructor
ReadWriteAccumulator bar{5, 5}; // <-- performs aggregate initialization
```

src/dawn/Optimizer/PassSetNonTempCaches.cpp
thfabian: Singletons in library is usually a bad design, I would argue you have to pass it around. You can also simply attach it to the `StencilInstantiation` i.e ISIR

src/dawn/Optimizer/PassDataLocalityMetric.h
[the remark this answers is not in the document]
twicki: you cannot write 

```
struct foo{
int i = 1;
int j = 2;
};

int main(){
 foo{2,3};
}
```

It was possible to work around that by carlos' fix, so this is not necessary anymore but just for the sake of completeness

src/dawn/Optimizer/PassSetNonTempCaches.cpp
twicki: ISIR seems like an ok place to put it. This allows for a very simple pass to set it properly. I see two problems with this though: 
- The first one is that we add an additional field to the ISIR which is not really an ISIR feature?
- This is not something that is not really dependent on a Stencil/Stencil-Function but is the same thing on a global scope. 

I'd suggest to add it to the optimizer.

Thoughts?

PR #33, opened 2017-11-23
Bundle cmake

cosunae:
Create a bundle cmake to resolve all the dependencies with cmake

# Description:

The CMakeLists.txt still assumes that we pass all the dependencies with <project>_DIR paths to cmake. 
In addition,every project contains a bundle/CMakeLists.txt (that is an additional cmake project) that will describe the dependencies of the current project and resolve them (i.e. clone, and compile) them one by one. 
All the common cmake functionality was moved to a new msbuild repository. 

Compiling only requires 
```
cd bundle
mkdir bundle
cmake .. -DCMAKE_CXX_COMPILER=`which c++`
```

Additional one can pass any option to cmake, that will be propagated to all dependent projects


bundle/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+    CMAKE_ARGS 
+        ${dawn_cmake_args}
+)
+
+if(NOT(dawn_binary_dir))
+  message(FATAL_ERROR "dawn binary dir was not defined by External_dawn")
+endif()
+
+add_test(NAME dawn-tests 
+  COMMAND  make test -C ${dawn_binary_dir}
```
thfabian: This is not portable as you 1. cannot assume `make` is the actual build tool (people tend to also use `ninja`) and you can also not assume it is in `PATH` (if you really need make use the CMake built-in variable `${CMAKE_MAKE_PROGRAM}` and check it actually is make).

A more portable way of doing this is:
```
add_test(NAME dawn-tests 
  COMMAND  ${CMAKE_COMMAND} --build ${dawn_binary_dir} --target test
)
```
which uses CMake to invoke the make program.

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+)
+
+file(WRITE ${CMAKE_BINARY_DIR}/Conf.bundle "conf")
```
thfabian: Maybe prefix the files with the package name `dawn-conf.bundle` or something like this

bundle/cmake/DawnCloneRepository.cmake
```diff
@@ -0,0 +1,93 @@
...
+##
+##
+##  This file is distributed under the MIT License (MIT). 
+##  See LICENSE.txt for details.
+##
+##===------------------------------------------------------------------------------------------===##
+
+include(CMakeParseArguments)
+
+# dawn_clone_repository
```
thfabian: Again (move to MSBuild)

cmake/thirdparty/dawnAddprotobuf.cmake
```diff
@@ -0,0 +1,40 @@
+##===------------------------------------------------------------------------------*- CMake -*-===##
```
thfabian: Inconsistent file name, should be **D**awnAddProtobuf

Or maybe just don't prefix them at all as I think later on you want to all these `AddX` files in repository and reuse them for all projects (in this case you need to clean them up to get rid of all gtclang and dawn references)

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+project(dawn C CXX)
+enable_testing()
+cmake_minimum_required(VERSION 3.3)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/../cmake")
+
+include(DawnCloneRepository)
+
+set(MB_GIT_URL "https:///github.com/cosunae/msbuild.git"
```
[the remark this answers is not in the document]
cosunae: what do you mean? this is the url of msbuild, not DAWN, isnt?

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+include(msbuildAddTargetCleanAll)
+include(msbuildCheckInSourceBuild)
+include(msbuildGetArchitectureInfo)
+include(msbuildGetCompilerInfo)
+include(msbuildGetPlatformInfo)
+include(msbuildMakeStringPair)
+include(msbuildReportResult)
+
+# Include all options (including all Dawn and GTClang options)
+#include(GTClangAllOptions)
```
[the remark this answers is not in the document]
cosunae: done

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+  msbuild_make_string_pair(${FIRST} ": ${SECOND}" 20 out)
+  list(APPEND config_info ${out})
+endmacro()
+make_config_string("Platform" ${MSBUILD_PLATFORM_STRING})
+make_config_string("Architecture" ${MSBUILD_ARCHITECTURE_STRING})
+make_config_string("Compiler" ${MSBUILD_COMPILER_STRING})
+make_config_string("Build type" ${CMAKE_BUILD_TYPE})
+msbuild_report_result("Configuration summary" ${config_info})
+
+include(DawnAddDependency)
```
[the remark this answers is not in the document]
cosunae: I actually removed the add_dependency, is not doing anything but including the dawnAddProtobuf. 
They are making a call to msbuild_find_package with a specific version, tag, url ... so isnt this kind of client dependent? 
If I move it to msbuild we would force all packages download a specific version

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+
+dawn_add_dependency(NAME "protobuf")
+
+
+set(dawn_cmake_args -DProtobuf_DIR=${Protobuf_DIR})
```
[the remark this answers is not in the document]
cosunae: In thirdparty/dawnAddProtobuf. 
The idea or convention was to put external package dependencies like protobuf and Boost hidden in these AddXXX packages, since we wont change the version that often, while others like dawn, gtclang, gridtools we can accomodate the branch we compile against more often

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+    BINARY_DIR dawn_binary_dir
+  DEPENDS "protobuf"
+  ADDITIONAL
+    SOURCE_DIR "${CMAKE_SOURCE_DIR}/../"
+    MSBUILD_ROOT "${DAWN_MSBUILD_SOURCE_DIR}"
+    CMAKE_ARGS 
+        ${dawn_cmake_args}
+)
+
+if(NOT(dawn_binary_dir))
```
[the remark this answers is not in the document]
cosunae: sorry this was not clear (and also not yet documented).  dawn_binary_dir is not setby the user, but by the find_package. It is the varable that indicates where is the build directory, so that we can call the ctest from there. The idea is that the FORWARD_VARS will tell find_package that , in case we need to build the package, they need to be passed to External_pkg.cmake and set properly. There are checks at several levels that all FORWARD_VARS are set. The idea was also to not used a convention, like expect that the External_pkg.cmake will put that in certain fixed global variable, since we are then connecting too much functions via global variables. Instead we pass output variables that need to be set.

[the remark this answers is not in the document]
cosunae: what does `if(NOT(var))` ? does it check that the variable is defined and the string is not empty? 
That is what I intended, since is better than checking that is defined (but could be empty string)... but I am not sure

bundle/CMakeLists.txt
cosunae: true, good point. I ll change that

bundle/CMakeLists.txt
cosunae: done

bundle/cmake/DawnCloneRepository.cmake
cosunae: this one is actually a copy of the one in msbuild. The problem is that if msbuild is not found, the bundle will first clone it, and then it can use all the msbuild infrastructure. But how do we clone msbuild if we dont have msbuild?

cmake/thirdparty/dawnAddprotobuf.cmake
cosunae: I did rename it. 
But are u sure we should move it to msbuild? it is just a convenient indirection that specifies the version of protobuf, but other than that the main function is find_package. Different packages might require different versions of protobuf

CMakeLists.txt
```diff
@@ -31,47 +31,53 @@ cmake_minimum_required(VERSION 3.3)
...
 
+if(NOT(DEFINED MSBUILD_ROOT))
+    message(FATAL_ERROR "Specifying a MSBUILD_ROOT is required")
+endif()
+list(APPEND CMAKE_MODULE_PATH "${MSBUILD_ROOT}/cmake")
+
 include(DawnCMakeInit)
 set(DAWN_ROOT ${CMAKE_CURRENT_LIST_DIR})
+
 dawn_cmake_init()
```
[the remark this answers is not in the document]
cosunae: not sure, I think we still need this, isnt? we want to be able to run this CMakeLists.txt w/o the bundle

PR #36, opened 2017-12-04
MCHBUILD_ROOT as a cmake cache variable

cosunae:
Description
=============
add mchbuild_root as a cache variable to CMakeLists.txt so that qt creator is not resetting it all the time


CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ project(dawn C CXX)
 cmake_minimum_required(VERSION 3.3)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+set(MCHBUILD_ROOT CACHE PATH "path to mchbuild package")
+
 if(NOT(DEFINED MCHBUILD_ROOT))
```
twicki: should be not empty instead of not defined

cosunae: fixed

cosunae:
launch jenkins

PR #42, opened 2018-01-04
Compute extents

cosunae:
Technical Description
===================

compute the extent of each stage associated to the redundant computations. 
Normally this extent is computed by GridTools. However GridTools provides an alternative API in which the user can specify the extent. This bypasses a complex metaprogram that computes the extents in GT and as a result speedsup the compilation of GT generated examples. 
Additionally the extent is required in order to determine the loop bounds of a naive and standalone CUDA/OMP code generation. 

Changes:
* in DawnCompiler provide a runOptimizer(SIR) interface so that we can obtain an optimizer (an a StencilInstantiation). This is used by in memory unittests, in order to check the state of StencilInstantiation. 
* Add a PassComputeStageExtents that computes the extent of each stage of the StencilInstantiation
* Add in memory tests for optimizer passes test/unit-test/Optimizer/Passes
* Add tests for PassComputeStageExtents that loads a reference SIR, compute the extents and checks against expected


src/dawn/Optimizer/Extents.cpp
```diff
@@ -37,6 +37,11 @@ void Extents::merge(const Extents& other) {
     extents_[i].merge(other.extents_[i]);
 }
 
+void Extents::expand(const Extents& other) {
+  for(std::size_t i = 0; i < extents_.size(); ++i)
```
twicki: should and equal size check be here to avoid unexpected results?

test/unit-test/dawn/Optimizer/Passes/TestEnvironment.h
```diff
@@ -0,0 +1,26 @@
+/*===----------------------------------------------------------------------------------*- C -*-===*\
```
twicki: This functionality is in the wrong spot here. 

The goal of the test-directory is a mirror of the src with each file representing the file it tests. This is test-required functionality and should therefore be in src/dawn/Unittest/
I know that there are some not-so-nice ways where this has not been done in the codebase but I think we should stick to it nevertheless.

test/unit-test/dawn/Optimizer/Passes/TestPassComputeStageExtents.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include <fstream>
+#include <gtest/gtest.h>
+#include <streambuf>
+
+using namespace dawn;
+
+namespace {
+
+TEST(ComputeStageExtents, test_stencil_01) {
+  std::string filename = TestEnvironment::path_ + "/compute_extent_test_stencil_01.sir";
```
twicki: should we denote the serialized stencil somewhere in the comments to make it, such that tests are better understood on their own?

test/unit-test/dawn/Optimizer/Passes/TestPassComputeStageExtents.cpp
```diff
@@ -0,0 +1,213 @@
...
+  std::ifstream file(filename);
+  DAWN_ASSERT_MSG((file.good()), std::string("File " + filename + " does not exists").c_str());
+
+  std::string jsonstr((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
+
+  std::shared_ptr<SIR> sir = SIRSerializer::deserializeFromString(jsonstr, SIRSerializer::SK_Json);
```
twicki: should we pack all of this up to the tests into a function? Seems this is repeated in every test?
or should we use the `TEST_P` env?

twicki:
can we merge the travis fix back in here to verify?

src/dawn/Optimizer/Extents.cpp
cosunae: yes, indeed, good catch, today is hardcoded to 3, but in the future we want to generalize it. 
I added asserts in several places

test/unit-test/dawn/Optimizer/Passes/TestEnvironment.h
cosunae: alright, got it. 
This TestEnvironmnent is a small class that contains information specific to how we run the tests in this subproject "Passes", since they require a std::string path. 
I can move it to `src/dawn/Unittest` but it will be used specifically only by unittest in the Passes. Do you still think we should move it there? How do we handle if we need multiple of this TestEnvironmnet for different subprojects?

test/unit-test/dawn/Optimizer/Passes/TestPassComputeStageExtents.cpp
cosunae: yes, exactly. They are generated from there. The idea is that everytime SIR changes we would need to regenerate them with gtclang. We would need a simple python script that generate all the SIR references used in testing of dawn from gtclang sources

test/unit-test/dawn/Optimizer/Passes/TestPassComputeStageExtents.cpp
cosunae: good point, I did created a fixture, can you have a look?

cosunae:
launch jenkins

test/unit-test/dawn/Optimizer/Passes/TestEnvironment.h
twicki: ah, got it. Yeah, you're right. We can definitely keep it as is.

twicki:
Changes look really good!

PR #43, opened 2018-01-04
Support of (inlined) nested function calls for naive code generation 

cosunae:
NOT FOR REVIEW, SINCE IT INTEGRATES #42. 
JUST FOR TESTING THE PR

Technical Description
==================
Support for one line nested function calls like
```fn(gn(i+1, st1, hn(st2, st3) ) ) ```

Changes: 
* Added a CodeGenProperties to hold all data structures that are used for stencils for code generation, instead of the, so far used, custom maps and vectors. 
* Add the extents of the stage as loop bounds of the naive backend


cosunae:
launch jenkins

cosunae:
launch jenkins

PR #44, opened 2018-01-10
fix sir serialization of globals

cosunae:
Technical description
====================
`gtclang -write-sir` 

crashes for the example in 
https://github.com/MeteoSwiss-APN/gtclang/blob/master/test/integration-test/CodeGen/globals.cpp
This is a fix.


twicki:
launch jenkins

PR #45, opened 2018-01-11
Fix travis

cosunae:
Technical Description
==================
remove hardcoded path to python3. 
Now it compiles in travis, however setuptools is not found, so protobuf python is not yet installed


.travis.yml
```diff
@@ -38,6 +38,8 @@ env:
     - CACHE_DIR=${TRAVIS_BUILD_DIR}/deps/install
     - CMAKE_VERSION=3.8.1
     - PROTOBUF_VERSION=3.4.1
+install:
```
twicki: why do we need this?

PR #47, opened 2018-01-11
Support globals in stencil functions

cosunae:
Technical Description
==================

added support for using globals inside stencil functions. 

Additional change: in StencilInstantiation and StencilFunctionInstantiation add interfaces to add/retrieve entries from the maps of Expr/Stmt to AccessID. Currently the whole map is returning as a ref, and then from the caller scope it is modified. In the changes proposed here we have specific API for adding or retrieving entries to the maps.

### DIFF
--- a/src/dawn/Optimizer/StencilInstantiation.h
+++ b/src/dawn/Optimizer/StencilInstantiation.h
@@ -258,6 +258,12 @@ public:
   /// @brief Get the `AccessID` of the Expr (VarAccess or FieldAccess)
   int getAccessIDFromExpr(const std::shared_ptr<Expr>& expr) const;
 
+  /// @brief Set the `AccessID` of the Expr (VarAccess or FieldAccess)
+  void setAccessIDOfExpr(const std::shared_ptr<Expr>& expr, const int accessID);
+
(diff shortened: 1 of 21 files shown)


src/dawn/Optimizer/StencilInstantiation.h
```diff
@@ -275,7 +290,8 @@ class StencilInstantiation : NonCopyable {
 
   /// @brief Remove the stencil function given by `expr`
   ///
-  /// If `callerStencilFunctionInstantiation` is not NULL (i.e the stencil function is called within
+  /// If `callerStencilFunctionInstantiation` is not NULL (i.e the stencil function is called
```
twicki: this should fit...

src/dawn/Optimizer/StencilInstantiation.h
```diff
@@ -284,7 +300,8 @@ class StencilInstantiation : NonCopyable {
 
   /// @brief Register a new stencil function
   ///
-  /// If `curStencilFunctionInstantiation` is not NULL, the stencil function is treated as a nested
+  /// If `curStencilFunctionInstantiation` is not NULL, the stencil function is treated as a
+  /// nested
```
twicki: again, I think your format file is off..

PR #49, opened 2018-01-22
Added Functionality of Boundary Conditions

twicki:
## Technical Description

Implementation of boundary conditions.
Now in the stencil description, one can specify boundary conditions:
```
stencil_function zero{
  storage a;
  Do {
    a = 0;
  }
};

stencil Test{
  storage a;

  boundary_condition(zero(), a);

  Do{
    .....
  }
};
```

These boundary conditions are ignored when they are not used but if stencils are split (for whatever reason) and fields need to be updated between those, halo updates and boundary conditions are automatically inserted (if a specification is available, otherwise, a compile-time error is thrown)


src/dawn/CodeGen/ASTCodeGenGTClangStencilDesc.cpp
```diff
@@ -62,12 +87,103 @@ void ASTCodeGenGTClangStencilDesc::visit(const std::shared_ptr<VerticalRegionDec
...
+            }
+            break;
+          }
+        }
+      }
+    }
+    // This Field already had BC's applied. We remove all those stencils from the compuattion
+    else {
+      std::vector<int> stencilIDsToVisit(StencilIDsVisited_);
+      for(int traveresedID : StencilBCsApplied_.find(fieldname)->second) {
```
cosunae: how about replacing this with 
```
copy_if(StencilIDsVisited_.begin(), StencilIDsVisited_.end(), stencilIDsToVisit.begin(), [&](int stencilID){
auto listStencilIDs = StencilBCsApplied_.find(fieldname);
return std::find(stencilID, listStencilIDs) != listStencilIDs.end();}
```

src/dawn/CodeGen/ASTCodeGenGTClangStencilDesc.cpp
```diff
@@ -62,12 +87,103 @@ void ASTCodeGenGTClangStencilDesc::visit(const std::shared_ptr<VerticalRegionDec
...
 void ASTCodeGenGTClangStencilDesc::visit(const std::shared_ptr<BoundaryConditionDeclStmt>& stmt) {
-  DAWN_ASSERT_MSG(0, "BoundaryConditionDeclStmt not yet implemented");
+
+  auto calculateHaloExtents = [&](std::string fieldname) {
+
+    Extents fullExtent;
+    // Did we already apply a BoundaryCondition for this field?
+    // This is the first time we apply a BC to this field, we traverse all stencils that were
+    // applied before
+    if(StencilBCsApplied_.count(fieldname) == 0) {
```
cosunae: I think the if and the else are pretty similar, so I would try to merge the algorithm, by first extracting the
`stencilIDsToVisit` which in the case of the if is just a copy of `StencilIDsVisited_`, and from this on the algorithm should be the same

src/dawn/CodeGen/ASTCodeGenGTClangStencilDesc.cpp
```diff
@@ -62,12 +87,103 @@ void ASTCodeGenGTClangStencilDesc::visit(const std::shared_ptr<VerticalRegionDec
...
+      for(auto& stencil : instantiation_->getStencils()) {
+        for(const int& ID : StencilIDsVisited_) {
```
cosunae: could we replace the for by an if, to have less nested loops?

src/dawn/CodeGen/ASTCodeGenGTClangStencilDesc.cpp
```diff
@@ -20,6 +20,31 @@
 
 namespace dawn {
 
+static Extents analyzestencilExtents(const std::shared_ptr<Stencil>& s, int ID) {
```
cosunae: Can we have some doc describing the alrgorithm, i.e. what is suppose to return? 
Also the parameters, ID is not clear whether is the ID of a field, stencil, etc... so I would also probably use a longer name

cosunae: Capitalize S: analyzeStencilExtents?

src/dawn/CodeGen/GTClangCodeGen.cpp
```diff
@@ -81,6 +81,69 @@ GTClangCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
                               Interval::makeCodeGenName(stencilFun->getInterval()));
 }
 
+class StencilFunctionReader : public ASTCodeGenGTClangStencilBody {
```
cosunae: missing some documentation

src/dawn/CodeGen/GTClangCodeGen.cpp
```diff
@@ -81,6 +81,69 @@ GTClangCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
...
+private:
+  std::shared_ptr<sir::StencilFunction> function;
+
+public:
+  using Base = ASTCodeGenGTClangStencilBody;
+  StencilFunctionReader(const StencilInstantiation* stencilInstantiation,
+                        const std::unordered_map<Interval, std::string>& intervalToNameMap,
+                        const std::shared_ptr<sir::StencilFunction>& foo)
```
cosunae: I would suggest to not use `foo` in naming conventions to have good readability

src/dawn/CodeGen/GTClangCodeGen.cpp
```diff
@@ -81,6 +81,69 @@ GTClangCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
                               Interval::makeCodeGenName(stencilFun->getInterval()));
 }
 
+class StencilFunctionReader : public ASTCodeGenGTClangStencilBody {
+private:
+  std::shared_ptr<sir::StencilFunction> function;
+
+public:
+  using Base = ASTCodeGenGTClangStencilBody;
+  StencilFunctionReader(const StencilInstantiation* stencilInstantiation,
```
cosunae: Is this related to BC only? Could we have a more descriptive name ?  Also Reader sounds strages when it does `ss << `

src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -165,17 +165,18 @@ std::unique_ptr<TranslationUnit> DawnCompiler::compile(const SIR* SIR, CodeGenKi
...
+  passManager.pushBackPass<PassSetBoundaryCondition>();
   passManager.pushBackPass<PassDataLocalityMetric>();
 
   // Run optimization passes
   for(auto& stencil : optimizer->getStencilInstantiationMap()) {
     StencilInstantiation* instantiation = stencil.second.get();
     DAWN_LOG(INFO) << "Starting Optimization and Analysis passes for `" << instantiation->getName()
                    << "` ...";
-
+    instantiation->dump();
```
cosunae: I guess this should be removed?

src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -165,17 +165,18 @@ std::unique_ptr<TranslationUnit> DawnCompiler::compile(const SIR* SIR, CodeGenKi
...
     if(!passManager.runAllPassesOnStecilInstantiation(instantiation))
       return nullptr;
-
+    instantiation->dump();
```
cosunae: same here?

src/dawn/CodeGen/ASTCodeGenGTClangStencilDesc.cpp
```diff
@@ -62,12 +87,103 @@ void ASTCodeGenGTClangStencilDesc::visit(const std::shared_ptr<VerticalRegionDec
...
   int StencilID = instantiation_->getStencilCallToStencilIDMap().find(stmt)->second;
+  StencilIDsVisited_.push_back(StencilID);
   for(const std::string& stencilName : StencilIDToStencilNameMap_.find(StencilID)->second)
     ss_ << std::string(indent_, ' ') << stencilName << ".get_stencil()->run();\n";
 }
 
 void ASTCodeGenGTClangStencilDesc::visit(const std::shared_ptr<BoundaryConditionDeclStmt>& stmt) {
-  DAWN_ASSERT_MSG(0, "BoundaryConditionDeclStmt not yet implemented");
+
+  auto calculateHaloExtents = [&](std::string fieldname) {
```
cosunae: General question, there is quite some algorithmic content here, that could belong to the compiler. 
Why dont we do all these in the optimizer, in the BC pass and store it in the BoundaryCondition struct of the StencilInstantiation?  Since this is important information that being stored in the ISIR can be used later on by anyone else, including other passes

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  std::unordered_map<int, Extents> dirtyFields;
+  std::unordered_map<int, bcStorage> allBCs;
+
+  // Fetch all the boundary conditions stored in the instantiation
+  for(const auto& bc : stencilInstantiation->getBoundaryConditions()) {
+    auto it = stencilInstantiation->getNameToAccessIDMap().find(bc.first);
+    if(it != stencilInstantiation->getNameToAccessIDMap().end()) {
+      int ID = stencilInstantiation->getAccessIDFromName(bc.first);
+      bcStorage storage;
```
cosunae: bcStorage is copied from bc, so I dont see the need for having the additional  struct `bcStorage`

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,250 @@
...
+      std::cout << " No boundary conditions applied\n";
+    }
+    return true;
+  }
+  // returns the original ID of a variable since we need to make sure that if
+  auto getOriginalID = [&](int ID) {
+    // check if the variable exists to be sure not to assert
+    auto it = stencilInstantiation->getNameToAccessIDMap().find(
+        stencilInstantiation->getOriginalNameFromAccessID(ID));
+    if(it != stencilInstantiation->getNameToAccessIDMap().end()) {
```
cosunae: why do we need this if? Do you have an example where ID is a field but it has not `getOriginalNameFromAccessID`?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  // check if we need to run this pass
+  if(stencilInstantiation->getStencils().size() == 1) {
+    if(stencilInstantiation->getOptimizerContext()->getOptions().ReportBoundaryConditions) {
+      std::cout << "\nPASS: " << getName() << ": " << stencilInstantiation->getName() << " :";
+      std::cout << " No boundary conditions applied\n";
+    }
+    return true;
+  }
+  // returns the original ID of a variable since we need to make sure that if
```
cosunae: we need documentation, and I have the feeling we should move this to StencilInstantation

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,250 @@
...
+            Accesses& acesses = *(stmtAccess->getAccesses());
+            const auto& allReadAccesses = acesses.getReadAccesses();
+            const auto& allWriteAccesses = acesses.getWriteAccesses();
+
+            // ReadAccesses can trigger Halo-Updates and Boundary conditions if the following
+            // criteria are fullfilled:
+            // It is a Field (ID!=-1) and we had a write before from another stencil (is in
+            // dirtyFields)
+            for(const auto& readacccess : allReadAccesses) {
+              int originalID = getOriginalID(readacccess.first);
```
cosunae: early return `if(originalID== -1) continue;` ?

src/dawn/CodeGen/ASTCodeGenGTClangStencilDesc.cpp
twicki: I'm in for syntactic sugar

src/dawn/CodeGen/ASTCodeGenGTClangStencilDesc.cpp
twicki: true, fixed

src/dawn/CodeGen/GTClangCodeGen.cpp
twicki: It is somewhat unique to BC's since it specifically parses the stencil function and rewrites the field-accesses to be the templated parameters. This is not needed anywhere else

As for the name, is StencilFunctionToBoundaryConditionConverter useful? is it too long? What would you suggest?

src/dawn/Compiler/DawnCompiler.cpp
twicki: sure

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: we can move this up one call

src/dawn/CodeGen/GridTools/ASTStencilDesc.cpp
```diff
@@ -65,7 +65,53 @@ void ASTStencilDesc::visit(const std::shared_ptr<StencilCallDeclStmt>& stmt) {
 }
 
 void ASTStencilDesc::visit(const std::shared_ptr<BoundaryConditionDeclStmt>& stmt) {
-  DAWN_ASSERT_MSG(0, "BoundaryConditionDeclStmt not yet implemented");
+  if(instantiation_->getBoundaryConditionToExtentsMap().count(stmt) == 0) {
+    DAWN_ASSERT_MSG(false, "Boundary Condition did not trigger extent-calculation");
+  }
+  Extents extents = instantiation_->getBoundaryConditionToExtentsMap().find(stmt)->second;
```
cosunae: This would need some protection, in case the stmt is not found in the map?
And a getter function in stencil instantiation could help make the syntax cleaner here
`Extent getBoundaryConditionExtent(stmt)`

src/dawn/CodeGen/GridTools/ASTStencilDesc.cpp
```diff
@@ -65,7 +65,53 @@ void ASTStencilDesc::visit(const std::shared_ptr<StencilCallDeclStmt>& stmt) {
...
+      "%s->get_storage_info_ptr()->total_length<2>());\n",
+      haloIMinus, haloIPlus, fieldname, fieldname, fieldname, haloJMinus, haloJPlus, fieldname,
+      fieldname, fieldname, haloKMinus, haloKPlus, fieldname, fieldname, fieldname);
+  std::string makeView = "";
+
+  // Create the views for the fields
+  for(int i = 0; i < stmt->getFields().size(); ++i) {
+    auto fieldName = stmt->getFields()[i]->Name;
+    makeView +=
+        dawn::format("auto %s_view = GT_BACKEND_DECISION_viewmaker(%s);\n", fieldName, fieldName);
```
cosunae: where is this `viewmaker` coming from? is not GT code isnt? 
aside comment, where are creating views from several places, so it could make sense to have a function that in codegen that given the field name and backends, it returns the string to create a view... but you can take it for another PR if you prefer

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -82,6 +82,73 @@ GTCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
                               Interval::makeCodeGenName(stencilFun->getInterval()));
 }
 
+/// @brief The StencilFunctionReader class parses a stencil function that is used as a boundary
+/// condition into it's stringstream. In order to use stencil_functions as boundary conditions, we
+/// need them to be memnbers of the stencil-wrapper class. The goal is to template the function s.t
```
cosunae: typo (good doc though)

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -82,6 +82,73 @@ GTCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
...
+/// every field is a template argument.
+class StencilFunctionReader : public ASTStencilDesc {
```
cosunae: Given the doc, a name like `StencilFunctionAsBCGenerator` would sound more intittuive to me. This is specific visitor for BC, so I think would be good to have a name that reflects it generates BC code

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -82,6 +82,73 @@ GTCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
...
+
+public:
+  using Base = ASTStencilDesc;
+  StencilFunctionReader(const StencilInstantiation* stencilInstantiation,
+                        const std::unordered_map<int, std::vector<std::string>>& stencilIDToStencilNameMap,
+                        const std::shared_ptr<sir::StencilFunction>& functionToAnalyze)
+      : Base(stencilInstantiation, stencilIDToStencilNameMap), function(functionToAnalyze) {}
+
+  void visit(const std::shared_ptr<FieldAccessExpr>& expr) {
+    auto printOffset = [](const Array3i& argumentoffsets) {
```
cosunae: I think we have a similar one
https://github.com/MeteoSwiss-APN/dawn/blob/master/src/dawn/CodeGen/CXXNaive/ASTStencilBody.h#L53

could we merge the two  to avoid redundancy?

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -82,6 +82,73 @@ GTCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
...
+      if(argumentoffsets[2] == 0) {
+        retval += "k";
+      } else {
+        retval += dawn::format("k + %i", argumentoffsets[2]);
+      }
+      return retval;
+    };
+    expr->getName();
+    auto getArgumentIndex = [&](const std::string& name) {
+      for(int i = 0; i < function->Args.size(); ++i) {
```
cosunae:
```size_t pos = std::distance( function->Args.begin(), std::find( function->Args.begin(), function->Args.end(), name) );
DAWN_ASSERT_MSG(pos < function->Args.size());
return pos;
```

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
cosunae: What is the reason for inhering from ASTStencilDesc?
It seems they are completely different visitors, and we dont use any code from ASTStencilDesc? 
I was wondering since also we need to pass `StencilIDToStencilNameMap` to this BC generator, which has no semantic in this context, and therefore we pass to ctr a empty map, since it is not used.

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+      fullExtents.merge(field.Extent);
+      fullExtents.add(stageExtent);
+    }
+  }
+
+  return fullExtents;
+}
+}
+
+class VisitStencilCalls : public ASTVisitorForwarding {
```
cosunae: add a bit a doc to this class

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  void visit(const std::shared_ptr<StencilCallDeclStmt>& stmt) {
+    stencilCallsInOrder_.push_back(stmt);
+    std::cout << "here: " << stmt->getStencilCall()->Callee << std::endl;
+  }
+};
+
+/// @brief Get all field and variable accesses identifier by `AccessID`
+class AddBoundaryConditions : public ASTVisitorForwarding {
```
cosunae: Are the doc and the name of the class related? does not seem like...

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
cosunae: uhmm where? I didnt understand it

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+            // ReadAccesses can trigger Halo-Updates and Boundary conditions if the following
+            // criteria are fullfilled:
+            // It is a Field (ID!=-1) and we had a write before from another stencil (is in
+            // dirtyFields)
+            for(const auto& readacccess : allReadAccesses) {
+              int originalID = getOriginalID(readacccess.first);
+              if(originalID == -1)
+                continue;
+              auto idWithExtents = dirtyFields.find(originalID);
+              if(idWithExtents != dirtyFields.end()) {
```
cosunae: I would suggest to use more the early exit pattern with return here. In general blocks statements which are nested and large like here are hard to read
```
int readAccessID = readacccess.first;
Extent readAccessExtent = readacccess.second;
if(! dirtyFields.count(originalID)) continue;
if( readAcccessExtent.isHorizontalPointwise() ||
                   stencilInstantiation->getCachedVariableSet().count(readAcccessID) )
continue;
```
Also above I tried to avoid `first` and `second` for readability. Best would be to have a proper structure in `readAccesses_ ` of 
 `Accesses`, that is also probably separate work.

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+                // If the access is horizontally pointwise or it is a horizontally cached field,
+                // we do not need to trigger a BC
+                if(!readacccess.second.isHorizontalPointwise() &&
+                   !stencilInstantiation->getCachedVariableSet().count(readacccess.first)) {
+                  auto finder = allBCs.find(originalID);
```
cosunae: also to avoid the nested blocks: 
```
if( ! allBCs.count(originalID) ) DAWN_ASSERT();

// continue work without adding another else block
```

src/dawn/Optimizer/StencilInstantiation.cpp
```diff
@@ -879,7 +879,14 @@ class StencilDescStatementMapper : public ASTVisitor {
...
   void visit(const std::shared_ptr<BoundaryConditionDeclStmt>& stmt) override {
-    DAWN_ASSERT_MSG(0, "BoundaryConditionDeclStmt not yet implemented");
+    BoundaryConditions bc;
+    bc.functor = stmt.get()->getFunctor();
+    // Extracting all the stencil_function arguments except for the field we apply it to
+    for(int i = 1; i < (*stmt).getFields().size(); ++i) {
+      bc.arguments.push_back(stmt.get()->getFields()[i].get()->Name);
+    }
+    // store it with the fieldname to apply it to as the key
+    instantiation_->getBoundaryConditions().emplace(stmt.get()->getFields()[0].get()->Name, bc);
```
cosunae: I know this is a common used pattern we have in the code, but I would avoid it in the future, meaning we have a getter, but in reality we are inserting members,

I would rather have 
```instantiation_->insertBoundaryCondition```
like that is clear when we want to search for when the creation and insertion of BC happens

File: src/dawn/Optimizer/PassSetBoundaryCondition.cpp [hunk 26]
```diff
@@ -0,0 +1,337 @@
...
+                  // Check if a boundary condition for this variable was defined
+                  if(finder != allBCs.end()) {
+                    // Create the Statement to insert the Boundary conditon into the StencilDescAST
+                    std::shared_ptr<BoundaryConditionDeclStmt> boundaryConditionCall =
+                        std::make_shared<BoundaryConditionDeclStmt>(finder->second.functor);
```
cosunae: Here we create the BoundaryConditionDeclStmt from a BoundaryCondition. Just wondering, why not creating it directly in the StencilInstantiation instead of storing BoundaryConditions
https://github.com/MeteoSwiss-APN/dawn/pull/49/files#diff-c86dcdb35edb57df3601bd8e76cc64ccR889
?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+                        stencilInstantiation->getNameFromAccessID(readacccess.first)));
+                    for(const auto& arg : finder->second.arguments) {
+                      boundaryConditionCall->getFields().emplace_back(
+                          std::make_shared<sir::Field>(arg));
+                    }
+
+                    // Calculate the extent and add it to the boundary-condition - Extent map
+                    Extents fullExtents = calculateHaloExtents(
+                        stencilInstantiation->getNameFromAccessID(readacccess.first));
+                    stencilInstantiation->getBoundaryConditionToExtentsMap().emplace(
```
cosunae: similar comment I made before, I would suggest to have a method
```insertBoundaryConditionExtent```

File: src/dawn/Optimizer/PassSetBoundaryCondition.cpp [hunk 28]
```diff
@@ -0,0 +1,337 @@
...
+                        boundaryConditionCall, fullExtents);
+
+                    // check if this stencil is called and get its StencilCallDeclStmt (the one to
+                    // replace)
+                    auto test =
```
cosunae: early exit,
```
if( ! stencilInstantiation->getIDToStencilCallMap().count(stencil.getStencilID()))  DAWN_ASSER();
```

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+                        false,
+                        dawn::format("In stencil %s we need a halo update on field %s but no "
+                                     "boundary condition is set.\nUpdate the stencil (outside the "
+                                     "do-method) with a boundary condition that calls a "
+                                     "stencil_function, e.g \n'boundary_condition(zero(), %s);'\n",
+                                     stencilInstantiation->getName(),
+                                     stencilInstantiation->getOriginalNameFromAccessID(originalID),
+                                     stencilInstantiation->getOriginalNameFromAccessID(originalID))
+                            .c_str());
+                    dirtyFields.erase(originalID);
```
cosunae: I guess this line does not make sense after an assert(false) ?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+                  stencilDirtyFields.emplace(originalID, writeaccess.second);
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+    // Write all the fields set to dirty within this stencil to the global dirty map
+    for(const auto& fieldWithExtends : stencilDirtyFields) {
```
cosunae: ```std::copy(stencilDirtyFields.begin(), stencilDirtyFields.end(), std::back_inserter(dirtyFields));```

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,337 @@
...
+                    dirtyFields.erase(originalID);
+                  }
+                }
+              }
+            }
+            // Any write-access requires a halo update once is is read off-center therefore we set
+            // the fields to modified
+            for(const auto& writeaccess : allWriteAccesses) {
+              int originalID = getOriginalID(writeaccess.first);
+              if(originalID != -1) {
```
cosunae: ``` if(originalID == -1) continue```

cosunae: btw, in the future we should make `-1` a tag, enum or something rather than hardcoded

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
cosunae: should we check that the field is not already in dirtyFields, wouldnt be that an anomaly?

src/dawn/Optimizer/PassSetCaches.cpp
```diff
@@ -182,6 +182,7 @@ bool PassSetCaches::run(StencilInstantiation* instantiation) {
              !field.Extent.isHorizontalPointwise()) {
 
             Cache& cache = multiStagePtr->setCache(Cache::IJ, Cache::local, field.AccessID);
+            instantiation->getCachedVariableSet().emplace(field.AccessID);
```
cosunae: `insertCacheVariable` ? 
Aside comment, this has not to do with BC, was that a bug we had?

src/dawn/Optimizer/PassStencilSplitter.h
```diff
@@ -31,7 +31,7 @@ class PassStencilSplitter : public Pass {
   /// @brief Maximum number of allowed fields per stencil
   ///
   /// This is the threshold for the splitting pass to be invoked.
-  static constexpr int MaxFieldPerStencil = 40;
+  static constexpr int MaxFieldPerStencil = 2;
```
cosunae: should we revert it? 

Should we control it as we discussed with the use of max-halo? and for the particular stencils where you want to unittest BC?

src/dawn/Optimizer/Replacing.h
```diff
@@ -25,6 +26,23 @@ struct Statement;
 class StatementAccessesPair;
 class StencilInstantiation;
 
+/// @brief Get all field and variable accesses identifier by `AccessID`
```
cosunae: documentation seems to be wron

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: I'll double-check where this comes up and document it

src/dawn/Optimizer/PassSetCaches.cpp
twicki: This was not needed before BC's

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -82,6 +82,102 @@ GTCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
...
+  using Base = ASTCodeGenCXX;
+  StencilFunctionAsBCGenerator(
+      const StencilInstantiation* stencilInstantiation,
+      const std::shared_ptr<sir::StencilFunction>& functionToAnalyze)
+      : function(functionToAnalyze), instantiation_(stencilInstantiation) {}
+
+  void visit(const std::shared_ptr<FieldAccessExpr>& expr) {
+    auto printOffset = [](const Array3i& argumentoffsets) {
+      std::string retval = "";
+      if(argumentoffsets[0] == 0) {
```
cosunae: proposal to reduce code 
```std::array<std::string, 3> dims{"i","j","k"};
for(int i=0; i < 3; ++i) {
  retval = dims[i]+(argumentoffset[i]!=0 ? std::to_string(argumentoffsets[i]) : "");
}
```

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+/// @brief This method analyzes the Extents in which a field with a given ID is used within a a
+/// Stencil. This is used to compute how big the halo exchanges need to be.
+/// @param s A shared ptr to the stencil to be analyzed
+/// @param ID the FieldID of the Field to be analized
+/// @return the full extent of the field in the stencil
+static dawn::Extents analyzeStencilExtents(const std::shared_ptr<Stencil>& s, int fieldID) {
+  Extents fullExtents;
+  Stencil& stencil = *s;
+
+  int numStages = stencil.getNumStages();
```
cosunae: there is no need for this declaration,
just use a for range
```
for(auto stage: stencil.getStages())
```

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
cosunae: I still dont understand these nested conditionals. Cant we replace it (from line 120) by:
```
if(stencilInstantiation->isField(ID)) 
         return stencilInstantiation->getAccessIDFromName(
             stencilInstantiation->getOriginalNameFromAccessID(ID));
else
  return (int) FieldType::FT_NotOriginal;
```

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+      return (int) FieldType::FT_NotOriginal;
+    }
+
+  };
+
+  std::unordered_map<int, Extents> dirtyFields;
+  std::unordered_map<int, std::shared_ptr<BoundaryConditionDeclStmt>> allBCs;
+
+  // Fetch all the boundary conditions stored in the instantiation
+  for(const auto& bc : stencilInstantiation->getBoundaryConditions()) {
```
cosunae: all this loop can be achieved with a single line
```
std::transform(stencilInstantiation->getBoundaryConditions().begin(), stencilInstantiation->getBoundaryConditions().end(), allBCs.begin(),
                   [&](std::pair<std::string, std::shared_ptr<BoundaryConditionDeclStmt>> bcPair) -> std::pair<int, std::shared_ptr<BoundaryConditionDeclStmt>> { return stencilInstantiation->getAccessIDFromName(bcPair.first); });
```

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+  for(const std::shared_ptr<Statement>& statement :
+      stencilInstantiation->getStencilDescStatements()) {
+    statement->ASTStmt->accept(findStencilCalls);
+  }
+  std::vector<int> StencilIDsVisited_;
+  for(const auto& stencilcall : findStencilCalls.getStencilCalls()) {
+    StencilIDsVisited_.push_back(
+        stencilInstantiation->getStencilCallToStencilIDMap().find(stencilcall)->second);
+  }
```
cosunae: why do we need to store all the stencil calls in a vector, and traverse it again to store later the ID? 
Why not directly storing the IDs by the visitor? `findStencilCalls.getStencilCalls()` is not used later one anyhow

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+  for(const auto& bc : stencilInstantiation->getBoundaryConditions()) {
+    auto it = stencilInstantiation->getNameToAccessIDMap().find(bc.first);
+    if(it != stencilInstantiation->getNameToAccessIDMap().end()) {
+      int ID = stencilInstantiation->getAccessIDFromName(bc.first);
+      allBCs.emplace(ID, bc.second);
+    }
+  }
+
+  // Get the order in which the stencils are called:
+  VisitStencilCalls findStencilCalls;
```
cosunae: Here you want to obtain a list of all the stencils executed. Are you sure what you want to do is to visit the StencilCalls? 
Did you think what happen if the stencil call is inlined or fused? 
Isnt `StencilInstantiation::stencils_` the list that you need?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+  for(const std::shared_ptr<Statement>& statement :
+      stencilInstantiation->getStencilDescStatements()) {
+    statement->ASTStmt->accept(findStencilCalls);
+  }
+  std::vector<int> StencilIDsVisited_;
+  for(const auto& stencilcall : findStencilCalls.getStencilCalls()) {
+    StencilIDsVisited_.push_back(
+        stencilInstantiation->getStencilCallToStencilIDMap().find(stencilcall)->second);
```
cosunae: general comment: the following pattern is unsafe
```
map.find(key)->second
```
if the key is not found it will seg fault. We should assert with `assert(map.count(key)); map.[key] `

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+  auto calculateHaloExtents = [&](std::string fieldname) {
+
+    Extents fullExtent;
+    // Did we already apply a BoundaryCondition for this field?
+    // This is the first time we apply a BC to this field, we traverse all stencils that were
+    // applied before
+    std::vector<int> stencilIDsToVisit(StencilIDsVisited_);
+    if(StencilBCsApplied_.count(fieldname) == 0) {
+    } else {
+      for(int traveresedID : StencilBCsApplied_.find(fieldname)->second) {
```
cosunae: assert that the key exists

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+        stencilIDsToVisit.erase(
+            std::remove(stencilIDsToVisit.begin(), stencilIDsToVisit.end(), traveresedID),
+            stencilIDsToVisit.end());
+      }
+    }
+    for(auto& stencil : stencilInstantiation->getStencils()) {
+      for(const int& ID : stencilIDsToVisit) {
```
cosunae: you can replace the second for loop and the if by
```
if(stencilIDsToVisit.count [ stencil->getStencilID() ] 
```
(also replace stencilIDsToVisit by an unordered_set, so that the lookup is constant time

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  auto calculateHaloExtents = [&](std::string fieldname) {
+
+    Extents fullExtent;
+    // Did we already apply a BoundaryCondition for this field?
+    // This is the first time we apply a BC to this field, we traverse all stencils that were
+    // applied before
+    std::vector<int> stencilIDsToVisit(StencilIDsVisited_);
+    if(StencilBCsApplied_.count(fieldname) == 0) {
```
cosunae: remove the empty if by negating the condition and moving the else into the if

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+          }
+          break;
+        }
+      }
+    }
+    return fullExtent;
+  };
+
+  auto insertExtentsIntoMap = [](int fieldID, Extents extents, std::unordered_map<int, Extents>& map){
+      auto fieldExtentPair = map.find(fieldID);
```
cosunae: naming convention: the result of a find is not a pair, so I would better name after iterator.
In any case I would recommend count, rather than
```
it = map.find(key)
it != map.end()
``` 
unless `it` is used later on

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+            const auto& allReadAccesses = acesses.getReadAccesses();
+            const auto& allWriteAccesses = acesses.getWriteAccesses();
+
+            // ReadAccesses can trigger Halo-Updates and Boundary conditions if the following
+            // criteria are fullfilled:
+            // It is a Field (ID!=-1) and we had a write before from another stencil (is in
+            // dirtyFields)
+            for(const auto& readacccess : allReadAccesses) {
+              int originalID = getOriginalID(readacccess.first);
+              if(originalID == FieldType::FT_NotOriginal)
```
cosunae: I think FT_NotOriginal is not very readable... Original is also used to refer to the original name of fields, when calling a function. 
Isnt this type of access refering to Literals? FT_Literal?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+                                 "stencil_function, e.g \n'boundary_condition(zero(), %s);'\n",
+                                 stencilInstantiation->getName(),
+                                 stencilInstantiation->getOriginalNameFromAccessID(originalID),
+                                 stencilInstantiation->getOriginalNameFromAccessID(originalID))
+                        .c_str());
+              }
+              // Calculate the extent and add it to the boundary-condition - Extent map
+              Extents fullExtents = calculateHaloExtents(
+                  stencilInstantiation->getNameFromAccessID(readacccess.first));
+              stencilInstantiation->getBoundaryConditionToExtentsMap().emplace(finder->second,
```
cosunae: use a set pattern rather than get + emplace. And maybe rename `finder`, since it is not very readable, it does not say what it contains

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+                                                                               fullExtents);
+
+              // check if this stencil is called and get its StencilCallDeclStmt (the one to replace)
+              auto test =
+                  stencilInstantiation->getIDToStencilCallMap().find(stencil.getStencilID());
+              if(test == stencilInstantiation->getIDToStencilCallMap().end()) {
+                DAWN_ASSERT_MSG(false, "Stencil Triggering the Boundary Condition is not called");
```
cosunae: use count(key)
I dont understand this assertion, can you explain it? Should there be a map 1 to 1 between stencil call and the stencils of stencil instantiation?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
cosunae: StencilBCsApplied_ is only inserted by this lambda function, apparently, later at the end. 
So is it ever possible that contains elements for field key fieldname here? It does not look like this is the case

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -0,0 +1,332 @@
...
+              AddBoundaryConditions visitor(stencilInstantiation, stencil.getStencilID());
+
+              for(auto& statement : stencilInstantiation->getStencilDescStatements()) {
+                visitor.reset();
+
+                std::shared_ptr<Stmt>& stmt = statement->ASTStmt;
+
+                stmt->accept(visitor);
+                std::vector<std::shared_ptr<Stmt>> stencilCallWithBC_;
+                stencilCallWithBC_.emplace_back(finder->second);
```
cosunae: rename finder and test, by meaningful names

src/dawn/Optimizer/PassStencilSplitter.h
cosunae: this is still not resolved

src/dawn/Optimizer/StencilInstantiation.cpp
```diff
@@ -879,7 +879,16 @@ class StencilDescStatementMapper : public ASTVisitor {
   }
 
   void visit(const std::shared_ptr<BoundaryConditionDeclStmt>& stmt) override {
-    DAWN_ASSERT_MSG(0, "BoundaryConditionDeclStmt not yet implemented");
+//    BoundaryConditions bc;
+//    bc.functor = stmt.get()->getFunctor();
+//    // Extracting all the stencil_function arguments except for the field we apply it to
+//    for(int i = 1; i < (*stmt).getFields().size(); ++i) {
```
cosunae: remove

src/dawn/Optimizer/StencilInstantiation.h
```diff
@@ -31,6 +31,12 @@ namespace dawn {
 
 class OptimizerContext;
 
+///// @brief struct for storing the boundary conditions
+//struct BoundaryConditions {
+//  std::string functor;
```
cosunae: remove

test/unit-test/dawn/Optimizer/Passes/samples/boundary_condition_test_stencil_01.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+stencil SplitStencil {
+  storage foo;
+  storage bar;
+
+  boundary_condition(zero(), foo);
+
+  void Do() {
+    vertical_region(k_start, k_end) {
+      foo = bar[i + 1];
```
cosunae: lets try not to use foo in the code

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: stencil has no get-stages. We can either get the multistages and the stages from there or leave as is. Thoughts?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: This way the visitor really only does the visiting and therefore does not need the stencilinstantiation, I am open to changing it if you think it is beneficial though

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: can we be certain that the order (and potential repetitions) there is the right one? By fetching it form the StencilDescAST, it is definilty the right one. I am pretty sure this is the right way to go

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: no need to assert, we just do not do the erase (the empty if)

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: the key in an unordered_map is const and can therefore not be written with std::transform iirc

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: This refers also to not-originally present fields (such as merged temporaries, cached fields). What would be a better name?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
cosunae: could we have a method that encapsulates this under something like 
`isAFieldVersion(accessID)`  ?

twicki: hopefully this is more clear now

src/dawn/CodeGen/GridTools/ASTStencilDesc.cpp
twicki: we take this to a new pr

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: we leave as is since iterators are coming and will replace this as soon as they arrive

twicki:
launch jenkins

PR #52, opened 2018-01-26
Temporary intervals

cosunae:
Description
=============

Captures the intervals where temporaries are accessed in order to do a proper allocation size of storages in the code generation

Changes: 
* Fix a memory issue with shared_ptr being extracted (OptimizerContext) as raw pointers. When the shared_ptr was out of scope the whole StencilInstantiation was deallocated, while raw pointers still used. 
* Compute the intervals per field being accessed in the Stage::update()
* Unittest the interval computation

### DIFF
--- a/src/dawn/Optimizer/StencilInstantiation.cpp
+++ b/src/dawn/Optimizer/StencilInstantiation.cpp
@@ -506,7 +506,7 @@ public:
                              std::vector<std::shared_ptr<Statement>>& statements,
                              const std::unordered_map<std::string, int>& fieldnameToAccessIDMap)
       : instantiation_(instantiation) {
-
+    DAWN_ASSERT(instantiation);
     // Create the initial scope
     scope_.push(std::make_shared<Scope>(name, statements));
     scope_.top()->LocalFieldnameToAccessIDMap = fieldnameToAccessIDMap;
@@ -544,7 +544,7 @@ public:
   void makeNewStencil() {
     int StencilID = instantiation_->nextUID();
     instantiation_->getStencils().emplace_back(
-        make_unique<Stencil>(instantiation_, instantiation_->getSIRStencil(), StencilID));
+        std::make_shared<Stencil>(*instantiation_, instantiation_->getSIRStencil(), StencilID));
 
     // We create a paceholder stencil-call for CodeGen to know wehere we need to insert calls to
     // this stencil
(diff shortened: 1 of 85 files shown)


cosunae:
launch jenkins

src/dawn/Optimizer/StencilInstantiation.cpp
```diff
@@ -995,10 +995,10 @@ class StencilDescStatementMapper : public ASTVisitor {
 //===------------------------------------------------------------------------------------------===//
 
 StencilInstantiation::StencilInstantiation(OptimizerContext* context,
-                                           const sir::Stencil* SIRStencil, const SIR* SIR)
+                                           const std::shared_ptr<sir::Stencil> SIRStencil,
+                                           const std::shared_ptr<SIR> SIR)
     : context_(context), SIRStencil_(SIRStencil), SIR_(SIR) {
   DAWN_LOG(INFO) << "Intializing StencilInstantiation of `" << SIRStencil->Name << "`";
-  DAWN_ASSERT_MSG(SIRStencil, "Stencil does not exist");
```
twicki: Do we still want to check that the shared-ptr does not point to null
`DAWN_ASSERT_MSG(SIRStencil.get(), "Stencil does not exist");`

test/unit-test/dawn/Optimizer/Passes/TestFieldAccessIntervals.cpp
```diff
@@ -0,0 +1,184 @@
...
+    return optimizer->getStencilInstantiationMap()["compute_extent_test_stencil"]->getStencils();
+  }
+};
+
+TEST_F(TestFieldAccessIntervals, test_field_access_interval_01) {
+  auto stencils = loadTest("test_field_access_interval_01.sir");
+  ASSERT_TRUE((stencils.size() == 1));
+  std::shared_ptr<Stencil> stencil = stencils[0];
+
+  ASSERT_TRUE((stencil->getNumStages() == 2));
```
twicki: why is the extent calculation still here? Don't we do this in TestPAssComputeStageExtents?

test/unit-test/dawn/Optimizer/Passes/TestFieldAccessIntervals.cpp
```diff
@@ -0,0 +1,184 @@
+//===--------------------------------------------------------------------------------*- C++ -*-===//
```
twicki: as a general note, I had to change this kind of setup slightly for boundary condition checks and am now returning a unique-ptr to the ISIR rather then the vector of stencils. Maybe we should be doing that throughout?

src/dawn/Optimizer/StencilInstantiation.cpp
cosunae: done

test/unit-test/dawn/Optimizer/Passes/TestFieldAccessIntervals.cpp
cosunae: yes, but the input data set is different. Just to double check there is nothing weird?

twicki: We can keep it if we assume that the serialized files might be corrupted.  It is an extra layer of safety but I'm not sure if it belongs into a unittest?
PassComputeStageExtents checks that the extent-calculation is correct and ../../SIR/TestSIRSerializer checks that the de-/ serialization works.

test/unit-test/dawn/Optimizer/Passes/TestFieldAccessIntervals.cpp
cosunae: done

cosunae:
launch jenkins

PR #63, opened 2018-02-22
Fix BC bug

twicki:


src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -91,8 +91,12 @@ class AddBoundaryConditions : public ASTVisitorForwarding {
       : instantiation_(instantiation), StencilID_(StencilID) {}
 
   void visit(const std::shared_ptr<StencilCallDeclStmt>& stmt) override {
-    if(instantiation_->getStencilIDFromStmt(stmt) == StencilID_)
-      stencilCallsToReplace_.emplace_back(stmt);
+    auto it = instantiation_->getStencilCallToStencilIDMap().find(stmt);
```
cosunae: how about using a find_if with a lambda?
What happens if not found. Is it a failure, should we assert or something?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -259,13 +264,15 @@ bool PassSetBoundaryCondition::run(const std::shared_ptr<StencilInstantiation>&
...
                                                                         fullExtents);
 
-              // check if this stencil is called and get its StencilCallDeclStmt (the one to
-              // replace)
-              auto stencilcallstmt =
-                  stencilInstantiation->getIDToStencilCallMap().find(stencil.getStencilID());
-              if(stencilcallstmt == stencilInstantiation->getIDToStencilCallMap().end()) {
-                DAWN_ASSERT_MSG(false, "Stencil Triggering the Boundary Condition is not called");
+              bool checkStencilCalled = false;
+              for(auto a : stencilInstantiation->getStencils()) {
```
cosunae: another find_if would be more readable

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -279,8 +286,9 @@ bool PassSetBoundaryCondition::run(const std::shared_ptr<StencilInstantiation>&
                 stmt->accept(visitor);
                 std::vector<std::shared_ptr<Stmt>> stencilCallWithBC_;
                 stencilCallWithBC_.emplace_back(IDtoBCpair->second);
-                stencilCallWithBC_.emplace_back(stencilcallstmt->second);
-
+                if(visitor.getStencilCallsToReplace().size() > 0) {
+                  stencilCallWithBC_.emplace_back(visitor.getStencilCallsToReplace()[0]);
```
cosunae: can it be there are more than 1 elements? It does not look like, so how about using simply a srd::shared_ptr that is null if not found? instead of a vector?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -296,6 +304,8 @@ bool PassSetBoundaryCondition::run(const std::shared_ptr<StencilInstantiation>&
                 }
               }
 
+              std::cout << "here2" << std::endl;
```
cosunae: remove

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: upon testing, you seem to be right but I wonder why.
If we have a stencil like
```
stencil multiplecalls {
  storage a, b;

  boundary_condition(zero(), b);
  boundary_condition(zero(), a);

  void Do() {
    bar(a, b);
    if(global_flag) {
      bar(a, b);
    } else {
      foo(a, b);
    }
    if(global_flag) {
      bar(a, b);
    } else {
      foo(a, b);
    }
  }
};
```
we generate a total of 5 stencils (each only called once). Why is that?

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: nothing should happen. The visitor is called on every stmt in the StencilDescAST. If no stencil to be replaced is found, replacement is skipped and we move on

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: done

src/dawn/Optimizer/PassSetBoundaryCondition.cpp
twicki: sorry

PR #65, opened 2018-03-05
Fixing creation of new intervals for naive codegen

twicki:
Some of the splitting expected the handling of open intervals [..) but `vertical_regions` clearly operate on closed intervals [...]. 

Also,  the constructor for intervals that only gets levels (and therefore uses 2 arguments) breaks once we offset the top-level (as it is the arbitrary number 1<<20). This cleans up all the splitting into proper level / offset constructors taking four arguments.

The changes in the testing had to happen because if you compare elements of a set, c++ applies `!(a < b) && !(b < a)` rather than `a==b`. Since these functors are not overloaded in the dawn::interval (as they are somewhat ambiguous for intervals), we use  the setcomparison method to force a comparison with the `==` operator.

Various test were doing the wrong thing as we should handle intervals as closed ones and a lot of tests assumed open ones.


twicki:
launch jenkins

cosunae:
can you please add a description?

src/dawn/Optimizer/Interval.cpp
```diff
@@ -175,37 +176,84 @@ std::vector<Interval> Interval::computePartition(std::vector<Interval> const& in
...
+        Interval splitHighInterval(curLowInterval.upperLevel(), curTopInterval.upperLevel(),
+                                   curLowInterval.upperOffset() + 1, curTopInterval.upperOffset());
 
         *curLowIt = splitLowInterval;
         *curTopIt = splitMidInterval;
         newIntervals.insert(curTopIt, splitHighInterval);
+        change = true;
+      }
+
+      if(change) {
```
cosunae: can you document a bit the algorithm, when is change triggered and what is supposed to do with the intervals?

twicki: it has a description of what the algorithm is supposed to do in the header. I can update the final statement if you think it would help readability?

cosunae: The header should contain what, given an input , the algorithm compute and returns. 
But what I meant here, is to explain a bit what we are doing in the implementation of the algorithm, so that we can understand in the future how the algorithm works. Step, by step 

See 
 https://github.com/MeteoSwiss-APN/dawn/pull/65/files/fd44ef6844633c4e245db5e38f61047d39bb0683#diff-5ff44e30541a7c4ed622483d82ff9ec2R177
and 
https://github.com/MeteoSwiss-APN/dawn/pull/65/files/fd44ef6844633c4e245db5e38f61047d39bb0683#diff-5ff44e30541a7c4ed622483d82ff9ec2R212

since you modified the algorithm and added an `if(change) ` I was asking to add a comment line 
explaining what the algorithm does in case change variable is set, here
https://github.com/MeteoSwiss-APN/dawn/pull/65/files/fd44ef6844633c4e245db5e38f61047d39bb0683#diff-5ff44e30541a7c4ed622483d82ff9ec2R244

PR #66, opened 2018-03-29
fix tmptostencil function with fields are versioned

cosunae:
This is a fix of the temporary to stencil function pass for the following case, that was caused due to versioning of the fields 

```

stencil Test04 {
  storage a;
  storage b, c, d;
  var tmp_a, tmp_b;
  void Do() {
      tmp_a = a[i + 1] + b;
      tmp_b = c[j - 1] + d;
      c = tmp_a[i - 1] + b;
      d = tmp_b[j + 1] + a;
    }
  }
};
```

### DIFF
--- a/src/dawn/Optimizer/StencilInstantiation.cpp
+++ b/src/dawn/Optimizer/StencilInstantiation.cpp
@@ -630,20 +630,19 @@ void StencilInstantiation::setAccessIDNamePairOfGlobalVariable(int AccessID,
   GlobalVariableAccessIDSet_.insert(AccessID);
 }
 
-void StencilInstantiation::removeAccessID(int AccesssID) {
-  if(NameToAccessIDMap_.count(AccessIDToNameMap_[AccesssID]))
-    NameToAccessIDMap_.erase(AccessIDToNameMap_[AccesssID]);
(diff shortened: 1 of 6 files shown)


src/dawn/Optimizer/StencilInstantiation.cpp
```diff
@@ -630,20 +630,19 @@ void StencilInstantiation::setAccessIDNamePairOfGlobalVariable(int AccessID,
...
-        versions.end());
+void StencilInstantiation::removeAccessID(int AccessID) {
+  if(NameToAccessIDMap_.count(AccessIDToNameMap_[AccessID]))
+    NameToAccessIDMap_.erase(AccessIDToNameMap_[AccessID]);
+
+  AccessIDToNameMap_.erase(AccessID);
+  FieldAccessIDSet_.erase(AccessID);
+  TemporaryFieldAccessIDSet_.erase(AccessID);
+
+  if(variableVersions_.hasVariableMultipleVersions(AccessID)) {
```
twicki: This check in unnecessary, right?

cosunae: I think it is, because the getVertions is calling a map.at(key) which will throw if the key does not exist

PR #73, opened 2018-04-18
Fix in Codegen for temporaries spanning mulitple multistages

twicki:
## Technical Description

If temporaries span more than one  Multistage (or more than one stencil), they are initialized by the stencil-wrapper.  This fix determines the vertical extents that those temporaries must have. The functionality is similar to what was present for naive code generation.

In order to generalize this concept across naive and gridtools code-generation, I slightly adapted the naming convention in the naive code-generartion (going from `storage_t XXX_` to `storage_t m_XXX`. This was the convention used in the gridtools-codegen


twicki:
launch jenkins

src/dawn/CodeGen/CodeGen.cpp
```diff
@@ -8,6 +8,20 @@ size_t CodeGen::getVerticalTmpHaloSize(Stencil const& stencil) {
   return (tmpInterval != nullptr) ? std::max(tmpInterval->overEnd(), tmpInterval->belowBegin()) : 0;
 }
 
+size_t CodeGen::getVerticalTmpHaloSizeForMultipleStencils(
+    const std::vector<std::shared_ptr<Stencil>>& stencils) const {
+  std::shared_ptr<Interval> fullIntervals = stencils[0]->getEnclosingIntervalTemporaries();
+  for(auto stencil : stencils) {
+    auto tmpInterval = stencil->getEnclosingIntervalTemporaries();
+    if(tmpInterval != nullptr) {
+      fullIntervals->merge((*tmpInterval));
```
cosunae: If the first stenci has no temporaries, fullIntervals will be nullptr and this will segfault (on the second iteration)

src/dawn/CodeGen/CodeGen.cpp
```diff
@@ -8,6 +8,20 @@ size_t CodeGen::getVerticalTmpHaloSize(Stencil const& stencil) {
   return (tmpInterval != nullptr) ? std::max(tmpInterval->overEnd(), tmpInterval->belowBegin()) : 0;
 }
 
+size_t CodeGen::getVerticalTmpHaloSizeForMultipleStencils(
+    const std::vector<std::shared_ptr<Stencil>>& stencils) const {
+  std::shared_ptr<Interval> fullIntervals = stencils[0]->getEnclosingIntervalTemporaries();
```
cosunae: There are some issues, stencils can be an empty vector, and stencils[0] will segfault. 
I would suggest to replicate the same algorithm we have in the Stencil.cpp
But also I realized that we are abusing here shared_ptr, we just want to have an output that could have a non initialized state. I think we should use boost::optional, as I was doing in the PR #74 
But I would suggest you to fix this function as we do it in Stencil.cpp and later we refactor all the getEnclosingIntervalTemporaries using a boost::optional if you agree

Also would be good to unittest this

twicki:
launch jenkins

PR #74, opened 2018-04-18
Fix kcache interval

cosunae:
Technical Description
===================

the kcaches were code generated with the interval of the mss, but the interval of the kcache access might be different. Here we fix this


cosunae:
launch jenkins

cosunae:
launch jenkins

CMakeLists.txt
```diff
@@ -99,7 +99,7 @@ if(DAWN_DOCUMENTATION)
   set(doc_packages Python3 Sphinx LaTeX Doxygen)
 endif()
 
-foreach(package bash Protobuf ccache clang-format ${doc_packages})
+foreach(package bash Boost Protobuf ccache clang-format ${doc_packages})
```
twicki: we actively avoided boost in dawn so far. Do you think this was a bad decision or do you think this is necessary enough that we have to take the hit here?

I know that the non-empy checks are a little more verbose, but it seems to be worth it in my eyes. If we want to decrease verbosity, we can also create our own optional (similar to our own any)

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -91,11 +99,6 @@ GTCodeGen::IntervalDefinitions::IntervalDefinitions(const Stencil& stencil)
     for(const Interval& interval : DoMethodIntervals)
       IntervalToNameMap.emplace(interval, Interval::makeCodeGenName(interval));
   }
-
-  // Make sure the intervals for the stencil functions exist
-  for(const auto& stencilFun : stencil.getStencilInstantiation().getStencilFunctionInstantiations())
-    IntervalToNameMap.emplace(stencilFun->getInterval(),
-                              Interval::makeCodeGenName(stencilFun->getInterval()));
```
twicki: why do we no longer need this?

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -406,17 +409,21 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
         ssMS << RangeToString(", ", "gridtools::define_caches(", "),")(
             multiStage.getCaches(),
             [&](const std::pair<int, Cache>& AccessIDCachePair) -> std::string {
+              auto const& cache = AccessIDCachePair.second;
+              DAWN_ASSERT(cache.getInterval().is_initialized() ||
```
twicki: can we assert? should we not just ignore it if it was already initialized?

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -406,17 +409,21 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
...
+                          cache.getCacheIOPolicy() == Cache::local);
+
+              if(cache.getInterval().is_initialized())
```
twicki: if we already assert on top, should we not just merge these asserts?

test/unit-test/dawn/Optimizer/TestsFromSIR/TestStage.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    DAWN_ASSERT_MSG((optimizer->getStencilInstantiationMap().count("compute_extent_test_stencil")),
+                    "compute_extent_test_stencil not found in sir");
+
+    return optimizer->getStencilInstantiationMap()["compute_extent_test_stencil"];
+  }
+};
+
+TEST_F(TestFieldAccessIntervals, test_field_access_interval_01) {
+  auto stencilInstantiation = loadTest("test_field_access_interval_01.sir");
```
twicki: where is this sir file?

cmake/AddBoost.cmake
```diff
@@ -0,0 +1,26 @@
+##===------------------------------------------------------------------------------*- CMake -*-===##
+##                         _       _       
+##                        | |     | |
+##                    __ _| |_ ___| | __ _ _ __   __ _ 
+##                   / _` | __/ __| |/ _` | '_ \ / _` |
+##                  | (_| | || (__| | (_| | | | | (_| |
+##                   \__, |\__\___|_|\__,_|_| |_|\__, | - GridTools Clang DSL
+##                    __/ |                       __/ |
```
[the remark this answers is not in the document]
cosunae: done

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
cosunae: because it is done in line 82. It was very redundant, we were iterating over stencil function instantiations to insert into stencils and then again to insert into intervalToNameMap. Now we just insert into stencils, and from them we build the intervalToNameMap

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
cosunae: If the cache is local, then we dont need any interval, since we dont synchronize with main memory. 
Otherwise we need an interval, that means that the boost::optional<> has to be initialized (i.e. not null)

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
cosunae: the problem is  *(cache.getInterval) will seg fault if not initialized

test/unit-test/dawn/Optimizer/TestsFromSIR/TestStage.cpp
cosunae: It is in Optimizer/Passes.... 
I didnt create new sir files, but trying to resue the existing ones

CMakeLists.txt
twicki: Just to archive here:

We discussed this and came to the conclusion that we use header only libraries if they are fine in the current state (i.e a re-implementation does not gain us a significant advantage)

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
twicki: ok

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
twicki: but we already asserted if `cache.getInterval().is_initialized()`, so we can do:
```
DAWN_ASSERT(cache.getInterval().is_initialized() || cache.getCacheIOPolicy() == Cache::local);
DAWN_ASSERT(intervalDefinitions.IntervalToNameMap.count(*(cache.getInterval())));
```

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
twicki: yeah, but if we find a local cache, we should not assert, we should just not anything, right?

test/unit-test/dawn/Optimizer/TestsFromSIR/TestStage.cpp
twicki: ah, ok

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
cosunae: so say you have is_initialized() == False and cache.getCacheIOPolicy() == Cache::local

The first assert will pass, and second assert will seg fault. While you would like to pass

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
cosunae: yes, that is why we assert if is_initialized or policy is local (line below)

PR #84, opened 2018-05-02
remove debug MSS split and add debug Stage Split

twicki:
## Technical Description

Since our current syntax allows for iterative stencils in the form of 
```
vertical_region(k_start+1, k_end){
  var a = computation();
  var b = a[k-1] * computation();
  a = a / b;
}
```
we are not able to split every statement into a separate multistage since in the second iteration of the k-loop at line two we rely on the third line of this loop to be executed on the first k-level.

Therefore the suggested debugging backend is not a valid option. We moved the splitting of each statement to the stage-level. This relies on the multistage-splitter to be a mandatory, working piece to generate legal code. The option of having a separate algorithm that checks how many cuts we can maximally do in a given multistage without making the code illegal is though after but not yet finalized

### DIFF
--- a/src/dawn/Optimizer/PassStageSplitter.cpp
+++ b/src/dawn/Optimizer/PassStageSplitter.cpp
@@ -28,100 +28,25 @@
 
 namespace dawn {
 
-PassStageSplitter::PassStageSplitter(StageSplittingStrategy strategy)
-    : Pass("PassStageSplitter"), strategy_(strategy) {
-  isDebug_ = true;
-}
+PassStageSplitter::PassStageSplitter() : Pass("PassStageSplitter") {}
 
-namespace {
-std::function<void(std::list<std::shared_ptr<Stage>>::iterator&, std::deque<int>&,
-                   std::deque<std::shared_ptr<DependencyGraphAccesses>>&,
-                   const std::shared_ptr<StencilInstantiation>&, const Options&, int, int,
-                   const std::string&, const std::string&, int&, std::shared_ptr<MultiStage>&)>
-setOptimizedLoopContent() {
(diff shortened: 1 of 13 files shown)


twicki:
launch pipeline

src/dawn/Optimizer/PassStageSplitter.cpp
```diff
@@ -110,6 +166,45 @@ bool PassStageSplitter::run(const std::shared_ptr<StencilInstantiation>& stencil
           stage.update();
           ++stageIt;
         }
+
+      };
+}
+} // anonymous namespace
+
+bool PassStageSplitter::run(const std::shared_ptr<StencilInstantiation>& stencilInstantiation) {
```
cosunae: as discussed, maybe we can have the debug thing only for the mss, since it will already serve the purpose

twicki:
launch pipeline

src/dawn/Optimizer/PassStageSplitter.cpp
```diff
@@ -28,7 +28,7 @@
 
 namespace dawn {
 
-PassStageSplitter::PassStageSplitter() : Pass("PassStageSplitter") {}
+PassStageSplitter::PassStageSplitter() : Pass("PassStageSplitter") { isDebug_ = true; }
```
cosunae: minor thing, but would be better to have a ctr initialization
```, isDebug_(true){}```
Also in PassMultiStageSplitter.cpp
or pass it to the Pass ctr `Pass("name", true)` and then it could be set `const bool`

twicki: ctr does not work as it is a protected member of the base-class.

We can go the second route if you think it is prettier

twicki: done

twicki:
launch pipeline

PR #85, opened 2018-05-07
Addition of Field-Dimensions

twicki:
(this goes with a PR in [GTClang](https://github.com/MeteoSwiss-APN/gtclang/pull/69))

## Technical Description

This allows the user to specify the dimensionality of fields that are passed as input. Allowed inputs are:
```
storage
storage_i
storage_j
storage_k
storage_ij
storage_ik
storage_jk
storage_ijk
```
where the first and the last are the same

With this feature we gain two things:

* More error-checks are possible.
Currently you can pass a `storage_j_t` to the stencil and call it with i and k offsets. This can be prevented now if the user specifies the dimensionality of storages used.
* In the near future, there will be a Gridtools update that prohibits the notype-stencils that we currenly use.
In order for our setup to work, we need type information at the time of code-generation.

### DIFF
--- a/src/dawn/SIR/SIR.h
+++ b/src/dawn/SIR/SIR.h
@@ -159,9 +159,10 @@ struct StencilFunctionArg {
 /// @ingroup sir
 struct Field : public StencilFunctionArg {
   Field(const std::string& name, SourceLocation loc = SourceLocation())
-      : StencilFunctionArg{name, AK_Field, loc}, IsTemporary(false) {}
+      : StencilFunctionArg{name, AK_Field, loc}, IsTemporary(false), fieldDimensions({{0, 0, 0}}) {}
 
   bool IsTemporary;
+  Array3i fieldDimensions;
 
   static bool classof(const StencilFunctionArg* arg) { return arg->Kind == AK_Field; }
   bool operator==(const Field& rhs) const { return comparison(rhs); }
--- a/src/dawn/SIR/SIRSerializer.cpp
+++ b/src/dawn/SIR/SIRSerializer.cpp
@@ -129,6 +129,9 @@ static void setInterval(sir::proto::Interval* intervalProto, const sir::Interval
 static void setField(sir::proto::Field* fieldProto, const sir::Field* field) {
   fieldProto->set_name(field->Name);
   fieldProto->set_is_temporary(field->IsTemporary);
+  for(const auto& initializedDimension : field->fieldDimensions){
+    fieldProto->add_field_dimensions(initializedDimension);
+  }
   setLocation(fieldProto->mutable_loc(), field->Loc);
 }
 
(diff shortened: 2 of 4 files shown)


src/dawn/SIR/SIR.h
```diff
@@ -159,9 +159,10 @@ struct StencilFunctionArg {
 /// @ingroup sir
 struct Field : public StencilFunctionArg {
   Field(const std::string& name, SourceLocation loc = SourceLocation())
-      : StencilFunctionArg{name, AK_Field, loc}, IsTemporary(false) {}
+      : StencilFunctionArg{name, AK_Field, loc}, IsTemporary(false), fieldDimensions({{0, 0, 0}}) {}
```
cosunae: just `fieldDimensions{0,0,0} ` ?

twicki: we are actually using the double brackets pretty consistently for ArrayXi in out codes, examples [here](https://github.com/MeteoSwiss-APN/dawn/blob/26392782936f260876cf95b10e870a9df30a6459/src/dawn/Optimizer/AccessComputation.cpp#L150) or [here](https://github.com/MeteoSwiss-APN/dawn/blob/26392782936f260876cf95b10e870a9df30a6459/src/dawn/SIR/SIRSerializer.cpp#L763) amongst many others. 
I stuck to that style.
If you don't see a reason to do so, we can consider changing to single-bracket style?

cosunae: I was not aware of that, but what is the C++ reason behind that? 
Also here we are using triple brackets `({{`

src/dawn/SIR/SIRSerializer.cpp
```diff
@@ -597,6 +600,17 @@ static SourceLocation makeLocation(const T& proto) {
...
   field->IsTemporary = fieldProto.is_temporary();
+  if(!fieldProto.field_dimensions().empty()) {
+      auto throwException = [&fieldProto](const char* member) {
+        throw std::runtime_error(format("Field::%s (loc %s) exceeds 3 dimensions", member,
+                                        makeLocation(fieldProto)));
+      };
+    if(fieldProto.field_dimensions().size() > 3)
+      throwException("field_dimensions");
+
+    for(int i = 0; i < fieldProto.field_dimensions().size(); ++i)
```
cosunae: can we use a std::copy

src/dawn/SIR/SIR.h
twicki: the round brackets are from the initializer-list, but the Array-Ctor has the 2 as anywhere else.

It is because c++11 does not let you write  `std::array<int, 3> foo{1,2,3}` but you are required to write `std::array<int, 3> foo{{1,2,3}}` whereas you are allowed to write `std::array<int, 3> foo = {1,2,3}`.
I think that  to keep the look of it more consistent, we went with with double-brackets everywhere

src/dawn/SIR/SIRSerializer.cpp
twicki: sure

PR #83, opened 2018-05-02
[outdated] Fix kcaches

cosunae:
[DO NOT REVIEW yet]  I am doing a self review

PR #89, opened 2018-05-15
[outdated] New GT syntax

twicki:
This pull request goes with the one in [GTClang](https://github.com/MeteoSwiss-APN/gtclang/pull/71)

## Technical Description
- The abstract class that we used to store the stencils, `gridtools::stencil<gridtools::notype>`, does no longer exist. We updated the code-generation to incorporate the new interface.
- Since we have the information about the storage-dimensions, we incorporate them into code generation and do no longer need them to be templated.
- A better version for reading the performance meters has been implemented


twicki:
In order to keep master valid all the time, this and #83  are both merged in #103, hence outdated. Can be closed once we merge #103

twicki:
has been merged with #103

PR #96, opened 2018-06-08
add static assert code gen that checks extents against given halos

stefanmoosbrugger:


cosunae:
Hi Stefan, 

thanks for this, your first PR here, lets start the fun :)

Here some general comments to improve this. 
* There is quite some logic, so I think we should encapsulate it into a method so that we can reuse it and more importantly test it. Since we are not testing it, and did you try the result? I believe there is a bug, which I will point at later. 
I suggest we could have something like 
std::unordered_map<int, Extents> Stencil::computeEnclosingAccessExtents() 
that computes the enclosing extent of each accessID for all the mss/stages

* Create some unittests, this you will need to learn how to do, since it is difficult to create an IIR of an example computation. Today we do it via an HIR that we generate through gtclang and put it in json format in the repository (Optimizer/Passes/*.sir) -> Tobias can show you how to do this.

* You need to check that gtclang examples still compile with this, trigger "launch pipeline"

* Provide a technical description, explaining this PR

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -465,6 +469,14 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
           const auto& field = fields[accessorIdx];
           std::string paramName = stencilInstantiation->getNameFromAccessID(field.getAccessID());
 
+          // intialize map entry      
+          collectedExtents[paramName] = std::vector<int>(field.getExtents().getSize()*2);
```
cosunae: this I think is a bug, looks like you tried to compute the enclosing extent, but you are infact resetting the extent of each param for every stage

cosunae: Also no need to use std::vector, we normally try to use structs/classes that makes the internal data structures readable... here you can simply use Extents class. Also you have plenty of functionality in those classes. 
So for example you don need to compute the enclosing using max,min yourself. There are methods for that in the Extents

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -465,6 +469,14 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
...
+          // get largest and smallest extent for each field
+          for(int dim = 0; dim < field.getExtents().getSize(); ++dim) {
```
cosunae: But anyhow as said, I would put this logic in a method of Stencil.h, which I would call from here.

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -566,6 +578,28 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
 
     StencilConstructor.startBody();
 
+    // Add static asserts to check halos against extents
+    StencilConstructor.addComment("Check if extents do not exceed the halos");
+    for(int i = 0; i < numFields; ++i) {
+      if(!StencilFields[i].IsTemporary) {
+        auto max_extents = collectedExtents[StencilFields[i].Name];
```
cosunae: if you have a method 
``` std::unordered_map<int, Extents> Stencil::computeEnclosingAccessExtents()``` 
I would reference the field via accessID rather than name

File: src/dawn/CodeGen/GridTools/GTCodeGen.cpp [hunk 4]
```diff
@@ -566,6 +578,28 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
...
+        DAWN_ASSERT_MSG(max_extents.size() % 2 == 0,
+                      "Extents always have to come in pairs (+/-)");
+        for(int dim = 0; dim < (max_extents.size() / 2); ++dim) {
+          // assert for + accesses
```
cosunae: we will need the same protection in the naive code generation as well? 
Could we have this in a method that we can reuse between naive and gridtools code gen?

stefanmoosbrugger:
Thanks for the extensive review. :-) I didn't mention that I created the PR just to try the pipeline. It's not even close to be done as it is not tested at all yet (no unit test, etc.).

stefanmoosbrugger:
Ok, now it should be ready for a short review. Logic is moved to stencil.h! Unittests are there. @cosunae Please do a short check on plausibility of extents that are asserted in the unit test.

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -387,6 +387,9 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
 
     std::vector<std::string> makeComputation;
 
+    // map field names to maximum extent 
+    std::map<std::string, std::vector<int>> collectedExtents;
```
cosunae: this seems to be a leftover, remove?

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -566,6 +570,27 @@ GTCodeGen::generateStencilInstantiation(const StencilInstantiation* stencilInsta
 
     StencilConstructor.startBody();
 
+    // Add static asserts to check halos against extents
+    StencilConstructor.addComment("Check if extents do not exceed the halos");
+    auto exts = (*stencilInstantiation->getStencils()[stencilIdx]).computeEnclosingAccessExtents();
```
cosunae: I would not use auto here, but the real type, for readability. Specially since the name indicates this could be of class `Extents`

src/dawn/Optimizer/Stencil.cpp
```diff
@@ -445,4 +445,23 @@ std::ostream& operator<<(std::ostream& os, const Stencil& stencil) {
   return os;
 }
 
+std::unordered_map<int, Extents> const& Stencil::computeEnclosingAccessExtents() {
```
cosunae: yes, since this computation is expensive, I guess computing into a member a returning by const& makes lot of sense. 
However the problem, is that this creates information that should belong to the internal IR, which is derived from other information that is generated from other Passes. And is difficult to know when this has to be recomputed and when is properly updated. Also there might be other derived states like the method `updateFields`

So I propose the following convention:
* ```void class::computeXXX``` just compute a derive state of the internal IR, in this case the map of accessID to extent. 
* ```auto class::getXXX``` is just a getter
* there is an update method that updates all the derived states of the IIR, i.e.
```
void update() {
  updateFields();
  computeEnclosingAccessExtents();
}
```

you need then to figure where you would need to apply this update, which is probably at the and of all the passes (unless one of the passes would require this information, which is not the case)

test/unit-test/dawn/Optimizer/Passes/TestComputeMaxExtent.cpp
```diff
@@ -0,0 +1,139 @@
...
+  ASSERT_TRUE((stencils.size() == 1));
+  std::shared_ptr<Stencil> stencil = stencils[0];
+
+  ASSERT_TRUE((stencil->getNumStages() == 2));
+  auto exts = stencil->computeEnclosingAccessExtents();
+  ASSERT_TRUE((exts.size() == 3));
+  ASSERT_TRUE((stencil->getStencilInstantiation().getNameFromAccessID(1) == "u"));
+  ASSERT_TRUE((stencil->getStencilInstantiation().getNameFromAccessID(2) == "out"));
+  ASSERT_TRUE((stencil->getStencilInstantiation().getNameFromAccessID(3) == "lap"));
+  ASSERT_TRUE((exts[1] == Extents{-2,2,-2,2,0,0}));
```
cosunae: I would, as a general rule, use EXPECT_EQ, reason is: 
* in case of failure, it shows all other failures, which is useful, while the assert stops
* it prints value vs expected value, while ASSERT_TRUE will just fail

src/dawn/Optimizer/Stencil.cpp
stefanmoosbrugger: i thought about the issue that this computation is expensive and should probably not computed everytime somebody is calling the method. i left it there because of the issue you mentioned (probably modified by a pass or so).

But nevertheless the code is only generated once and therefore the computeEnclosingAccessExtents is only called once (at the moment). Do you see any other case where the maximum enclosing extents are use beside the code gen?

test/unit-test/dawn/Optimizer/Passes/TestComputeMaxExtent.cpp
stefanmoosbrugger: fixed with https://github.com/MeteoSwiss-APN/dawn/pull/96/commits/6e09d5205ce94885f57ba4c9620488826e682232

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
stefanmoosbrugger: fixed with https://github.com/MeteoSwiss-APN/dawn/pull/96/commits/6e09d5205ce94885f57ba4c9620488826e682232

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
stefanmoosbrugger: fixed with https://github.com/MeteoSwiss-APN/dawn/pull/96/commits/6e09d5205ce94885f57ba4c9620488826e682232

test/unit-test/dawn/Optimizer/Passes/TestComputeMaxExtent.cpp
```diff
@@ -0,0 +1,139 @@
...
+  ASSERT_TRUE((stencils.size() == 1));
+  std::shared_ptr<Stencil> stencil = stencils[0];
+
+  ASSERT_TRUE((stencil->getNumStages() == 3));
+  auto exts = stencil->computeEnclosingAccessExtents();
+  EXPECT_EQ(exts.size(), 6);
+  EXPECT_EQ((stencil->getStencilInstantiation().getNameFromAccessID(1)), "u");
+  EXPECT_EQ((stencil->getStencilInstantiation().getNameFromAccessID(2)), "out");
+  EXPECT_EQ((stencil->getStencilInstantiation().getNameFromAccessID(3)), "coeff");
+  EXPECT_EQ(exts[1], (Extents{-2,2,-2,2,0,0}));
```
twicki: needs beautification

src/dawn/Optimizer/Stencil.cpp
stefanmoosbrugger: Maybe I didn't get it right. So please correct me if I'm wrong! The extent information is contained in the different stages. The first place were we can think or reason about "maximum extents" of several stages is a stencil. 

So I would personally vote for keeping this information in the stencil as I think it is more confusing and complicated to keep it outside. What do you mean with derived information?

**I've got another question regarding the update and expensiveness of the computation:** Is there any other case where the maximum enclosing extents are use beside the code gen? Is there any possibility that maximum enclosing extents are changing after the IR was generated?

twicki: Yes, the first point where it makes sense to reason about that is a stencil. What happens though if we split a user-stencil [IStencil] (aka input) into two stencils being generated [GStencil]. There is a 1:n relationship between those two.
On what level is it useful to think about extents? I think it might be interetsing for the user to see if the field he passes to the driver is big enough for all the GStencils and therefore we should link it to IStencil [which is 1:1 with stencilInstantiaion]

I consider every result of an algorithm on data to be "derived" information. Currently we do our best to store those things in the StencilInstantiation

Not yet (otherwise the algorithm would have been in place :P) But I think it could happen in the future. @cosunae thoughts?
Depending on where we compute the maximum extents, it might happen. Think of what happens if this is the first thing we do and later we merge two stages (in the stage-merging pass), one of which has an extent, one does not. The one that did not was computed without and now has an extent.
There is where the automatic reordering / re-triggering of passes would be helpful. Assume we create a new pass that needs to be ran before stage-merging and requires full extent computation as input. we would need to run it before said pass and at the end once more.

https://github.com/MeteoSwiss-APN/dawn/blob/56af8b3eb43de26af4682f14fd3d4bcd5b4bf449/src/dawn/Compiler/DawnCompiler.cpp#L158
is included twice since this is not in place yet and has to be done manually

PR #97, opened 2018-06-18
Fix cmake2

cosunae:


cosunae:
launch pipeline

cosunae:
launch pipeline

PR #100, opened 2018-06-21
Source location comparison fix

stefanmoosbrugger:
Tiny bug fix in source location comparison tool. 
Should not affect neither performance, test results, etc.


stefanmoosbrugger:
launch jenkins

PR #101, opened 2018-07-03
add boost flags

cosunae:
To be merged after the new codegen and kcache fixes have been merged


twicki:
why do we need this?

twicki:
done

PR #106, opened 2018-07-09
Fix temporary to comp on the fly with nested functions

cosunae:


src/dawn/Optimizer/DoMethod.cpp
```diff
@@ -80,4 +80,32 @@ const std::shared_ptr<DependencyGraphAccesses>& DoMethod::getDependencyGraph() c
...
+
+public:
+  CheckNonNullStatementVisitor() {}
+  virtual ~CheckNonNullStatementVisitor() {}
+
+  bool getResult() const { return result_; }
+
+  virtual void visit(const std::shared_ptr<ExprStmt>& expr) override {
+    if(!isa<NOPExpr>(expr->getExpr().get()))
+      result_ = true;
```
twicki: this should also have a  `expr->getExpr()->accept(*this);` in the else clause

src/dawn/Optimizer/DoMethod.cpp
```diff
@@ -80,4 +80,32 @@ const std::shared_ptr<DependencyGraphAccesses>& DoMethod::getDependencyGraph() c
...
+  }
+};
+
+bool DoMethod::isEmptyOrNullStmt() const {
+  for(auto const& statementAccessPair : statementAccessesPairs_) {
+    std::shared_ptr<Stmt> root = statementAccessPair->getStatement()->ASTStmt;
```
twicki: should be `const std::shared_ptr<Stmt>& root`

src/dawn/Optimizer/DoMethod.cpp
```diff
@@ -80,4 +80,32 @@ const std::shared_ptr<DependencyGraphAccesses>& DoMethod::getDependencyGraph() c
...
+    CheckNonNullStatementVisitor checker;
+    root->accept(checker);
+
+    if(checker.getResult())
```
twicki: if our convention is adding brackets around if's we should also do it here?

src/dawn/Optimizer/MultiStage.cpp
```diff
@@ -210,4 +210,12 @@ void MultiStage::renameAllOccurrences(int oldAccessID, int newAccessID) {
   }
 }
 
+bool MultiStage::isEmptyOrNullStmt() const {
+  for(auto stageIt = getStages().begin(); stageIt != getStages().end(); ++stageIt) {
+    if(!(*stageIt)->isEmptyOrNullStmt())
+      return false;
```
twicki: brackets

src/dawn/Optimizer/PassTemporaryToStencilFunction.cpp
```diff
@@ -323,8 +373,8 @@ bool PassTemporaryToStencilFunction::run(
             if(stmt->ASTStmt->getKind() != Stmt::SK_ExprStmt)
               continue;
 
-            // TODO catch a temp expr
             {
+              // TODO catch a temp expr
```
twicki: still todo?

src/dawn/Optimizer/PassTemporaryToStencilFunction.cpp
```diff
@@ -372,9 +422,11 @@ bool PassTemporaryToStencilFunction::run(
               isATmpReplaced = isATmpReplaced || (tmpReplacement.getNumTmpReplaced() != 0);
 
               if(tmpReplacement.getNumTmpReplaced() != 0) {
+
                 std::vector<std::shared_ptr<StatementAccessesPair>> listStmtPair;
 
-                // TODO need to combine call to StatementMapper with computeAccesses in a clean way
+                // TODO need to combine call to StatementMapper with computeAccesses in a clean
+                // way
```
twicki: still todo

src/dawn/Optimizer/StatementMapper.cpp
```diff
@@ -222,6 +222,7 @@ void StatementMapper::visit(const std::shared_ptr<StencilFunCallExpr>& expr) {
       // Clone the AST s.t each stencil function has their own AST which is modifiable
       ast = ast->clone();
 
+      // TODO decouple the funciton of stencil function instantiation from the statement mapper
```
twicki: still open?

src/dawn/Optimizer/StencilFunctionInstantiation.cpp
```diff
@@ -149,12 +149,23 @@ StencilFunctionInstantiation::getCallerInitialOffsetFromAccessID(int callerAcces
...
+  ArgumentIndexToCallerAccessIDMap_[argIdx] = accessID;
+}
+
 void StencilFunctionInstantiation::setCallerInitialOffsetFromAccessID(int callerAccessID,
                                                                       const Array3i& offset) {
   CallerAcceessIDToInitialOffsetMap_[callerAccessID] = offset;
 }
 
 bool StencilFunctionInstantiation::isProvidedByStencilFunctionCall(int callerAccessID) const {
+  //  auto pos = std::find_if(ArgumentIndexToCallerAccessIDMap_.begin(),
```
twicki: comments

src/dawn/Optimizer/StencilFunctionInstantiation.cpp
```diff
@@ -294,6 +305,7 @@ void StencilFunctionInstantiation::setAccessIDOfExpr(const std::shared_ptr<Expr>
 
 void StencilFunctionInstantiation::mapExprToAccessID(const std::shared_ptr<Expr>& expr,
                                                      int accessID) {
+  // TODO redo assertion in one line
```
twicki: todo open

src/dawn/Optimizer/StencilFunctionInstantiation.cpp
```diff
@@ -336,20 +348,13 @@ StencilFunctionInstantiation::getExprToStencilFunctionInstantiationMap() const {
 
 void StencilFunctionInstantiation::insertExprToStencilFunction(
     const std::shared_ptr<StencilFunctionInstantiation>& stencilFun) {
+  // TODO make sure it does not exist already
```
twicki: todo open

src/dawn/Optimizer/StencilFunctionInstantiation.cpp
```diff
@@ -624,17 +629,25 @@ void StencilFunctionInstantiation::dump() const {
 }
 
 void StencilFunctionInstantiation::closeFunctionBindings() {
+  std::vector<int> arglist(getArguments().size());
+  for(std::size_t argIdx = 0; argIdx < getArguments().size(); ++argIdx) {
```
twicki: `std::iota(arglist.begin(), arglist.end(), 0);`

src/dawn/Optimizer/StencilFunctionInstantiation.h
```diff
@@ -360,8 +358,8 @@ class StencilFunctionInstantiation {
   getExprToStencilFunctionInstantiationMap() const;
 
   /// @brief Get StencilFunctionInstantiation by name
-  const std::unordered_map<std::string, std::shared_ptr<StencilFunctionInstantiation>>&
-  getNameToStencilFunctionInstantiationMap() const;
+  //  const std::unordered_map<std::string, std::shared_ptr<StencilFunctionInstantiation>>&
```
twicki: rm

src/dawn/Optimizer/StencilFunctionInstantiation.h
```diff
@@ -427,6 +425,8 @@ class StencilFunctionInstantiation {
 
   /// @brief finalizes the binding of the arguments of a stencil function.
   /// In particular it associates new accessIDs of arguments that are nested stencil function calls
+  void closeFunctionBindings(const std::vector<int>&);
```
twicki: is it necessary to do this on the IIR-Level? why not let the pass handle this?

src/dawn/Optimizer/StencilInstantiation.cpp
```diff
@@ -1104,17 +1074,49 @@ StencilInstantiation::makeStencilFunctionInstantiation(
 
 void StencilInstantiation::insertExprToStencilFunction(
     std::shared_ptr<StencilFunctionInstantiation> stencilFun) {
+  // TODO gather all the stencil function properties in a struct
```
twicki: todo

src/dawn/Optimizer/StencilInstantiation.cpp
```diff
@@ -1104,17 +1074,49 @@ StencilInstantiation::makeStencilFunctionInstantiation(
...
+  //  DAWN_ASSERT(ExprToStencilFunctionInstantiationMap_.count(stencilFun->getExpression()));
+  //  DAWN_ASSERT(nameToStencilFunctionInstantiationMap_.count(stencilFun->getExpression()));
   ExprToStencilFunctionInstantiationMap_.emplace(stencilFun->getExpression(), stencilFun);
-  nameToStencilFunctionInstantiationMap_.emplace(stencilFun->getExpression()->getCallee(),
-                                                 stencilFun);
+}
+
+void StencilInstantiation::deregisterStencilFunction(
+    std::shared_ptr<StencilFunctionInstantiation> stencilFun) {
+  // TODO move all this to the Remove function
```
twicki: todo

src/dawn/Support/RemoveIf.hpp
```diff
@@ -0,0 +1,35 @@
...
+//
+//===------------------------------------------------------------------------------------------===//
+
+#ifndef DAWN_REMOVEIF_H
+#define DAWN_REMOVEIF_H
+#include <list>
+
+namespace dawn {
+
+// TODO unittest
```
twicki: testing

src/dawn/Optimizer/StencilFunctionInstantiation.h
twicki: the call to close the function bindings is only ever used by this one pass. Having the functionality in the StencilInstantiation and exposing it to everyone is great for re-use but adds to compile-time and is worse in separation of concerns if it is only used once by one specific instance.

Would it make sense that the pass itself has a function call (in an anonymous namespace) that does exactly this that gets a shared-pointer to a StencilInstantiation as an argument?

src/dawn/Optimizer/StencilInstantiation.cpp
twicki: ok

twicki:
just one more minor thing before we can merge

cosunae:
launch pipeline

src/dawn/Optimizer/StencilFunctionInstantiation.h
cosunae: ok, I see now what you meant, thanks. 
This function is called from within  `finalizeStencilFunctionSetup`, that is also called from StatementMapper. I added all these function to separate different functionality into separate functions, mainly to improve readability. The thing is that StatementMapper is doing lot of stuff, among others instantiating the stencil functions. It is doing low level stuff with the expressions, and so on and very hard to follow. Now, in the temp -> stencil function we have new sir::StencilFunction that we need to instantiate in the same manner. So I had to extract the functionality in methods that we can reuse, and tried to make it readable. For example, since the stencil function is instantiated in different phases, I created a StencilFunctionCandidate that is being parsed and filled and only in the finalize method, the function arguments are bound, and the candidate is promoted from candidate to a stencil function....
In summary, I think we reuse this code, but furthermore improves the readability of it, but maybe have a look again, and let me know otherwise if we can do better

PR #117, opened 2018-07-23
better getters for the stencils

twicki:
## Technical Descriptions

With the new syntax for the performance-meters of Gridtools, it is easier to access them in this fashion

Note that this PR goes with the one in [GTClang](here)


twicki:
launch jenkins

Issue #121, opened 2018-08-07
illegal field versioning


twicki:
## Technical issue
If field conversioning is applied to an input field that has not been written to before, no synchronization happens. This leads to inputdata being completely ignored

## Example
```
stencil test {
  storage a;
  Do {
    vertical_region(k_start, k_end) { 
        a += a[i + 1];
    }
  }
};
```
Output of the field [initialzied with one] is 1 rather than 2.

mroethlin:
this is still open as of 06.11.2019 / master at 03778b49d7fbec23e2ecc9d7cab30fa79af09ec5

PR #122, opened 2018-08-13
fix field versioning

twicki:
## Technical Description
Implementation of a fix to the issue reported in https://github.com/MeteoSwiss-APN/dawn/issues/121
The thing to note here is that the three function calls in PassFieldversioning are the same as the ones we have in PassSetNonTempCaches. We keep them here in two seperate places and diff once the IIR refactoring is done


twicki:
launch pipeline

twicki:
launch pipeline

twicki:
no longer needed

PR #128, opened 2018-09-18
Backend option

cosunae:
Technical Description
===================

The code generation belongs to dawn, so shouldnt the code generation option [naive, gt, ...] be in dawn instead of gtclang?

Works together with https://github.com/MeteoSwiss-APN/gtclang/pull/90


twicki:
launch jenkins

PR #135, opened 2018-10-10
fix vertical extents

cosunae:
This is the typical problem, that when we compute the extents of accesses, and it is computed from an offset [k+2], the resulting extent is Extent{0,2}, instead of Extent[2,2]


cosunae:
launch pipeline

cosunae:
launch pipeline

cosunae:
launch pipeline

PR #138, opened 2018-10-13
Fix PassInlining for conditionals within stencil functions

cosunae:


cosunae:
launch pipeline

src/dawn/CodeGen/Cuda/CudaCodeGen.cpp
```diff
@@ -360,8 +360,8 @@ void CudaCodeGen::generateCudaKernelCode(
       }
       if(useTmpIndex(ms, stencilInstantiation) && !kmin.null()) {
         cudaKernel.addComment("jump tmp iterators to match the beginning of next interval");
-        cudaKernel.addStatement("idx_tmp += kstride_tmp*(" + std::to_string(interval.lowerBound()) +
-                                "-" + intervalDiffToString(kmin, "ksize-1") + ")");
+        cudaKernel.addStatement("idx_tmp += kstride_tmp*(" + intervalDiffToString(kmin, "ksize-1") +
```
twicki: "ksize - 1"

src/dawn/Optimizer/PassInlining.cpp
```diff
@@ -74,6 +74,7 @@ class Inliner : public ASTVisitor {
   };
 
   std::stack<ArgListScope> argListScope_;
+  std::stack<std::unique_ptr<iir::StatementAccessesPair> const*> currentStmtAccessesPair_;
```
twicki: const T

src/dawn/CodeGen/Cuda/CudaCodeGen.cpp
cosunae: done. 
question in general: do we want to format these strings? because clang-format will format anyhow the generated code .

src/dawn/Optimizer/PassInlining.cpp
cosunae: done

src/dawn/CodeGen/Cuda/CudaCodeGen.cpp
twicki: I am unsure, I would but if the overhead is too big it's not worth it

twicki:
merge as soon as we get the ok from jenkins

PR #152, opened 2018-11-16
Latest gt

cosunae:


src/dawn/CodeGen/CodeGen.cpp
```diff
@@ -132,9 +132,7 @@ void CodeGen::generateBCHeaders(std::vector<std::string>& ppDefines) const {
   }
 
   if(containBC) {
-    ppDefines.push_back("#ifdef __CUDACC__\n#include "
-                        "<boundary-conditions/apply_gpu.hpp>\n#else\n#include "
-                        "<boundary-conditions/apply.hpp>\n#endif\n");
+    ppDefines.push_back("#include <gridtools/boundary-conditions/boundary.hpp>\n");
```
twicki: phu, nice

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -605,41 +606,42 @@ void GTCodeGen::generateStencilClasses(
...
-                      ">(p_" + stencilInstantiation->getNameFromAccessID(AccessIDCachePair.first) +
-                      "())")
-                  .str();
-            });
+
+        std::vector<iir::Cache> ioCaches;
+        for(const auto& cacheP : multiStage.getCaches()) {
+          if((cacheP.second.getCacheIOPolicy() == iir::Cache::CacheIOPolicy::bpfill) ||
+             (cacheP.second.getCacheIOPolicy() == iir::Cache::CacheIOPolicy::epflush)) {
+            continue;
```
twicki: why can we just continue? We should need to fully fill or flush, right?

cosunae: here we filter the container with only the caches that we support, i.e. in case of bpfill or epflush we continue, that means that we dont insert into the container, but others will be considered. Was that the question?

cosunae:
launch pipeline

cosunae:
launch jenkins

PR #153, opened 2018-11-16
nothing

twicki:
to test codegen in the pipe


twicki:
launch pipeline

twicki:
launch pipeline

PR #155, opened 2018-11-20
fix inlining

cosunae:
When expr's of a function are inlined in parent scope, the name of the expr should be changed to the caller expr name

Goes together with https://github.com/MeteoSwiss-APN/gtclang/pull/103

### DIFF
--- a/src/dawn/SIR/ASTExpr.h
+++ b/src/dawn/SIR/ASTExpr.h
@@ -402,6 +402,8 @@ public:
   const std::string& getName() const { return name_; }
   std::string& getName() { return name_; }
 
+  void setName(std::string name) { name_ = name; }
+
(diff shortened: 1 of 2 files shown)


cosunae:
launch pipeline

src/dawn/SIR/ASTExpr.h
```diff
@@ -402,6 +402,8 @@ class VarAccessExpr : public Expr {
   const std::string& getName() const { return name_; }
   std::string& getName() { return name_; }
 
+  void setName(std::string name) { name_ = name; }
```
twicki: if we have a setter, we should avoid the non-const getter

cosunae: good point. Done

PR #160, opened 2018-12-19
sort of generated code

twicki:
## Technical Description
we want the code-generation to be reporducible and therefore need the iterators to be over sorted containers. For this to happen we sort the unordered containers used in the compiler to ordered ones in the code-generation where sorting is based on field-id's.

### DIFF
--- a/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
+++ b/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
@@ -164,6 +164,15 @@ void CXXNaiveCodeGen::generateStencilWrapperCtr(
       continue;
 
     const auto& StencilFields = stencil.getFields();
+    std::vector<std::pair<int, iir::Stencil::FieldInfo>> sortedStencilFields;
+    for(auto stencilField : StencilFields) {
+      sortedStencilFields.push_back(stencilField);
+    }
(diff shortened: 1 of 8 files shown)


twicki:
launch pipeline

src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
```diff
@@ -160,14 +160,23 @@ void CXXNaiveCodeGen::generateStencilWrapperCtr(
       continue;
 
     const auto& StencilFields = stencil.getFields();
+    std::vector<std::pair<int, iir::Stencil::FieldInfo>> sortedStencilFields;
+    for(auto stencilField : StencilFields) {
```
cosunae: my main comment: 
the pattern here below is quite repetitive and explicit and low level that reduce the readability, everytime we create a copy of the container and then we sorted, and we keep in the scope both containers. 
I would suggest the following, replace it with a single helper function 
```
const auto StencilFields = order(stencil.getFields());
```
```
template<typename Key, typename Value>
std::map<Key, Value> order(std::unordered_map<Key,Value>& umap) {
   std::map<Key,Value> m;
   for(const auto& f: umap) m.insert(f); 
}
```
the std::map ensures that the elements are ordered, which is the requirement here

twicki: you are right, I was too deep down, simplified and reopened

PR #161, opened 2018-12-20
fix kcache

twicki:


src/dawn/CodeGen/Cuda/MSCodeGen.hpp
```diff
@@ -107,6 +107,10 @@ class MSCodeGen {
   static std::string kBegin(const std::string dom, iir::LoopOrderKind loopOrder,
                             iir::Interval const& interval);
 
+  bool intervalPreviouslyAccessed(const int accessID, iir::Interval const& targetInterval,
```
twicki: `T const&` again. I really think we should stick to one. 

If you feel like is is much more readable we can debate but since we once said we'd stick to the LLVM style, we should use `const T&` until we change that (or make an exception for this case). This is btw also the style that the cpp standard uses

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -193,9 +230,17 @@ void MSCodeGen::generatePreFillKCaches(
...
+                                            vertExtent.Minus, vertExtent.Plus)
+                            : iir::Interval(interval.lowerLevel(), interval.lowerLevel(),
+                                            vertExtent.Minus, vertExtent.Plus);
+
+    // if the bound of a kcgeneratePreFillKCachesache is the same as the interval, indicates that we
+    // will start using the
     // kcache for the first time with this k-leg, and a pre-fill might be required
-    if(cacheInterval.bound(intervalBound) == (interval.bound(intervalBound))) {
+    if(!intervalPreviouslyAccessed(accessID, prevInterval, interval, vertExtent)) {
+      //    if(cacheInterval.bound(intervalBound) == (interval.bound(intervalBound))) {
```
twicki: rm

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -255,7 +300,12 @@ void MSCodeGen::generateFillKCaches(MemberFunction& cudaKernel, const iir::Inter
 
     DAWN_ASSERT(cache.getInterval().is_initialized());
     const auto cacheInterval = *(cache.getInterval());
-    auto vertExtent = cacheProperties_.getKCacheVertExtent(accessID);
+    auto extents = ms_->computeExtents(accessID, interval);
```
twicki: what does getKCacheVertExtent do now? It's still used in two places (declaration and slide)

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -161,6 +161,43 @@ void MSCodeGen::generateKCacheFillStatement(MemberFunction& cudaKernel,
                           "] =" + ss.str());
 }
 
+bool MSCodeGen::intervalPreviouslyAccessed(const int accessID, iir::Interval const& targetInterval,
```
twicki: can we get 1 line of doc what this does to not have to read the method?

src/dawn/IIR/DoMethod.h
```diff
@@ -110,6 +110,8 @@ class DoMethod : public IIRNode<Stage, DoMethod, StatementAccessesPair> {
   /// The fields are computed during `DoMethod::update`.
   const std::unordered_map<int, Field>& getFields() const { return derivedInfo_.fields_; }
 
+  bool hasField(int accessID) const { return derivedInfo_.fields_.count(accessID); }
```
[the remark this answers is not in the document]
cosunae: I think count & find both have complexity
Constant on average, worst case linear in the size of the container. 
https://en.cppreference.com/w/cpp/container/unordered_map/count

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
cosunae: done

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
cosunae: done

cosunae:
launch pipeline

src/dawn/IIR/DoMethod.h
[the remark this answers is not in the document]
twicki: Interesting, I found thsese numbers in cpluplus:
http://www.cplusplus.com/reference/unordered_map/unordered_map/find/
http://www.cplusplus.com/reference/unordered_map/unordered_map/count/

Thoughts?

src/dawn/CodeGen/Cuda/MSCodeGen.hpp
twicki: :+1:

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
twicki: thanks for clarification

twicki:
is in 158 now

PR #165, opened 2019-01-21
Codegen sort

twicki:
## Technical Description
we want the code-generation to be reporducible and therefore need the iterators to be over sorted containers. For this to happen we sort the unordered containers used in the compiler to ordered ones in the code-generation where sorting is based on field-id's.

### DIFF
--- a/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
+++ b/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
@@ -159,7 +159,7 @@ void CXXNaiveCodeGen::generateStencilWrapperCtr(
     if(stencil.isEmpty())
       continue;
 
-    const auto& StencilFields = stencil.getFields();
+    const auto& StencilFields = orderMap(stencil.getFields());
 
     const std::string stencilName =
         codeGenProperties.getStencilName(StencilContext::SC_Stencil, stencil.getStencilID());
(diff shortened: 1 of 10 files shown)


src/dawn/CodeGen/CodeGen.h
```diff
@@ -25,6 +25,15 @@
 namespace dawn {
 namespace codegen {
 
+template <typename Key, typename Value>
+extern std::map<Key, Value> orderMap(const std::unordered_map<Key, Value>& umap) {
```
cosunae: why extern ?

src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
```diff
@@ -258,7 +258,7 @@ void CXXNaiveCodeGen::generateStencilClasses(
       continue;
 
     // fields used in the stencil
-    const auto& StencilFields = stencil.getFields();
+    const auto& StencilFields = orderMap(stencil.getFields());
```
cosunae: no impact in performance, nor what the compiler will do with it, but the `&` could be confusing for readability, since the function is constructing and returning a new object, instead of returning simply a reference. 
Wouldnt `const auto stencilFields` be more clear? (notice also `s` instead of `S`)

src/dawn/CodeGen/CodeGen.h
cosunae: ok

src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
cosunae: good, then that is all, if you remove the & in all these cases, and all testing in green go ahead with merge

twicki:
launch pipeline

PR #168, opened 2019-01-28
Launch bounds ii

cosunae:


cosunae:
launch pipeline

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -25,10 +26,11 @@
 namespace dawn {
 namespace codegen {
 namespace cuda {
-MSCodeGen::MSCodeGen(std::stringstream& ss, const std::unique_ptr<iir::MultiStage>& ms,
+MSCodeGen::MSCodeGen(std::stringstream& ss, OptimizerContext* context,
+                     const std::unique_ptr<iir::MultiStage>& ms,
                      const std::shared_ptr<iir::StencilInstantiation>& stencilInstantiation,
                      const CacheProperties& cacheProperties)
-    : ss_(ss), ms_(ms), stencilInstantiation_(stencilInstantiation),
+    : ss_(ss), context_(context), ms_(ms), stencilInstantiation_(stencilInstantiation),
```
twicki: the MS already holds the `stencilInstantiation`, we can fetch the context from there

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -717,6 +719,40 @@ void MSCodeGen::generateCudaKernelCode() {
...
+
+  int maxThreadsPerBlock =
+      blockSize_[0] * (blockSize_[1] + maxExtents[1].Plus - maxExtents[1].Minus +
+                       (maxExtents[0].Minus < 0 ? 1 : 0) + (maxExtents[0].Plus > 0 ? 1 : 0));
+
+  std::string arch = context_->getOptions().arch;
+  std::string domain_size = context_->getOptions().domain_size;
+  // if we have information about the architecture and the domain sizes we can further specify the
+  // minimum number of blocks per SM of the GPU
+  if((arch == "P100" || arch == "K40" || arch == "K80") && !domain_size.empty()) {
```
twicki: should we not just be able to pass the SM's as an option-argument? This way we're less restricted

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -717,6 +719,40 @@ void MSCodeGen::generateCudaKernelCode() {
...
+    std::istringstream idomain_size(domain_size);
+    std::string arg;
+    getline(idomain_size, arg, ',');
+    int isize = std::stoi(arg);
+    getline(idomain_size, arg, ',');
+    int jsize = std::stoi(arg);
+
+    int nSM = -1;
+    if(context_->getOptions().arch == "P100") {
```
twicki:
```suggestion
    if(arch == "P100") {
```

src/dawn/CodeGen/Cuda/MSCodeGen.hpp
```diff
@@ -46,6 +46,7 @@ class MSCodeGen {
 
 private:
   std::stringstream& ss_;
+  OptimizerContext* context_;
```
twicki: rm

src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -165,6 +166,7 @@ std::unique_ptr<OptimizerContext> DawnCompiler::runOptimizer(std::shared_ptr<SIR
   optimizer->checkAndPushBack<PassSetBoundaryCondition>();
   optimizer->checkAndPushBack<PassDataLocalityMetric>();
   optimizer->checkAndPushBack<PassSetSyncStage>();
+  optimizer->checkAndPushBack<PassSetBlockSize>();
```
twicki: I would prefe it to be before DataLocalityMetric

src/dawn/Compiler/Options.inc
```diff
@@ -40,6 +40,14 @@ OPT(std::string, Backend, "gridtools", "backend", "",
...
     "\n - c++-opt     = optimized C++ code"
     "\n - cuda        = optimized cuda", "<backend>", true, false)
+OPT(std::string, arch, "", "arch", "",
+    "Set the computing architecture. Possible values for cuda backend are:"
+    "\n - K80\n"
+    "\n - P100\n", "<arch>", true, false)
+OPT(std::string, domain_size, "", "domain-size", "",
+    "domain size for compiler optimization", "", true, false)
+OPT(std::string, block_size, "", "block-size", "",
+        "block size for compiler optimization", "", true, false)
```
twicki: non-specific comment. Specify what the meaning of the block-size is

src/dawn/Compiler/Options.inc
```diff
@@ -40,6 +40,14 @@ OPT(std::string, Backend, "gridtools", "backend", "",
     "\n - c++-naive   = naive C++ code\n"
     "\n - c++-opt     = optimized C++ code"
     "\n - cuda        = optimized cuda", "<backend>", true, false)
+OPT(std::string, arch, "", "arch", "",
+    "Set the computing architecture. Possible values for cuda backend are:"
+    "\n - K80\n"
+    "\n - P100\n", "<arch>", true, false)
```
twicki: you have less options in here then you support. Again, I suggest to drop it allthogether

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -717,6 +719,40 @@ void MSCodeGen::generateCudaKernelCode() {
...
+
+    int minBlocksPerSM = isize * jsize / (blockSize_[0] * blockSize_[1]);
+    if(solveKLoopInParallel_)
+      minBlocksPerSM *= 80 / blockSize_[2];
+    minBlocksPerSM /= nSM;
+
+    fnDecl = fnDecl + " __launch_bounds__(" + std::to_string(maxThreadsPerBlock) + "," +
+             std::to_string(minBlocksPerSM) + ") ";
+  } else {
+    fnDecl = fnDecl + " __launch_bounds__(" + std::to_string(maxThreadsPerBlock) + ") ";
```
twicki: if we keep it we want to exit with an illegal architecture and do this if nothing is specified

src/dawn/IIR/IIR.h
```diff
@@ -25,7 +25,7 @@ namespace iir {
 /// @ingroup optimizer
 class IIR : public IIRNode<void, IIR, Stencil> {
 
-  const std::array<unsigned int, 3> blockSize_ = {{32, 4, 4}};
+  std::array<unsigned int, 3> blockSize_ = {{32, 4, 4}};
```
twicki: this should be derived info, right? (since we're deriving it in `passSetBlockSize`)

src/dawn/Optimizer/PassSetBlockSize.cpp
```diff
@@ -0,0 +1,72 @@
...
+//===------------------------------------------------------------------------------------------===//
+
+#include "dawn/Optimizer/PassSetBlockSize.h"
+#include "dawn/IIR/IIRNodeIterator.h"
+#include "dawn/Optimizer/OptimizerContext.h"
+
+namespace dawn {
+
+PassSetBlockSize::PassSetBlockSize() : Pass("PassSetBlockSize") {
+  dependencies_.push_back("PassSetStageName");
```
twicki: why?

src/dawn/Optimizer/PassSetBlockSize.cpp
```diff
@@ -0,0 +1,72 @@
...
+        const auto& field = fieldP.second;
+
+        auto jExtents = field.field.getExtentsRB()[1];
+        if(jExtents.Plus != 0 || jExtents.Minus != 0) {
+          verticalPattern = false;
+        }
+      }
+    }
+
+    if(verticalPattern) {
```
twicki: can we have a one-liner on the choice?

src/dawn/Optimizer/PassSetBlockSize.cpp
```diff
@@ -0,0 +1,72 @@
...
+bool PassSetBlockSize::run(const std::shared_ptr<iir::StencilInstantiation>& stencilInstantiation) {
+  OptimizerContext* context = stencilInstantiation->getOptimizerContext();
+
+  const auto& IIR = stencilInstantiation->getIIR();
+
+  std::array<unsigned int, 3> blockSize{0, 0, 0};
+  if(!context->getOptions().block_size.empty()) {
+    std::string blockSizeStr = context->getOptions().block_size;
+    std::istringstream idomain_size(blockSizeStr);
+    std::string arg;
```
twicki: can we check that the argument is properly formatted here?

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
cosunae: ok

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
cosunae: I am not sure, yes, maybe, although this would leave all responsibility to the user in setting the right number...

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
cosunae: the idea here was not to exit, but rather to not specify the min number of blocks per sm if we dont know how many sms we have. That should still be allowed

src/dawn/CodeGen/Cuda/MSCodeGen.hpp
cosunae: ok

src/dawn/Compiler/DawnCompiler.cpp
cosunae: ok

src/dawn/IIR/IIR.h
cosunae: yes and no, the pass is computing a default (in case the user does not specify anything), but in the future we should be able to not want this and set it in the IIR ourselves (similar argument to caches). The information is computed using a strategy but not really derived deterministicylly from the information in the IIR (is more like a choice)

src/dawn/Optimizer/PassSetBlockSize.cpp
cosunae: mistake

src/dawn/Optimizer/PassSetBlockSize.cpp
cosunae: I didnt understand this, how would you build here a one-liner?

src/dawn/Compiler/Options.inc
cosunae: done

src/dawn/Optimizer/PassSetBlockSize.cpp
cosunae: stoi is throwing if conversion is not possible

cosunae: stoi is throwing if conversion is not possible

src/dawn/IIR/IIR.h
twicki: I see your point to some extent. Thought I would argue that all the information that is generated in the optimizer-process in dawn should be derived and everything that we get from the HIR should be non-derived. If you derive the info from a command-line argument or via smart algorithmic decisions is not really relevant is it?

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
twicki: Alright. I would advise to at least print it into the log though. If we have a typo in a script and are not getting the optimal block-size silently it might be a hassle

src/dawn/Optimizer/PassSetBlockSize.cpp
twicki: I meant can we have a line of comments describing why 32,1,4 vs 32,4,4

src/dawn/Optimizer/PassSetBlockSize.cpp
twicki: so we only need to check that the string is empy after the last getline

src/dawn/IIR/IIR.h
cosunae: not really, plenty of information encoded in the IIR is a result of optimizers and is not derived. Examples are the way the computations are structured in stages, ms, etc. caches, computations on the fly or not, etc. 

What is derived is information that is redundant and can be derived from IIR (itself) without algorithmic choices ?

src/dawn/Optimizer/PassSetBlockSize.cpp
cosunae: ok

src/dawn/Optimizer/PassSetBlockSize.cpp
cosunae: ok, yes

src/dawn/CodeGen/Cuda/MSCodeGen.cpp
cosunae: I am not sure where to put this, we dont have a well structured logging (only the INFO is working). And I am not sure you want always to print a message that for most people here wont mean anything. 
In many cases you will compile an executable for which the domain sizes are not known, so it is not possible to fully specify the launch bounds, and that should be ok, performance would not be bad

cosunae:
launch pipeline

cosunae:
launch pipeline

PR #174, opened 2019-02-18
Fix dependencies

cosunae:
The dependencies with proto files didnt work in my system with a newer cmake. 
It seems to be related to the propagation with the dependencies on objects targets. 
Here I moved the affected libraries to static libraries rather than objects

### DIFF
--- a/bundle/CMakeLists.txt
+++ b/bundle/CMakeLists.txt
@@ -39,7 +39,7 @@ include(DawnCloneRepository)
 
 set(DAWN_YODA_GIT_URL "https:///github.com/Meteoswiss-APN/yoda.git"
     CACHE PATH "URL of the dawn git repository to clone")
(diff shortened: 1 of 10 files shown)


bundle/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ include(DawnCloneRepository)
 
 set(DAWN_YODA_GIT_URL "https:///github.com/Meteoswiss-APN/yoda.git"
     CACHE PATH "URL of the dawn git repository to clone")
-set(DAWN_YODA_GIT_BRANCH "1.0.0" CACHE STRING "Branch of the dawn git repository to clone")
+set(DAWN_YODA_GIT_BRANCH "1.0.1" CACHE STRING "Branch of the dawn git repository to clone")
```
twicki: can you show me what this does?

cosunae: it added the `EXPORT_GROUP` so that  that put all the libraries under the same target group (easier to carry dependencies)

PR #176, opened 2019-02-24
add option to disable rbw

cosunae:


twicki:
why would we want this?

twicki:
(also this is a branch on the MCH-repo)

PR #178, opened 2019-02-26
dump

cosunae:
* provide a full dump of the IIR for verbose compilation mode
(current dump is very limited and broken for several examples)
* new possibility to dump from any level of IIR
* Cleanup quite some code (see Accesses.cpp)
* Remove the json dump (to avoid having multiple versions of the same thing)


cosunae:
launch pipeline

src/dawn/Compiler/Options.inc
```diff
@@ -61,8 +60,8 @@ OPT(std::string, ReorderStrategy, "greedy", "reorder", "",
     "\n - scut   = Use S-cut graph partitioning\n", "<strategy>", true, false)
 OPT(int, MaxFieldsPerStencil, 40, "max-fields", "",
     "Set the maximum number of fields in any given stencils", "<N>", true, false)
-
-
+OPT(bool, PassVerbose, false, "pass-verbose", "",
```
twicki: what is the best way of doing this?

We currently have quite a mess with our options as we have this as a option for all passes, then we have individual dumps and runtime arguments for some passes. Also I'm not sure if we're really consistent in the use of the F_GROUP in here. Should we clean this up now?

src/dawn/Unittest/PrintAllExpressionTypes.cpp
```diff
@@ -1,112 +0,0 @@
-//===--------------------------------------------------------------------------------*- C++ -*-===//
```
twicki: why remove this?

cosunae: because the idea is to always call this json dumper in debug mode from gtclang. So that we always call the path of code related to that. Then we dont need these tests anymore. They are not even running (to my understanding), just compiling the visitor...

src/dawn/Compiler/Options.inc
twicki: can you open up a task then?

src/dawn/Unittest/PrintAllExpressionTypes.cpp
twicki: Upon more detailed looks at it, this piece seems to be completely unused. Do you agree? If so I'm fine removing it

twicki:
looks good to me, can be merged once tests are passing

twicki:
launch pipeline

PR #179, opened 2019-03-01
implementation of the two missing comments

twicki:
coming from unresoved issues in 172 that were overlooked


twicki:
launch pipeline

Issue #180, opened 2019-03-06
verbose dump


twicki:
## Issue
the -fpass-verbose currently adds the multistages as keys to the stencil. For iteration purposes it would be better to have one dict mulitstages that holds them. Same for lower levels of the tree.

## Additional information
The discussion led to the conclusion that we should have everything that does not require ordering (fields, caches, etc...) in dictionaries and everything that has an order in arrays:
```
{
  "Stencil": {
    "Fields": {
      "Field1": {
        "name": "b"
      },
      "Field2": {
        "name": "a"
      }
    },
    "MSS": [
      "ms1",
      "ms0",
      "ms2"
    ]
  }
}
```

## Example
```
// RUN: %gtclang% %file% -fpass-verbose

#include "gridtools/clang_dsl.hpp"
using namespace gridtools::clang;

stencil test {
  storage a, b;
  Do {
    vertical_region(k_start, k_end) {
      // copy stencil
      a = b;
    }
  }
};
````

cosunae:
the -fpass-verbose crashes on horizontal_advection_wwcon (because of stencil functions). 
No worth fixing this until we fix the design of the metadata and IIR

cosunae:
The ordering problem of the fields is solved in #184 
But we also wanted that multistages are a collection (instead of dict)
We will do that after the redesign, #184 is only a bugfix

PR #184, opened 2019-03-13
fix ordering problem of fields described in #180

cosunae:
This is a bugfix as described in #180.


cosunae:
launch pipeline

twicki:
looks good but is, as you mentioned only one part of #180. I would suggest updating #180  to now only make sure that the tree (where order is required) is in lists moving forward.

twicki:
comment in #180 was made, ready to merge

PR #185, opened 2019-03-15
Fix tmp idx

cosunae:
Bug fix for use of temporary indices

- in code gen there is a logic to use temporaries or not. Even if the field is a temporary, for example if it is local cached, we dont generate temporaries. Also in the case there are no redundant computations, the temporaries become normal fields. However the logic was applied on a per ms level, which is wrong, since the temporaries are used across multiple MS's. 
We need to check if the logic for generating temporaries at the stencil level (not MS)

Additionally there is a bug fix in the launch_bounds, which could set the minBlocksPerSM beyond the maximum allowed by the architecture. ptx is ignoring this, so it has no effect, but we are fixing it here anyhow


cosunae:
launch jenkins

PR #189, opened 2019-03-22
properly format our source code

twicki:
## Technical Description

We've been sloppy with the last few commits to master, here is the full repo just running clang-format on it

#### Enhances

Cleans up the repo

## Dependencies

standalone


twicki:
launch pipeline

cosunae:
yes, two comments: 
- we should switch the clang format to 6.0
- could we do that after #183  ? Since we delayed #183 and several PRs are already jumping ahead, with a formatting PR it will be even more difficult to recover the work there.

PR #191, opened 2019-04-03
Java SIR

cosunae:


twicki:
how do we proceed here?

cosunae:
I think we first need to go for #193

PR #187, opened 2019-03-21
Metadata fix fv

twicki:
## Technical Description

- Field-Versioning had a very odd way of storing things that made the fix of https://github.com/MeteoSwiss-APN/dawn/issues/121 impossible
- In order to 
- Accesses to field versioning were sub-optimally handled, there was too much public access given

## Dependencies
- This changes the indexing to be 0-based instead of 1-based and therefore need to go with https://github.com/MeteoSwiss-APN/gtclang/pull/122

PR #204, opened 2019-05-29
Fix fieldversioning

twicki:
## Technical Description

- Field-Versioning had a very odd way of storing things that made the fix of https://github.com/MeteoSwiss-APN/dawn/issues/121 impossible
- This was already looked at in  https://github.com/MeteoSwiss-APN/dawn/pull/187 but is now reworked here

## Dependencies
- This changes the indexing to be 0-based instead of 1-based and therefore need to go with https://github.com/MeteoSwiss-APN/gtclang/pull/134

## Testing
Since the indexing changes to be 0 based instead of 1 based, we need to adapt some reference files and the testing is in https://github.com/MeteoSwiss-APN/gtclang/pull/134

### DIFF
--- a/src/dawn/IIR/FieldAccessMetadata.h
+++ b/src/dawn/IIR/FieldAccessMetadata.h
@@ -17,7 +17,6 @@
 
 #include "boost/variant.hpp"
 #include "dawn/Support/Assert.h"
-#include "dawn/Support/Assert.h"
 #include "dawn/Support/Json.h"
 #include <set>
 #include <unordered_map>
--- a/src/dawn/IIR/FieldAccessMetadata.h
+++ b/src/dawn/IIR/FieldAccessMetadata.h
@@ -17,6 +17,7 @@
 
 #include "boost/variant.hpp"
 #include "dawn/Support/Assert.h"
+#include "dawn/Support/Assert.h"
 #include "dawn/Support/Json.h"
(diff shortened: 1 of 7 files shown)


src/dawn/IIR/FieldAccessMetadata.h
```diff
@@ -26,38 +28,122 @@ namespace iir {
...
 
   json::json jsonDump() const;
+
+  // now that we only have getters and setters this should always be in a consistent state.
+  // Should we remove this enirely, have it here as a safety valve if one wants to mess with the
+  // versioning and needs a clean update or have this as part of the update-level step for the
+  // top-most level holding the metadata?
+  // I personally think option 1 should be desired as this is (hopefully) never necessary but I've
+  // put it here if the reviewers think this would be helpful
+  void recomputeDerivedInfo() {
```
Stagno: I think this has now no use, so it should be removed. 
As a side note, having derived data structures makes me think that maybe a Design by Contract framework, where things like this are checked in an invariant expression, could be helpful, but it may be too much of a long step.

Stagno:
Whether or not member `recomputeDerivedInfo()` is going to be removed, once all tests pass on the related gtclang PR, this PR is ready to be merged

twicki:
launch pipeline

PR #205, opened 2019-05-31
Fix Algorithm for FieldVersioning

twicki:
## Technical Description

This fixes the fact that if input fields are versioned, the versioned fields are not synchronized before being used

#### Resolves

https://github.com/MeteoSwiss-APN/dawn/issues/121

#### Example

```
stencil test {
  storage a;
  Do {
        a += a[i + 1];
    }
  }
};
```

## Testing

Perftests launched in https://github.com/MeteoSwiss-APN/clang-gridtools/pull/113

## Dependencies

This PR is independent.


twicki:
launch pipeline

src/dawn/Optimizer/PassFieldVersioning.cpp
```diff
@@ -89,71 +91,165 @@ static void reportRaceCondition(const Statement& statement,
...
+    newAccess->addReadExtent(assigneeID, iir::Extents(Array3i{{0, 0, 0}}));
+  }
+  pair->setAccesses(newAccess);
+  domethod->insertChild(std::move(pair));
+
+  // Add the new expressions to the map
+  context->getMetaData().insertExprToAccessID(fa_assignment, assignmentID);
+  context->getMetaData().insertExprToAccessID(fa_assignee, assigneeID);
+}
+/// @brief Creates the stage in which assignment happens (fill and flush)
```
cosunae: fill and flush are concepts of kcaches, so it is confusing to me to use it here in the doc

src/dawn/Optimizer/PassFieldVersioning.cpp
```diff
@@ -89,71 +91,165 @@ static void reportRaceCondition(const Statement& statement,
...
+          for(auto interval : multiInterval.getIntervals()) {
+
+            // we create the new stage that holds these do-methods
+            auto insertedStage = createAssignmentStage(
+                interval, id, stencilInstantiation->getMetaData().getOriginalVersionOfAccessID(id),
+                stencilInstantiation, extents);
+            // and insert them at the beginnning of the MultiStage
+            mss->insertChild(mss->childrenBegin(), std::move(insertedStage));
+          }
+          // update the mss: #TODO: this is still a workaround since we don't have level-and below:
```
cosunae: I dont understand the comment, in principle the derived info of node levels below does not depend on parent. So we dont need an update of level and below, isnt?

src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -179,6 +180,8 @@ std::unique_ptr<OptimizerContext> DawnCompiler::runOptimizer(std::shared_ptr<SIR
   optimizer->checkAndPushBack<PassStencilSplitter>(maxFields);
   optimizer->checkAndPushBack<PassTemporaryType>();
   optimizer->checkAndPushBack<PassTemporaryMerger>();
+  optimizer->checkAndPushBack<PassFieldVersioning>(
+      PassFieldVersioning::FieldVersioningPassMode::FM_FixAccess);
```
cosunae: To me, even if related to field versioning, the create version and fix access are different passes. They take different input of the IIR and they do different transformations. I propose to put the FixAccess mode as a separate pass (DoubleBufferCopyOfFieldVersions)

src/dawn/Optimizer/PassFieldVersioning.cpp
```diff
@@ -89,71 +91,165 @@ static void reportRaceCondition(const Statement& statement,
...
-                                     loopOrder, stageIdx, stmtIndex);
-
-          if(rc == RCKind::RK_Unresolvable) {
-            // Nothing we can do ... bail out
-            return false;
-          } else if(rc == RCKind::RK_Fixed) {
-            // We fixed a race condition (this means some fields have changed and our current graph
-            // is invalid)
-            newGraph = oldGraph;
+  if(mode_ == FieldVersioningPassMode::FM_CreateVersion) {
```
cosunae: As said, I would propose to move it into separate pass and as our convention document the algorithm (Input/Output and what transformation applies to IIR)

src/dawn/Optimizer/PassFieldVersioning.cpp
```diff
@@ -89,71 +91,165 @@ static void reportRaceCondition(const Statement& statement,
   }
 }
 
-} // anonymous namespace
+static void addAssignmentToDoMethod(std::unique_ptr<iir::DoMethod>& domethod, int assignmentID,
+                                    int assigneeID,
+                                    const std::shared_ptr<iir::StencilInstantiation>& context,
+                                    boost::optional<iir::Extents>& readExtents) {
```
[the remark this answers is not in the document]
twicki: you're correct. I was thinking that we need to pass the RB-Extents but they are calculated later.

src/dawn/Optimizer/PassFieldVersioning.cpp
```diff
@@ -89,71 +91,165 @@ static void reportRaceCondition(const Statement& statement,
   }
 }
 
-} // anonymous namespace
+static void addAssignmentToDoMethod(std::unique_ptr<iir::DoMethod>& domethod, int assignmentID,
```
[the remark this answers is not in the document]
twicki: done

src/dawn/Optimizer/PassFieldVersioning.cpp
twicki: changed the wording

src/dawn/Optimizer/PassFieldVersioning.cpp
twicki: In this specific case we modify the children of an instance. Now the ideal thing would be:
- Update all the Do-Methods
- Update all the above (once)

Now we have level-and-tree above for Do-Methods but this recomputes the the derived info multiple times, doesn't it?

So with the current tools the perfect way would be:
iteateIIROver<DoMethod>
  update(level)
iteateIIROver<Stage>
  update(level)
MS->update(levelAndTreeAbove)

now the question is if we want to wrap this into a function (like update-range (From DoMethod to MS) or have one that does MS and below (which does level on every child so we do it only once).

Thoughts?

src/dawn/Compiler/DawnCompiler.cpp
twicki: let's briefly discuss in person

src/dawn/Optimizer/PassFieldVersioning.cpp
```diff
@@ -89,71 +91,165 @@ static void reportRaceCondition(const Statement& statement,
...
+                                    int assigneeID,
+                                    const std::shared_ptr<iir::StencilInstantiation>& context,
+                                    boost::optional<iir::Extents>& readExtents) {
+  // Create the StatementAccessPair of the assignment with the new and old variables
+  auto fa_assignee =
+      std::make_shared<FieldAccessExpr>(context->getMetaData().getNameFromAccessID(assigneeID));
```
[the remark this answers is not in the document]
twicki: This is the case for every one of the >25 calls to `insertExprToAccessID`. Is this functionality generally flawed?

twicki:
launch pipeline

src/dawn/Optimizer/PassFieldVersioning.cpp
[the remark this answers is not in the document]
cosunae: not always, isnt? In this example
https://github.com/MeteoSwiss-APN/dawn/blob/master/src/dawn/Optimizer/StatementMapper.cpp#L348
we mapped an existing expr, we dont create it. 
I think we would need the two levels, but for cases where we create the shared_ptr<Expr> and insert the mapping it would make sense to have it in a maker. It follows the ideas of the last big refactoring to always have the IIR and metadata in consistent state, we just didnt do it yet. 

Actually generalizing the idea, instead of a makeFieldAccessExpr, we could have a 
```
template<typename ExprType, typename ... Args>
makeExpr(Args && ... args)
```

src/dawn/Optimizer/PassFieldVersioning.cpp
cosunae: yeah, ok I see, 
the API was not designed for updating efficiently when all the leaves are updated, 
ok, I would leave it like that then

src/dawn/Optimizer/PassFieldVersioning.cpp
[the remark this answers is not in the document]
twicki: sorry, I was unclear. What I meant is that we enter this state of inconsistency whenever we create a new Expr / Stmt before we do the mapping. This happens in various parts of the code.

I think the generalization makes sense and that way we can have a register function that handles all the things that need to be handled (filling this map, filling `accessIfTpye`, potential future thinigs) in a centralized place. I would do this in a separate PR though. Does that make sense?

twicki:
launch pipeline

twicki:
unfortunately, a wrong behavior in bc-extent computation was discovered with the new stages that are added. This is also fixed here

twicki:
launch pipeline

twicki:
launch pipeline

twicki:
launch pipeline

twicki:
launch pipeline

twicki:
this needs major rework

PR #206, opened 2019-06-05
Add diagnostic error for unsupported Read-before-Write conflict inside if block

### LINKED ISSUE #190: bug in fastwaves DppDz
Here https://github.com/cosunae/clang-gridtools/blob/fast_waves_sc/src/dycore/fast_waves_sc.cpp#L387

The computation of `sum(i + 1, DppDzS)` is moved before the comp of DppDzS

```
     const gridtools::clang::float_type __local_sum_6727 =
          (__tmp_DppDzS_3726_ijcache[ijcacheindex + 1] + __tmp_DppDzS_3726_ijcache[ijcacheindex]);
      if (((bool)false == (int)0)) {
        __tmp_DppDzS_3726_ijcache[ijcacheindex] =
            (((wgtOneSided_kcache[2] * __ldg(&(__tmp_ppMinusDiv_3727[idx111]))) +
              (wgtOneSided_kcache[1] * __ldg(&(__tmp_ppMinusDiv_3727[idx111 + stride_111_2 * -1])))) +
             (wgtOneSided_kcache[0] * __ldg(&(__tmp_ppMinusDiv_3727[idx111 + stride_111_2 * -2]))));
        const gridtools::clang::float_type __local_dp_dx_rho_4363 =
            (__ldg(&(frac1RhoDx[idx111])) *
             (__local_delta_6724 - (__ldg(&(dz_lam_u_half[idx111])) * __local_sum_6727)));
        u[idx111] =
            (__ldg(&(u_pos[idx111])) + (globals_.dts * (__ldg(&(du_dt_slow[idx111])) - __local_dp_dx_rho_4363)));
      }
```
but furthermore there is a missing syncthreads between computing DppDz and consuming it

Stagno:
## Technical Description

Since, for now, the stage splitting inside an if block doesn't work, this code adds a check for cases in which this would happen, reports a diagnostic message and returns error.
Also the `oldGraph` uninitialized pointer bug is fixed here. However the code that dumps dependency graphs for each split should be refactored in a different PR.

#### Enhances

#190 

#### Example
```

globals {
  bool x = false;
};

stencil_function sum {
  offset off;
  storage data;

  Do { return data[off] + data; }
};

stencil test {

  storage b,c,d;
  var z, a;

  Do {
    vertical_region(k_end, k_end) {
      if(x == 0) {
        z = a * b;
        const double e = d * (c * sum(i + 1, z));
      }
    }
  }
};
```

### DIFF
--- a/src/dawn/Optimizer/PassStageSplitter.cpp
+++ b/src/dawn/Optimizer/PassStageSplitter.cpp
@@ -28,8 +28,6 @@
 
 namespace dawn {
 
-namespace {
-
 static void reportDataDependencyInsideConditionalBlock(
     const std::shared_ptr<Statement>& statement,
     const std::shared_ptr<iir::StencilInstantiation>& instantiation) {
@@ -38,8 +36,6 @@ static void reportDataDependencyInsideConditionalBlock(
   diag << "Read-before-Write conflict inside conditional block is not supported.";
   instantiation->getOptimizerContext()->getDiagnostics().report(diag);
 }
-
-} // anonymous namespace
 
 PassStageSplitter::PassStageSplitter() : Pass("PassStageSplitter", true) {}
 
--- a/src/dawn/Optimizer/PassStageSplitter.cpp
+++ b/src/dawn/Optimizer/PassStageSplitter.cpp
@@ -28,15 +28,6 @@
 
 namespace dawn {
 
-static void reportDataDependencyInsideConditionalBlock(
-    const std::shared_ptr<Statement>& statement,
-    const std::shared_ptr<iir::StencilInstantiation>& instantiation) {
-
-  DiagnosticsBuilder diag(DiagnosticsKind::Error, statement->ASTStmt->getSourceLocation());
-  diag << "Read-before-Write conflict inside conditional block is not supported.";
-  instantiation->getOptimizerContext()->getDiagnostics().report(diag);
-}
-
 PassStageSplitter::PassStageSplitter() : Pass("PassStageSplitter", true) {}
 
 bool PassStageSplitter::run(
@@ -87,8 +78,11 @@ bool PassStageSplitter::run(
               if(hasHorizontalReadBeforeWriteConflict(&conditionalBlockGraph)) {
                 // Since splitting inside a conditional block is not supported, report and return an
                 // error.
-                reportDataDependencyInsideConditionalBlock(stmtAccessesPair->getStatement(),
-                                                           stencilInstantiation);
+                auto statement = stmtAccessesPair->getStatement();
+                DiagnosticsBuilder diag(DiagnosticsKind::Error,
+                                        statement->ASTStmt->getSourceLocation());
+                diag << "Read-before-Write conflict inside conditional block is not supported.";
+                stencilInstantiation->getOptimizerContext()->getDiagnostics().report(diag);
                 return false;
               }
             }
--- a/src/dawn/Optimizer/PassStageSplitter.cpp
+++ b/src/dawn/Optimizer/PassStageSplitter.cpp
@@ -28,6 +28,19 @@
 
 namespace dawn {
 
+namespace {
+
+static void reportDataDependencyInsideConditionalBlock(
+    const std::shared_ptr<Statement>& statement,
+    const std::shared_ptr<iir::StencilInstantiation>& instantiation) {
+
+  DiagnosticsBuilder diag(DiagnosticsKind::Error, statement->ASTStmt->getSourceLocation());
+  diag << "Read-before-Write conflict inside conditional block is not supported.";
+  instantiation->getOptimizerContext()->getDiagnostics().report(diag);
+}
+
+} // anonymous namespace
+
 PassStageSplitter::PassStageSplitter() : Pass("PassStageSplitter", true) {}
 
 bool PassStageSplitter::run(
@@ -57,6 +70,7 @@ bool PassStageSplitter::run(
         std::shared_ptr<iir::DependencyGraphAccesses> newGraph, oldGraph;
         newGraph =
             std::make_shared<iir::DependencyGraphAccesses>(stencilInstantiation->getMetaData());
+        oldGraph = newGraph->clone();
 
         // Build the Dependency graph (bottom to top)
         for(int stmtIndex = doMethod.getChildren().size() - 1; stmtIndex >= 0; --stmtIndex) {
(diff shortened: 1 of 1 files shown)


Stagno:
launch pipeline

src/dawn/Optimizer/PassStageSplitter.cpp
```diff
@@ -28,6 +28,19 @@
 
 namespace dawn {
 
+namespace {
+
+static void reportDataDependencyInsideConditionalBlock(
```
twicki: so in c++ static and the anonymous namespace do exactly the same thing if used on a function.

the rule of thumb is to keep anonymous namespaces as small as possible since it's really tedious to figure out if functions are inside one on not. use them when you need to declare static classes or structs like here
https://github.com/MeteoSwiss-APN/dawn/blob/0842c266d23eada1a47a9140834e2f1564b12003/src/dawn/IIR/DependencyGraphAccesses.cpp#L273
but use the static keyword on functions if possible (as it is more explicit since it's next to the function and not (potentially) several lines up.

src/dawn/Optimizer/PassStageSplitter.cpp
```diff
@@ -28,6 +28,19 @@
...
+    const std::shared_ptr<Statement>& statement,
+    const std::shared_ptr<iir::StencilInstantiation>& instantiation) {
+
+  DiagnosticsBuilder diag(DiagnosticsKind::Error, statement->ASTStmt->getSourceLocation());
```
twicki: does it make sense to pack a 3-liner into a function that is only called once?

src/dawn/Optimizer/PassStageSplitter.cpp
```diff
@@ -57,6 +70,7 @@ bool PassStageSplitter::run(
         std::shared_ptr<iir::DependencyGraphAccesses> newGraph, oldGraph;
         newGraph =
             std::make_shared<iir::DependencyGraphAccesses>(stencilInstantiation->getMetaData());
+        oldGraph = newGraph->clone();
```
twicki: this changes what we're doing. I still think what should happen is that lines 71 to 73 should be removed and we should be in the clear. This is since these are just a copy/paste error and this is why we're getting the nullptr exception since 95-97 is overwriting what you dump in 71-73.

twicki:
I would argue that this does not solve the issue but enhances the diagnostics. We should still work on #190 but it is a good step in the right direction

src/dawn/Optimizer/PassStageSplitter.cpp
Stagno: The nullptr exception happens when a conflict is detected on the first iteration of this loop: https://github.com/MeteoSwiss-APN/dawn/blob/cc0613e43a3b6a77fee7de783c04ca113c407e70/src/dawn/Optimizer/PassStageSplitter.cpp#L76-L83
Even if the dumping code is removed, still a null pointer would be saved in the `graphs` container:
https://github.com/MeteoSwiss-APN/dawn/blob/cc0613e43a3b6a77fee7de783c04ca113c407e70/src/dawn/Optimizer/PassStageSplitter.cpp#L100-L106
`oldGraph` must be initialized somehow before being used and i thought that having it initialized as a clone of `newGraph` would keep the same semantics for the first iteration as for another iteration, as each iteration ends this way:
https://github.com/MeteoSwiss-APN/dawn/blob/cc0613e43a3b6a77fee7de783c04ca113c407e70/src/dawn/Optimizer/PassStageSplitter.cpp#L121

src/dawn/Optimizer/PassStageSplitter.cpp
Stagno: I was mirroring the same functional semantics separation that is also employed here:
https://github.com/MeteoSwiss-APN/dawn/blob/9513f9723c597042b1ca02d02280f8b7f9790514/src/dawn/Optimizer/PassFieldVersioning.cpp#L68-L90
Maybe more verbose diagnostic could be required and it would be ideal then to have already a function for the reporting purpose only.

src/dawn/Optimizer/PassStageSplitter.cpp
Stagno: Ok, i will only use the static keyword and remove the anonymous namespace.

src/dawn/Optimizer/PassStageSplitter.cpp
twicki: The difference is that this function is called more than once and it would have duplicated code to inline it whereas your call is only needed in one place.

src/dawn/Optimizer/PassStageSplitter.cpp
twicki: sounds reasonable!
Are we still removing the reduntand write to dot though?

src/dawn/Optimizer/PassStageSplitter.cpp
Stagno: Ok, I'm moving the code to the `run` method on the next commit.

Stagno:
launch pipeline

src/dawn/Optimizer/PassStageSplitter.cpp
Stagno: Would it be ok to do it in this PR? Or opening a new one would be a better choice?

twicki: you're already touching this file heavily, I'd do it here

Stagno: If I understand correctly what you propose is to remove the `oldGraph->toDot(...)` call inside the loop and only have `newGraph->toDot(...)` outside. However, in this way, we would only dump the dependency graph of the first stage (from the last split). Isn't the intended functionality to dump the dep. graphs of every stage generated by this pass?

twicki: yeah, seems like you're correct. I misread one line. We should definitely keep it

twicki:
did you see that we're triggering this in clang gridtools in the pipeline?

Stagno:
> did you see that we're triggering this in clang gridtools in the pipeline?

actually i didn't. But what PR of clang gridtools is interested?

Stagno:
launch pipeline

Stagno:
launch pipeline

Stagno:
@twicki Tests now pass. Should I merge?

PR #208, opened 2019-06-17
Fix json dump access by index bug

Stagno:
## Technical Description

Changes insertions into json object of the form `jsonObj[integerVar]` (which force the json object to be an array without `integerVar` necessarily being a valid array index) into `jsonObj[std::to_string(integerVar)]`.

#### Example

Crash when compiling the following code with ` -fpass-verbose`:
```

globals {
  bool x = false;
};

stencil_function sum {
  offset off;
  storage data;

  Do { return data[off] + data; }
};

stencil test {

  storage b,c,d;
  var z, a;

  Do {
    vertical_region(k_end, k_end) {
      if(x == 0) {
        a = 0;
        z = a * b;
        const double e = d * (c * sum(i + 1, z));
      }
    }
  }
};
```

### DIFF
--- a/src/dawn/IIR/MultiStage.cpp
+++ b/src/dawn/IIR/MultiStage.cpp
@@ -419,7 +419,7 @@ json::json MultiStage::jsonDump() const {
 
   json::json cachesJson;
   for(const auto& cache : derivedInfo_.caches_) {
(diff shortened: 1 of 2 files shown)


Stagno:
launch pipeline

PR #209, opened 2019-06-19
Fixes dash filling bug

Stagno:
## Technical Description

When `ReportDataLocalityMetric` is on, PassDataLocalityMetric run method was trying to create a padding string filled with '-' possibly giving a negative number as fill size. Fixed by imposing that the fill size must be non-negative.

#### Example

Bug was happening for long stencil names such as: HorizontalDiffusionColdPoolTPCopyStencil


Stagno:
launch pipeline

cosunae:
looks good to me. 
Kesch failed but related to the upgrade, so I think is ok

PR #215, opened 2019-06-24
make nodeiterator private

twicki:
## Technical Description

There was a default constructor setting it's const-ptr to a nullptr and should therefore never be used. This makes them private. (Solution by @Stagno)

#### Resolves / Enhances

Cleanup of the node iterator class.

## Dependencies

This PR is independent


twicki:
launch pipeline

twicki:
launch pipeline

twicki:
unsuccessful experiment

PR #223, opened 2019-08-07
Fix pytests

twicki:
## Technical Description

Changes the default behavior of the bundle to not run the python examples as the default is not to install the required python modules.

#### Resolves

This should fix the currently broken master builds of jenkins.
If the bundle downloads dawn, the default flags are off for python-bindings but on for python tests. So the (post)-install step fails. 

#### Example

http://jenkins-mch.cscs.ch/view/PASCHA/job/gtclang/

## Testing

The pipeline test does not work here since it works with a dawn that is installed on a previous step. So a "maunal" launch of the pipeline https://github.com/MeteoSwiss-APN/gtclang/pull/146

## Dependencies

This PR is indepdent


twicki:
launch jenkins

PR #224, opened 2019-08-07
disable python if protobuf python not found

cosunae:


lukasm91:
This doesn't fix the problem (and `DAWN_PYTHON` is already disabled in `python/CMakeLists.txt`)

cosunae:
right, can you help me once you are back? I can not reproduce the problem... so we can debug it in your environment

twicki:
This is solved here:
https://github.com/MeteoSwiss-APN/dawn/pull/249

PR #227, opened 2019-08-09
Fix stage merge

twicki:
## Technical Description

Moves the axis computation to the stencil-level so we're not computing illegal states

#### Resolves

Issue with the integration-tests when both dawn and gtclang are used in debug mode

#### Example

https://github.com/MeteoSwiss-APN/gtclang/blob/master/test/integration-test/CodeGen/intervals_stencil.cpp

## Dependencies

This PR is independent


twicki:
launch pipeline

PR #235, opened 2019-08-15
Namespace rename

mroethlin:
## Technical Description
 
This fixes the issues with the namespaces used in generated code. Namespaces like "cuda" or "gridtools" should be used only by cuda and gridtools themselves. Hence the namespaces in all code generated now have the suffix "_dawn" added

 #### Resolves
 
[Issue 144](https://github.com/MeteoSwiss-APN/gtclang/issues/144) reported in the gtclang repo 
 
 ## Dependencies
 
 This PR is independent.


lukasm91:
would it make sense to use `dawn::cxx_naive`?

twicki:
there was an unwanted delta, we'r reopening this on a clean branch

PR #240, opened 2019-08-19
Fix iir optimizer dependency

twicki:
## Technical Description

We remove all the backwards dependencies from Code-Gen to the Optimizer. This is part of cleaning up all unwanted (circular) dependencies between modules

####  Enhances

Now Code-Gen is indepdendent of OptimizerContext

## Testing

See https://github.com/MeteoSwiss-APN/gtclang/pull/148

## Dependencies

This requires the changes in GTClang seen in https://github.com/MeteoSwiss-APN/gtclang/pull/148

### DIFF
--- a/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
+++ b/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
@@ -557,8 +557,8 @@ std::unique_ptr<TranslationUnit> CXXNaiveCodeGen::generateCode() {
   CodeGen::addMplIfdefs(ppDefines, 30);
   DAWN_LOG(INFO) << "Done generating code";
 
-  return make_unique<TranslationUnit>(context_.begin()->second->getMetaData().getFileName(),
-                                      std::move(ppDefines), std::move(stencils),
+  std::string filename = generateFileName(context_);
+  return make_unique<TranslationUnit>(filename, std::move(ppDefines), std::move(stencils),
                                       std::move(globals));
 }
 
--- a/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
+++ b/src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
@@ -18,7 +18,6 @@
 #include "dawn/CodeGen/CXXUtil.h"
 #include "dawn/CodeGen/CodeGenProperties.h"
 #include "dawn/IIR/StencilInstantiation.h"
-#include "dawn/Optimizer/OptimizerContext.h"
 #include "dawn/SIR/SIR.h"
 #include "dawn/Support/Assert.h"
 #include "dawn/Support/Logging.h"
@@ -63,7 +62,10 @@ static std::string makeKLoop(const std::string dom, bool isBackward,
                     : makeLoopImpl(iir::Extent{}, "k", lower, upper, "<=", "++");
 }
 
-CXXNaiveCodeGen::CXXNaiveCodeGen(OptimizerContext* context) : CodeGen(context) {}
+CXXNaiveCodeGen::CXXNaiveCodeGen(
+    std::map<std::string, std::shared_ptr<iir::StencilInstantiation>>& ctx,
+    DiagnosticsEngine& engine, int maxHaloPoint)
+    : CodeGen(ctx, engine, maxHaloPoint) {}
 
 CXXNaiveCodeGen::~CXXNaiveCodeGen() {}
 
@@ -446,7 +448,7 @@ void CXXNaiveCodeGen::generateStencilFunctions(
                                 stencilInstantiation->getMetaData().getStencilLocation());
         diag << "no storages referenced in stencil '" << stencilInstantiation->getName()
              << "', this would result in invalid gridtools code";
-        context_->getDiagnostics().report(diag);
+        diagEngine.report(diag);
         return;
       }
 
@@ -466,7 +468,7 @@ void CXXNaiveCodeGen::generateStencilFunctions(
         DiagnosticsBuilder diag(DiagnosticsKind::Error, stencilFun->getStencilFunction()->Loc);
         diag << "no storages referenced in stencil function '" << stencilFun->getName()
              << "', this would result in invalid gridtools code";
-        context_->getDiagnostics().report(diag);
+        diagEngine.report(diag);
         return;
       }
 
@@ -531,14 +533,18 @@ std::unique_ptr<TranslationUnit> CXXNaiveCodeGen::generateCode() {
 
   // Generate code for StencilInstantiations
   std::map<std::string, std::string> stencils;
-  for(const auto& nameStencilCtxPair : context_->getStencilInstantiationMap()) {
+  for(const auto& nameStencilCtxPair : context) {
     std::string code = generateStencilInstantiation(nameStencilCtxPair.second);
     if(code.empty())
       return nullptr;
     stencils.emplace(nameStencilCtxPair.first, std::move(code));
   }
 
-  std::string globals = generateGlobals(context_->getSIR(), "cxxnaive");
+  std::string globals = "";
+  if(context.size() > 0) {
+    const auto& globalsMap = context.begin()->second->getIIR()->getGlobalVariableMap();
+    globals = generateGlobals(globalsMap, "cxxnaive");
+  }
 
   std::vector<std::string> ppDefines;
   auto makeDefine = [](std::string define, int value) {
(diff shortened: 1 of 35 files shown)


src/dawn/CodeGen/CodeGen.h
```diff
@@ -38,7 +39,11 @@ extern std::map<Key, Value> orderMap(const std::unordered_map<Key, Value>& umap)
 /// @ingroup codegen
 class CodeGen {
 protected:
-  OptimizerContext* context_;
+  std::map<std::string, std::shared_ptr<iir::StencilInstantiation>>& context;
```
havogt: maybe create an alias

src/dawn/CodeGen/Cuda/MSCodeGen.hpp
```diff
@@ -54,11 +55,12 @@ class MSCodeGen {
   std::string cudaKernelName_;
   Array3ui blockSize_;
   const bool solveKLoopInParallel_;
+  CudaCodeGen::CudaCodeGenOptions& options_;
```
havogt: const& or maybe not even a reference?

src/dawn/Optimizer/PassInlining.h
```diff
@@ -34,7 +34,7 @@ namespace dawn {
 class PassInlining : public Pass {
 public:
   /// @brief Inlining strategies
-  enum InlineStrategyKind {
+  enum class InlineStrategyKind {
     IK_InlineProcedures,    ///< Inline functions with no return
```
havogt: I'd remove `IK_` as it is obvious after the move to scoped enums.
My personal taste would also to rename `InlineStrategyKind` -> `InlineStrategy`.

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
```diff
@@ -34,7 +34,10 @@ namespace dawn {
 namespace codegen {
 namespace gt {
 
-GTCodeGen::GTCodeGen(OptimizerContext* context) : CodeGen(context), mplContainerMaxSize_(20) {}
+GTCodeGen::GTCodeGen(std::map<std::string, std::shared_ptr<iir::StencilInstantiation>>& ctx,
+                     DiagnosticsEngine& engine, bool useParallelEP, int maxHaloPoints)
+    : CodeGen(ctx, engine, maxHaloPoints), mplContainerMaxSize_(20), useParallelEP_(useParallelEP) {
```
havogt: Not sure if needed: For consistency you could move useParallelEP_ into an option struct as in the other generators.

src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -190,6 +191,10 @@ std::unique_ptr<OptimizerContext> DawnCompiler::runOptimizer(std::shared_ptr<SIR
   optimizer->checkAndPushBack<PassSetBlockSize>();
   optimizer->checkAndPushBack<PassDataLocalityMetric>();
   optimizer->checkAndPushBack<PassSetSyncStage>();
+  // If we want to generate cuda code or serialize the iir we inline all function calls
+  optimizer->checkAndPushBack<PassInlining>(
+      getOptions().Backend == "cuda" || getOptions().SerializeIIR,
```
havogt: For reference improve comment to include why cuda and serialization needs the inlining.

src/dawn/CodeGen/Cuda/MSCodeGen.hpp
lukasm91: I clearly vote for no reference as this object is rather small and it's just the configuration

src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
```diff
@@ -531,14 +533,18 @@ std::unique_ptr<TranslationUnit> CXXNaiveCodeGen::generateCode() {
...
     std::string code = generateStencilInstantiation(nameStencilCtxPair.second);
     if(code.empty())
       return nullptr;
     stencils.emplace(nameStencilCtxPair.first, std::move(code));
   }
 
-  std::string globals = generateGlobals(context_->getSIR(), "cxxnaive");
+  std::string globals = "";
+  if(context.size() > 0) {
+    const auto& globalsMap = context.begin()->second->getIIR()->getGlobalVariableMap();
```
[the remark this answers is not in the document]
twicki: Is it better to pass the full map to the function even though it only needs a tiny part of it though?

src/dawn/CodeGen/Cuda/CudaCodeGen.cpp
```diff
@@ -663,23 +665,23 @@ std::unique_ptr<TranslationUnit> CudaCodeGen::generateCode() {
...
-  for(const auto& nameStencilCtxPair : context_->getStencilInstantiationMap()) {
+  for(const auto& nameStencilCtxPair : context) {
     std::shared_ptr<iir::StencilInstantiation> origSI = nameStencilCtxPair.second;
     // TODO the clone seems to be broken
     //    std::shared_ptr<iir::StencilInstantiation> stencilInstantiation = origSI->clone();
     std::shared_ptr<iir::StencilInstantiation> stencilInstantiation = origSI;
 
-    PassInlining inliner(true, PassInlining::InlineStrategyKind::IK_ComputationsOnTheFly);
-
-    inliner.run(stencilInstantiation);
```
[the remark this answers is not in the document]
twicki: The inliner is no longer run during code-gen time. This was a hacky way of saying it is the last pass that you should run before code-generating. This is now replaced by the line in the passmanager that runs it lasst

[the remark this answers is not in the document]
havogt: Sorry I meant: what happens if we forgot to apply the inliner during optimization?

src/dawn/CodeGen/GridTools/GTCodeGen.cpp
twicki: we can do that

src/dawn/Optimizer/PassInlining.h
twicki: sure

src/dawn/CodeGen/Cuda/MSCodeGen.hpp
twicki: until when do we determine objects small enough to copy?

src/dawn/CodeGen/Cuda/CudaCodeGen.cpp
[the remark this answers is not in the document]
twicki: https://github.com/MeteoSwiss-APN/dawn/blob/2cda16bb86824e302865bf02f5e71b030a00c3d3/src/dawn/CodeGen/Cuda/ASTStencilBody.cpp#L77

src/dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.cpp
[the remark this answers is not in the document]
twicki: done

src/dawn/CodeGen/CodeGen.h
twicki: since it's only used once afterwards I don't see much benefit. Would it help readability?

havogt: It's used in each backend.

src/dawn/Optimizer/PassInlining.h
twicki: spawned this issue:
https://github.com/MeteoSwiss-APN/dawn/issues/241

src/dawn/CodeGen/Cuda/MSCodeGen.hpp
lukasm91: Actually I need to correct myself, I only read the name of the variable and I thought this is the main options object. In that case it makes sense to do the copy because we want to look at the codegen part individually (and there it's done anyway).
In this case here, the usage is pretty simple (`MSCodeGen{options}.generate(...);`), so the object goes out of scope immediately. So that would be (and I know that I am contradicting what I wrote in the morning) an ideal candidate for `const&`. I don't think that it costs a lot to do the copy, because the string will be "small-string optimized" anyway (string=3*64bit), so it will only generate some boiler plate code for the unused copy.
So after carefully thinking about it again, I don't care that much; it is not critical code at all (regarding performance). Regarding pointer safety, it doesn't look critical either.

twicki: Hannes and I had a quick discussion on this yesterday. His stance was that we actually want a copy of the object since the reference (const or not) means I / someone can change the options during code-gen. This we can be sure will not happen if we copy the object. So I think if we have a object where we do not want changes by someone else that is small enough (tbd) to copy we can go for that.

Thoughts on that?

Issue #210, opened 2019-06-19
Possible inconsistency in AccessIDToNameMap for dycore's advection_pptp.cpp


Stagno:
When code generating (for gridtools) source https://github.com/MeteoSwiss-APN/clang-gridtools/blob/master/src/dycore/advection_pptp.cpp with `-freport-accesses` flag, this assertion fails:
```
ACCESSES: line 291: Assertion failed: `AccessIDToNameMap_.count(AccessID)' 
Function: 'std::__cxx11::string dawn::iir::StencilFunctionInstantiation::getFieldNameFromAccessID(int) const'
Location: /home/giacomo/dawn/src/dawn/IIR/StencilFunctionInstantiation.cpp:300
/usr/lib/llvm-6.0/lib/libLLVM-6.0.so.1(_ZN4llvm3sys15PrintStackTraceERNS_11raw_ostreamE+0x2a)[0x7f72805580ea]
/usr/lib/llvm-6.0/lib/libLLVM-6.0.so.1(_ZN4llvm3sys17RunSignalHandlersEv+0x56)[0x7f7280556366]
/usr/lib/llvm-6.0/lib/libLLVM-6.0.so.1(+0x81c49b)[0x7f728055649b]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7f727f30b890]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xc7)[0x7f727e9a5e97]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x141)[0x7f727e9a7801]
/home/giacomo/gtclang/bundle/install/bin/gtclang(+0x11900ae)[0x56496f6a90ae]
/home/giacomo/gtclang/bundle/install/bin/gtclang(+0x11900c4)[0x56496f6a90c4]
/home/giacomo/gtclang/bundle/install/bin/gtclang(_ZNK4dawn3iir28StencilFunctionInstantiation24getFieldNameFromAccessIDB5cxx11Ei+0x16f)[0x56496f82d6af]
/home/giacomo/gtclang/bundle/install/bin/gtclang(_ZNK4dawn3iir28StencilFunctionInstantiation19getNameFromAccessIDB5cxx11Ei+0xa5)[0x56496f82d795]
/home/giacomo/gtclang/bundle/install/bin/gtclang(+0x12ca8e8)[0x56496f7e38e8]
[...]

PR #242, opened 2019-08-19
Issue 199

### LINKED ISSUE #199: Assertion on existing keys
Ideally the double sided map should have two insert methods, once that asserts if the key is present and one that overrides. This way we can ensure we do the correct thing. The same goes for insertion into unique maps (see localvar7 PR)

mroethlin:
## Technical Description
 
Current proposal for issue 199. DoubleSidedMap now features two insertion methods, one which asserts that elements are not already present, and another one which will silently reject the insertion if the element is already present. Appropriate methods are called from StencilMetaInformation. Furthermore, all emplace calls to unordered_maps in StencilMetaInformation were prepended with an assert that the element is not already present"

 #### Resolves
 
#199 
 
 ## Dependencies
 
 This PR is independent.


src/dawn/CodeGen/CXXNaive/ASTStencilBody.cpp
```diff
@@ -94,7 +94,9 @@ void ASTStencilBody::visit(const std::shared_ptr<UnaryOperator>& expr) { Base::v
 
 void ASTStencilBody::visit(const std::shared_ptr<BinaryOperator>& expr) { Base::visit(expr); }
 
-void ASTStencilBody::visit(const std::shared_ptr<AssignmentExpr>& expr) { Base::visit(expr); }
+void ASTStencilBody::visit(const std::shared_ptr<AssignmentExpr>& expr) {
```
twicki: seems like your formatting is either incompatible or you're using something different?

src/dawn/IIR/StencilMetaInformation.cpp
```diff
@@ -579,8 +582,8 @@ int StencilMetaInformation::getStencilIDFromStencilCallStmt(
 }
 
 void StencilMetaInformation::insertStencilCallStmt(std::shared_ptr<StencilCallDeclStmt> stmt,
-                                                   int stencilID) {
-  StencilIDToStencilCallMap_.emplace(stencilID, stmt);
+                                                   int stencilID) {
```
twicki: format?

src/dawn/Support/DoubleSidedMap.h
```diff
@@ -26,10 +26,25 @@ class DoubleSidedMap {
...
+  /// present. That is, one to one relation (bijective mapping) is maintained
+  void add(const Key1& key1, const Key2& key2) {
+    DAWN_ASSERT(!directHas(key1));
+    DAWN_ASSERT(!reverseHas(key2));
+
+    directMap_.emplace(key1, key2);
+    reverseMap_.emplace(key2, key1);
+  }
+
+  /// Insert a new element into the map. This function works analogous to the emplace method in
```
havogt: emplace has to be deleted as a non-bijective map breaks the erase functions.

src/dawn/Support/DoubleSidedMap.h
```diff
@@ -26,10 +26,25 @@ class DoubleSidedMap {
   std::unordered_map<Key2, Key1> reverseMap_;
 
 public:
+
+  /// Add a new element to the map. This function asserts that neither key1 nor key2 is already
+  /// present. That is, one to one relation (bijective mapping) is maintained
+  void add(const Key1& key1, const Key2& key2) {
+    DAWN_ASSERT(!directHas(key1));
```
havogt: This should not be an assert. It should be checked also in release...

Stagno:
will this PR fix this issue? https://github.com/MeteoSwiss-APN/dawn/issues/210

mroethlin:
launch jenkins

PR #245, opened 2019-08-21
Fix Serializer -> Optimizer dependency

twicki:
## Technical Description

We remove all the backwards dependencies from the Seriaizer to the Optimizer. This is part of cleaning up all unwanted (circular) dependencies between modules

####  Enhances

Now Serializer is independent of Optimizer

## Dependencies

This PR is independent


twicki:
launch jenkins

twicki:
We remove all the backwards dependencies from the Seriaizer to the Optimizer. This is part of cleaning up all unwanted (circular) dependencies between modules

PR #247, opened 2019-08-23
restore performance

twicki:
## Technical Description

Restores performance to before this bug was introduced

#### Resolves

Performance issues encountered starting 8/20

#### Example

\<jenkins\>/view/PASCHA/job/cosmo-prerelease-gtclang_perftest_daily/

## Dependencies

This PR is indepenedent


twicki:
we left this as an open issue: https://github.com/MeteoSwiss-APN/dawn/issues/248

PR #249, opened 2019-08-23
Disable python if protobuf doesn't have python

havogt:


havogt:
@lukasm91 Please check if your issue is solved.

havogt:
launch jenkins

PR #257, opened 2019-09-03
fixed sir::value class

### LINKED ISSUE #256: Strange Behavior of SIR::Value
While working on [PR236](https://github.com/MeteoSwiss-APN/dawn/pull/236) I noticed some strange behavior of the SIR::Value class, outlined in the following code sample:

```
  dawn::sir::Value test, test2;
  test.setType(dawn::sir::Value::TypeKind::Integer);
  test.setValue(12.); //silently switches type
  printf("value has type %s\n", 
    test.getType() == dawn::sir::Value::TypeKind::Integer ? "Integer" : "Double");  //prints "Double"

  //this would mean an assertion failure,  even though type was set to integer
  //int ret_val = test.getValue<int>();  

  // so lets try to correct the type to what we want (int)
  //    this sets test into an inconsitent state
  test.setType(dawn::sir::Value::TypeKind::Integer);
  printf("value is %s\n", test.toString().c_str()); // prints 0

  //this is not just an issue of toString()
  int ret_val = test.getValue<int>();
  printf("value stored is in fact: %d\n", ret_val); // still prints 0
```

This is the root cause of some (de)serialzation bugs addressed in the PR.

mroethlin:
## Technical Description

This fixes the issues about the problematic behavior of SIR::Value.

#### Resolves

resolves #256 

#### Dependencies

This PR is independent.


mroethlin:
launch jenkins

src/dawn/SIR/SIR.h
```diff
@@ -348,9 +361,47 @@ struct Value : NonCopyable {
...
+  }
+
+  template <class T>
+  typename std::enable_if<std::is_arithmetic<T>::value>::type
+  setValue(const T& value) {
+    if (type_ == TypeKind::None) {
+      valueImpl_ = make_unique<ValueImpl<T>>(value);
+      type_ = TypeInfo<T>::Type;
+    } else {
+      if (isArithmetic()) {
```
lukasm91: Do we need this check? You replicate isArithmetic in the switch case, the default is unreachable, but then there is an else, which would be equivalent.

src/dawn/SIR/SIR.h
```diff
@@ -348,9 +361,47 @@ struct Value : NonCopyable {
...
+    }
+  }
+
+  template <class T>
+  typename std::enable_if<std::is_arithmetic<T>::value>::type
+  setValue(const T& value) {
+    if (type_ == TypeKind::None) {
+      valueImpl_ = make_unique<ValueImpl<T>>(value);
+      type_ = TypeInfo<T>::Type;
+    } else {
```
lukasm91: `elseif`

src/dawn/SIR/SIR.h
```diff
@@ -348,9 +361,47 @@ struct Value : NonCopyable {
...
+            valueImpl_ = make_unique<ValueImpl<int>>((int) value);
+            break;
+          case Value::Boolean:
+            valueImpl_ = make_unique<ValueImpl<bool>>((bool) value);
+            break;
+          case Value::Double:
+            valueImpl_ = make_unique<ValueImpl<double>>((double) value);
+            break;
+          default:
+            DAWN_ASSERT(false); //unreachable
```
lukasm91: `dawn_unreachable`?

src/dawn/SIR/SIR.h
```diff
@@ -348,9 +361,47 @@ struct Value : NonCopyable {
...
+        valueImpl_ = make_unique<ValueImpl<T>>(value);
+      } else {
+        //type mismatch
+        DAWN_ASSERT_MSG(false, "type mismatch trying to set sir::Value");
+      }
+    }
+  }
+
+  template <class T>
+  typename std::enable_if<std::is_arithmetic<T>::value>::type
```
lukasm91: Why two overloads? Might we be able to set the value if the type is constructible from T?

src/dawn/SIR/SIR.h
```diff
@@ -348,9 +361,47 @@ struct Value : NonCopyable {
...
   template <class T>
-  void setValue(const T& value) {
-    valueImpl_ = make_unique<ValueImpl<T>>(value);
-    type_ = TypeInfo<T>::Type;
+  typename std::enable_if<!std::is_arithmetic<T>::value>::type
+  setValue(const T& value) {
+    if (type_ == TypeKind::None) {
+      valueImpl_ = make_unique<ValueImpl<T>>(value);
+      type_ = TypeInfo<T>::Type;
+    } else {
```
lukasm91: `else if`

src/dawn/SIR/SIR.h
```diff
@@ -348,9 +361,47 @@ struct Value : NonCopyable {
...
+      if (TypeInfo<T>::Type == type_) {
```
lukasm91: So for non-arithmetic types, the behaviour is different and T actually needs to be the same as `type`. For arithmetic it only needs to be convertible. This is inconsistent.

src/dawn/SIR/SIR.h
```diff
@@ -330,12 +330,25 @@ struct Value : NonCopyable {
...
+  bool isArithmetic() const;
+
   /// @brief Check if value is empty
   bool empty() const { return valueImpl_ == nullptr; }
 
   /// @brief Get/Set the underlying type
   TypeKind getType() const { return type_; }
-  void setType(TypeKind type) { type_ = type; }
+  void setType(TypeKind type) {
+    if (type_ == type) {
```
lukasm91: This first if is redundant and can be moved into else.
```
if (type_ == TypeKind::None)
    type_ = type;
else
    DAWN_ASSERT_MSG(type_ == type, "...");
```

lukasm91:
Check in the description how you can link to an issue and tell github that it automatically closes it if the PR is merged.

mroethlin:
As discussed, sir::Value is now an immutable class

mroethlin:
launch jenkins

src/dawn/SIR/SIR.h
```diff
@@ -371,6 +368,12 @@ struct Value : NonCopyable {
   }
 
 private:
+  template <class T>
+  void setValue(const T& value) {
+    valueImpl_ = make_unique<ValueImpl<T>>(value);
+    type_ = TypeInfo<T>::Type;
```
lukasm91: you can also completely remove `type_`: Add a member to `ValueImpl` and its base (getType) which returns `TypeInfo<T>::Type` in the implementation.

src/dawn/SIR/SIR.h
```diff
@@ -314,9 +314,14 @@ struct Value : NonCopyable {
 
   template <class T>
   explicit Value(T&& value) : isConstexpr_(false) {
+    //TODO
+    //  I get a compiler error I don't understand when not using this redirection
```
lukasm91: I assume, in the end you will only have 
```
... : valueImpl(std::forward<T>(value)) {}
```
And the samme inside ValueImpl (and you can remove the pointer in ValueImpl). Maybe, this solves your compilre error too.

src/dawn/SIR/SIR.h
```diff
@@ -314,9 +314,14 @@ struct Value : NonCopyable {
```
lukasm91: Is the default constructor needed (and related `TypeKind::None`)? I am wondering how this would be used, because it complicates the implementation of Value and any user of Value (because they all need to scope with the possibility that Value might be nothing), e.g. CodeGen.cpp, Line 67.

src/dawn/SIR/SIR.h
```diff
@@ -346,13 +350,6 @@ struct Value : NonCopyable {
...
   }
 
-  /// @brief Set the value
-  template <class T>
-  void setValue(const T& value) {
-    valueImpl_ = make_unique<ValueImpl<T>>(value);
-    type_ = TypeInfo<T>::Type;
-  }
-
   struct EmptyType {};
```
lukasm91: Unused

src/dawn/SIR/SIR.h
mroethlin: Unfortunately valueImp_ has to be a pointer, see below

src/dawn/SIR/SIR.h
mroethlin: The default constructor is not needed and neither is `TypeKind::None`. What is needed is the `empyt()` method for things like `bool foo`, i.e. variables with a type but no value. However, in this case I believe that `empty()` is a bad name for the method at hand and I opted to rename it to `has_value()`

src/dawn/SIR/SIR.h
lukasm91: Not the pointer to valueImpl (that's clear), but the pointer in valueImpl (ValuePtr)

src/dawn/SIR/SIR.h
lukasm91: What's the reason for an empty Value? Respectively, how can a type have an empty set of possible values?

mroethlin:
launch jenkins

src/dawn/SIR/SIR.h
```diff
@@ -282,11 +282,17 @@ struct Value : NonCopyable {
   template <class T>
   explicit Value(T value) : isConstexpr_(false) {
     valueImpl_ = make_unique<ValueImpl<T>>(value);
+    //NOTE: perfect forwarding in this case can be achieved by using 
+    //explicit Value(T value) : isConstexpr_(false) {make_unique<ValueImpl<std::decay_t<T>>>(std::forward<T>(value));}
```
havogt: `typename std::decay<T>::type` or is there something else I am missing which doesn't work in c++11?

PR #259, opened 2019-09-04
AST refactoring (templating)

Stagno:
## Technical Description

Makes AST classes templated w.r.t. data containers. SIR component will have such classes with template parameter `SIRASTData`, while users of IIR will have them templated with `IIRASTData`.
`VerticalRegionDeclStmt` is moved to the SIR component, as it is not used after SIR goes into IIR. 
Also, the structure of SIR is changed to make component AST independent from it: StencilCall is moved to AST and references fields by string instead of through `sir::Field`, also `BoundaryConditionDeclStmt` references fields by string instead of through `sir::Field` and, finally, `VerticalRegion` (which is moved to `VerticalRegionDeclStmt`'s data container) doesn't reference anymore the underlying AST, which is moved to `VerticalRegionDeclStmt`.
Notice that, in component SIR, **derived** AST visitors (from base `ast::ASTVisitor<SIRASTData>`) are needed, due to the presence of a statement (`VerticalRegionDeclStmt`) which hasn't been defined in component AST (see `dawn/AST/ASTVisitor.h` and `dawn/SIR/ASTVisitor.h`).
`OptimizerContext.cpp` has been substantially changed to convert from sir's AST to iir's AST. `Statement` wrapper around `Stmt` has been removed.
Since `sir::StencilFunction` holds an sir's AST, it is converted (in `OptimizerContext.cpp`) to `iir::StencilFunction`.

## Dependencies

Related PR on gtclang: https://github.com/MeteoSwiss-APN/gtclang/pull/157

## TODOs

- newly introduced `iir::StencilFunction` is not (de)serialized, should wait for deserialize IIR PR to be merged.
- move more data to `IIRASTData` container (another upcoming PR)
- `ast::ASTVisitor` cannot be extended to Stmts introduced outside of component AST (such as `sir::VerticalRegionDeclStmt`), except by subclassing it. But this requires dynamic casts in the new Stmt's `accept(...)` method as `ast::ASTVisitor` doesn't have a method `visit(sir::VerticalRegionDeclStmt)`. This solution, however, is not compile-time safe.


Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

cosunae:
Some thoughts here: 
I went through a large fraction of the code, but I could not find any meaningful to comment on, since most of the code was adding DataTraits template to all the classes, methods and forwarding that template parameter. 
The code here looks solid to me, but somehow that made me seriously doubt this is the right approach/design for this. 
We have templatized a large fraction of the code, that is low level, and therefore it will be processed all for all translation units. And most of the code will not do anything with those templates, just forward to Expr or some other class. 
Additionally the code became more complex, and less readable. 
And in the end the differences between SIRData and IIRData are not big. 
So my question is: why choosing here a policy like pattern? 
At this point I think a type erasure looks more convenient for this problem to me... 
but perhaps we can discuss it, I am not sure I got a good understanding of this PR

twicki:
refactoring takes another route

PR #266, opened 2019-09-12
Fix python examples

twicki:
## Technical Description

The verbose mode of the python examples is broken due to renaming

#### Resolves / Enhances

makes the verbose mode of the example works (baused on Joerg's feedback)

#### Example

```
cd dawn/bundle/install/examples/python
export PYTHONPATH=dawn/bundle/install/python
python3.7 copy_stencil.py -v
```

## Dependencies

This PR is independent


twicki:
There were just some formatting violations

PR #268, opened 2019-09-16
Fix tests in clang-gridtools due to issue_199

mroethlin:
## Technical Description

Some edge cases in the clang-gridtools test suite were not correctly handled by the changes introduced due to the fix to issue_199 (`DoubleSidedMap` and general `StencilMetaData` clean up). `AccessID` s to literals are now always strictly negative, no matter the context (e.g. even if the literal is in fact a global which was overwritten by a config file). A test was adapted accordingly.

## Dependencies

This PR is independent.


mroethlin:
launch jenkins

twicki:
looks good. Testing on lower levels should launch the pipeline though

twicki:
launch pipeline

PR #271, opened 2019-09-17
Move -output option to dawn

Stagno:
### Technical Description
Moves `-output` option to dawn in order to use it while serializing IIR instead of the file name stored in the IIR's metadata. Also removes some debug printfs.

### Dependencies
https://github.com/MeteoSwiss-APN/gtclang/pull/160

### DIFF
--- a/src/dawn/Compiler/Options.h
+++ b/src/dawn/Compiler/Options.h
@@ -25,7 +25,6 @@ struct Options {
 #define OPT(TYPE, NAME, DEFAULT_VALUE, OPTION, OPTION_SHORT, HELP, VALUE_NAME, HAS_VALUE, F_GROUP) \
   TYPE NAME = DEFAULT_VALUE;
 #include "dawn/Compiler/Options.inc"
-#include "dawn/Optimizer/OptimizerOptions.inc"
 #undef OPT
 };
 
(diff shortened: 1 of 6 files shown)


Stagno:
launch jenkins

src/dawn/Compiler/Options.inc
```diff
@@ -32,6 +32,7 @@
 //                  command-line target will be generated which allows to negate the option i.e 
 //
 
+OPT(std::string, OutputFile, "", "output", "o", "Write output to <file>", "<file>", true, false)
```
twicki: I would move it to after the back end decision since this is the more "natural" flow to read the options. Decide on the back end and then choose the output file. Or what was your reasoning to have it here?

Stagno: moved in last commit

Stagno:
launch jenkins

src/dawn/Compiler/Options.h
```diff
@@ -25,6 +25,7 @@ struct Options {
 #define OPT(TYPE, NAME, DEFAULT_VALUE, OPTION, OPTION_SHORT, HELP, VALUE_NAME, HAS_VALUE, F_GROUP) \
   TYPE NAME = DEFAULT_VALUE;
 #include "dawn/Compiler/Options.inc"
+#include "dawn/Optimizer/OptimizerOptions.inc"
```
twicki: Isn't that more error prone, now you have to do it in multiple places. Why not include it in the Options.inc but in the end if you want it afterwards?

Stagno: So, I wasn't sure about this. It seems reasonable to separate the 2 options subsets, as we may need only one, as in https://github.com/MeteoSwiss-APN/dawn/blob/c4a639d808cb6c1253ee8b1140fbfa96a2f867e5/src/dawn/Compiler/DawnCompiler.cpp#L123 but it is also true that optimizer's options are also compiler's options. Since it's no harm to include more options than necessary, I will include again optimizer's opts into compiler's opts.

Stagno:
launch jenkins

PR #272, opened 2019-09-18
Add default value to globals

Stagno:
### Description
Since type is not explicitly serialized for globals, this makes it always known.

### Example
```
globals {
  bool ltur;
  double zdtr = 0.8;
  ...
}
```
becomes, in SIR:
```
"global_variables": {
   "ltur": {
    "is_constexpr": false,
    "boolean_value": false
   },
   "zdtr": {
    "is_constexpr": false,
    "double_value": 0.8
   }, ...
```


Stagno:
launch pipeline

PR #274, opened 2019-09-19
Add interval-partitioner as a pass

twicki:
## Technical Description

Adds the interval partitioner used in the cuda code-generation as a pass that can be enabled. 
Also adds the pretty-printer of the stencil-instantiation

#### Resolves / Enhances

This is used in cuda code-gen as well as helps to write DACE-sdfg's

## Dependencies

This PR is independent


twicki:
launch pipeline

src/dawn/Optimizer/PassIntervalPartitioner.h
```diff
@@ -0,0 +1,54 @@
...
+///
+/// This pass is not necessary to create legal code and is hence not in the debug-group
+class PassIntervalPartitioner : public Pass {
+public:
+  PassIntervalPartitioner(OptimizerContext& context) : Pass(context, "PassIntervalPartitioner") {}
+
+  /// @brief Pass implementation
+  bool run(const std::shared_ptr<iir::StencilInstantiation>& stencilInstantiation) override;
+
+private:
```
lukasm91:
```suggestion
```

src/dawn/Optimizer/PassTemporaryToStencilFunction.cpp
```diff
@@ -1,4 +1,4 @@
-﻿//===--------------------------------------------------------------------------------*- C++ -*-===//
+//===--------------------------------------------------------------------------------*- C++ -*-===//
```
lukasm91: what happened here?

src/dawn/Optimizer/PassIntervalPartitioner.cpp
```diff
@@ -0,0 +1,71 @@
+﻿//===--------------------------------------------------------------------------------*- C++ -*-===//
```
lukasm91: You will need to explain me the idea of this pass.

src/dawn/Optimizer/PassTemporaryToStencilFunction.cpp
twicki: I can't say

src/dawn/Optimizer/PassIntervalPartitioner.cpp
twicki: Assume we have a multistage with intervals [start : end], [start+1 : end-1] [start : start+4] [end-2 : end]

The pass will split those into disjoint intervals:

[start : start], [start+1 : start+4], [start+5 : end-3], [end-2 : end-1], [end : end] and will add multiple do-methods with these intervals for each computation

src/dawn/Optimizer/PassIntervalPartitioner.cpp
```diff
@@ -0,0 +1,71 @@
...
+      auto cloneMS = multiStage->clone();
+
+      auto multiInterval = multiStage->computePartitionOfIntervals();
+
+      // empty out the multistage
+      for(auto cloneStageIt = cloneMS->childrenBegin(); cloneStageIt != cloneMS->childrenEnd();
+          ++cloneStageIt) {
+        auto cloneDoMethodIt = (*cloneStageIt)->childrenBegin();
+        while(cloneDoMethodIt != (*cloneStageIt)->childrenEnd()) {
+          cloneDoMethodIt = (*cloneStageIt)->childrenErase(cloneDoMethodIt);
```
lukasm91: instead of cloning the MS and deleting its data, can't we just create a new multistage?

src/dawn/Optimizer/StatementMapper.cpp
```diff
@@ -120,37 +120,42 @@ void StatementMapper::visit(const std::shared_ptr<iir::IfStmt>& stmt) {
 void StatementMapper::visit(const std::shared_ptr<iir::VarDeclStmt>& stmt) {
   DAWN_ASSERT(initializedWithBlockStmt_);
 
-  // This is the first time we encounter this variable. We have to make sure the name is not
-  // already used in another scope!
-  int AccessID = instantiation_->nextUID();
+  int accessID = -1;
+  if(metadata_.hasStmtToAccessID(stmt)) {
```
lukasm91: is this change needed because after duplicating a do-method, we might have two local variables with the same name?

PR #281, opened 2019-09-24
Test deserialization

mroethlin:
## Technical Description 

This introduces testing for the `restoreIIR` function in `OptimizerContext` which computes the derived info of a stencil, given the stencils IIR. This is achieved by constructing a stencil in memory (A) and deserializing an equivalent stencil from disk (B). The in memory IIR, meta data and derived Info of A and B are then compared. 

Since those tests need to access the file system, they are considered integration tests instead of unit tests. These are the first integration tests for dawn. Build files have been adapted accordingly. 

Currently, to ensure equivalent ids between stencils from memory as well as from disk the UUID singleton needs to be reset. This is of course bad practice. The correct way to compare this would be to first try to find a bijective mapping between both id sets. To be discussed if this should go into a different PR are handled herein.


mroethlin:
launch pipeline

mroethlin:
launch pipeline

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,607 @@
...
+      doMethod->update(iir::NodeUpdateType::level);
+    }
+    stage.update(iir::NodeUpdateType::level);
+  }
+  for(const auto& MSPtr : iterateIIROver<iir::Stage>(*(target->getIIR()))) {
+    MSPtr->update(iir::NodeUpdateType::levelAndTreeAbove);
+  }
+}
+
+static void createLapStencilIIRInMemory(std::shared_ptr<iir::StencilInstantiation>& target) {
```
lukasm91: This part should hopefullly be much simpler now (if everything works as expected) with the IIRBuilder

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,607 @@
...
+  ((compareDerivedInformation(lhs->getIIR().get(), rhs->getIIR().get())));
+
+  return true;
+}
+
+TEST(IIRDeserializerTest, CopyStencil) {
+  char cCurrentPath[FILENAME_MAX];
+  getcwd(cCurrentPath, sizeof(cCurrentPath));
+  cCurrentPath[sizeof(cCurrentPath) - 1] = '\0'; /* not really required */
+  printf ("The current working directory is %s", cCurrentPath);
```
lukasm91: pritnf is a bit too much c-like ;-) Anyway I'm wondering why we should print this?

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+TEST(IIRDeserializerTest, LaplStencil) {
+  Options compileOptions;
+  OptimizerContext::OptimizerContextOptions optimizerOptions;
+  DawnCompiler compiler(&compileOptions);   
+  OptimizerContext optimizer(compiler.getDiagnostics(), optimizerOptions, std::make_shared<dawn::SIR>());
+
+  //read IIRs from file
+  std::shared_ptr<iir::StencilInstantiation> lap_stencil_from_file = std::make_shared<iir::StencilInstantiation>(
+    *optimizer.getSIR()->GlobalVariableMap, optimizer.getSIR()->StencilFunctions);
```
lukasm91: As far as I remember you can just pass empty maps here. No need to use OptimzierContext.

test/integration-test/reference_iir/copy_stencil.iir
```diff
@@ -0,0 +1,238 @@
+{
```
lukasm91: Can you put a README somewhere here how to update the reference (or a very simple bash script)

test/unit-test/CMakeLists.txt
```diff
@@ -39,5 +39,30 @@ macro(dawn_add_unittest_impl)
...
+# dawn_add_integrationtest_impl
+# ----------------------
+#
+# Compile the given objects into a runnable integration test executable (.exe) and register it within CTest.
+# The executable will be stored in ${CMAKE_BINARY_DIR}/bin/integrationtest.
+#
+#    NAME:STRING=<>     - Name of the unittest exectuable as well as the CMake target to build it.
+#    SOURCES:STRING=<>   - List of source files making up the exectuable.
+#
+macro(dawn_add_integrationtest_impl)
```
lukasm91: Shouldn't we put this into `integrationtest/CMakeLists.txt`?

test/unit-test/dawn/IIR/TestIIRSerializer.cpp
```diff
@@ -75,15 +75,23 @@ bool compareIIRs(iir::IIR* lhs, iir::IIR* rhs) {
 
           // checking each of the StmtAccesspairs
           for(int stmtidx = 0, stmtSize = lhsDoMethod->getChildren().size(); stmtidx < stmtSize;
-              ++stageidx) {
+              ++stmtidx) {
```
lukasm91: nice one!

test/integration-test/TestIIRDeserializer.cpp
mroethlin: Right, however, if I remember right we agreed at the sprint meeting yesterday that this issue was to be closed asap with the option to improve upon it later. Maybe we should put additional tasks w.r.t that into the backlog?

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,607 @@
...
+      // checking each of the stages
+      for(int stageidx = 0, stageSize = lhsMSS->getChildren().size(); stageidx < stageSize;
+          ++stageidx) {
+        const auto& lhsStage = lhsMSS->getChild(stageidx);
+        const auto& rhsStage = rhsMSS->getChild(stageidx);
+
+        // struct DerivedInfo {
+
+        //   /// Declaration of the fields of this stage
+        //   [X] std::unordered_map<int, Field> fields_;
```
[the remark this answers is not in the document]
mroethlin: Those were notes for me to check if I actually compare all the derived info. To be removed before merging

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+using namespace dawn;
+
+namespace {
+#define IIR_EARLY_EXIT(value)                                                                      \
+  if(!value)                                                                                       \
+    return false;
+
+static bool compareIIRstructures(iir::IIR* lhs, iir::IIR* rhs) {
+  IIR_EARLY_EXIT(lhs->checkTreeConsistency());
```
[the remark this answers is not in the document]
mroethlin: This is a pattern that is used throughout all of the IIR testing, I guess to be able to use instant exit assertions in functions returning something other than void. Changed it to gtest assertions for this PR, but maybe we should think about a clean up of the whole IIR test base, including unit tests

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,607 @@
...
+              ++stmtidx) {
+            const auto& lhsStmt = lhsDoMethod->getChild(stmtidx);
+            const auto& rhsStmt = rhsDoMethod->getChild(stmtidx);
+            // check the statement
+            IIR_EARLY_EXIT(
+                (lhsStmt->getStatement()->ASTStmt->equals(rhsStmt->getStatement()->ASTStmt.get())));
+
+            // check the accesses
+            IIR_EARLY_EXIT((lhsStmt->getCallerAccesses()->getReadAccesses().size()
+              == lhsStmt->getCallerAccesses()->getReadAccesses().size()));
```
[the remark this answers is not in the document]
mroethlin: good catch!

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+    IIR_EARLY_EXIT((lhsStencil->getStageDependencyGraph() == rhsStencil->getStageDependencyGraph())); //this _should_ work, to be tested
+    IIR_EARLY_EXIT((lhsStencil->getFields() == rhsStencil->getFields()));
+
+    // checking each of the multistages
+    for(int mssidx = 0, mssSize = lhsStencil->getChildren().size(); mssidx < mssSize; ++mssidx) {
+      const auto& lhsMSS = lhsStencil->getChild(mssidx);
+      const auto& rhsMSS = rhsStencil->getChild(mssidx);
+
+      IIR_EARLY_EXIT((lhsMSS->getCaches().size() == rhsMSS->getCaches().size()));
```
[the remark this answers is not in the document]
mroethlin: For some maps this makes sense, since they are storing shared pointers as elements. Hope I catched all of them for which == can be used

src/dawn/Serialization/IIRSerializer.cpp
```diff
@@ -584,11 +584,15 @@ void IIRSerializer::deserializeMetaData(std::shared_ptr<iir::StencilInstantiatio
         makeLocation(call.stencil_call_decl_stmt().stencil_call()));
     for(const auto& protoFieldName : call.stencil_call_decl_stmt().stencil_call().arguments()) {
       astStencilCall->Args.push_back(protoFieldName);
-    }
+    }    
+    
+    //this introduces problems when trying to retrieve the stencil_call_decl_stmt later on!
+    // auto stmt = std::make_shared<iir::StencilCallDeclStmt>(
+    //       astStencilCall, makeLocation(call.stencil_call_decl_stmt()));
```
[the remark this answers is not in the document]
mroethlin: ok, I really wonder why somebody changes this from retrieving to making a copy though

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+            // check the statement
+            ASSERT_TRUE(lhsStmt->getStatement()->ASTStmt->equals(rhsStmt->getStatement()->ASTStmt.get()));
+
+            // check the accesses
+            ASSERT_TRUE(lhsStmt->getCallerAccesses()->getReadAccesses()
+              == rhsStmt->getCallerAccesses()->getReadAccesses());
+            ASSERT_TRUE(lhsStmt->getCallerAccesses()->getWriteAccesses()
+              == rhsStmt->getCallerAccesses()->getWriteAccesses());
+
+            // if(lhsStmt->getCallerAccesses()) {
```
lukasm91: delete?

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+        const auto& rhsStage = rhsMSS->getChild(stageidx);
+        ASSERT_TRUE(lhsStage->getStageID() == rhsStage->getStageID());
+
+        // checking each of the doMethods
+        ASSERT_TRUE(lhsStage->getChildren().size() == rhsStage->getChildren().size());
+        for(int doMethodidx = 0, doMethodSize = lhsStage->getChildren().size();
+            doMethodidx < doMethodSize; ++doMethodidx) {
+          const auto& lhsDoMethod = lhsStage->getChild(doMethodidx);
+          const auto& rhsDoMethod = rhsStage->getChild(doMethodidx);
+          ASSERT_TRUE(lhsDoMethod->getID() == rhsDoMethod->getID());
```
lukasm91: Note that as a rule of thumb, you should use `EXPECT_*` if the test can continue, and `ASSERT_*` if the test cannot continue. So here you should use `EXPECT_EQ`, while at a point where you compare the number of children and then loop over the number of children, you should use `ASSERT_EQ`.

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+               rhsControlFlowStmts[i]->StackTrace->at(j))) {
+            FAIL();
+          }
+        }
+      }
+      FAIL();
+    }
+  }
+
+  SUCCEED();
```
lukasm91: I wouldn't put this, because it has only documentary purpose and getting to this point is obviously a succeed anyway.

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+  ASSERT_TRUE(lhsControlFlowStmts.size() == rhsControlFlowStmts.size());
+  for(int i = 0, size = lhsControlFlowStmts.size(); i < size; ++i) {
+    if(!lhsControlFlowStmts[i]->ASTStmt->equals(rhsControlFlowStmts[i]->ASTStmt.get()))
+      FAIL();
+    if(lhsControlFlowStmts[i]->StackTrace) {
+      if(rhsControlFlowStmts[i]->StackTrace) {
+        for(int j = 0, jsize = lhsControlFlowStmts[i]->StackTrace->size(); j < jsize; ++j) {
+          if(!(lhsControlFlowStmts[i]->StackTrace->at(j) ==
+               rhsControlFlowStmts[i]->StackTrace->at(j))) {
+            FAIL();
```
lukasm91: I don't think that you need any of this `FAIL` because all of them are inside a `if` which is the `ASSERT_*` or `EXPECT_*`. Like here, it is
```
ASSERT_EQ(lhsControlFlowStmts[i]->StackTrace->at(j), rhs...);
```
 Eventhough `lhsControlFlowStmts[i]->StackTrace->at(j)` is a `ast::StencilCall*`  which is not comparable I think.

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+
+  //IIR
+  // struct DerivedInfo {
+  //   /// StageID to name Map. Filled by the `PassSetStageName`.
+  //   [X] std::unordered_map<int, std::string> StageIDToNameMap_;
+  //   /// Set containing the AccessIDs of fields which are manually allocated by the stencil and serve
+  //   /// as temporaries spanning over multiple stencils
+  // };
+
+  // IIR_EARLY_EXIT(lhs->getStageIDToNameMap() == rhs->getStageIDToNameMap());  //this is not present in stencil from memory
```
lukasm91: lots of commented out stuff in this file. Delete? uncomment?

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+    ASSERT_TRUE(rhsValue->equals(lhsPair.second.get()));
+  }
+
+  SUCCEED();
+}
+
+void compareDerivedInformation(iir::IIR* lhs, iir::IIR* rhs) {
+  //IIR -> Stencil -> Multistage -> Stage -> Do Method -> (StatementAccessPair)
+
+  //IIR
```
lukasm91: If you keep this, please add a description, what this means.

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+}
+
+TEST(IIRDeserializerTest, CopyStencil) {
+  Options compileOptions;
+  OptimizerContext::OptimizerContextOptions optimizerOptions;
+  DawnCompiler compiler(&compileOptions);   
+  OptimizerContext optimizer(compiler.getDiagnostics(), optimizerOptions, std::make_shared<dawn::SIR>());
+
+  //read IIR from file
+  std::shared_ptr<iir::StencilInstantiation> copy_stencil_from_file = std::make_shared<iir::StencilInstantiation>(
```
lukasm91:
```suggestion
  auto copy_stencil_from_file = std::make_shared<iir::StencilInstantiation>(
```
Very natural place to put auto (also below).

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,598 @@
...
+#include <string>
+#include <unistd.h>
+#include <gtest/gtest.h>
+
+using namespace dawn;
+
+namespace {
+
+void compareIIRstructures(iir::IIR* lhs, iir::IIR* rhs) {
+  ASSERT_TRUE(lhs->checkTreeConsistency());
```
lukasm91: It would be cool if all of this could be implemented as a Visitor, but I don't think that works ... :-)

CMakeLists.txt
```diff
@@ -122,7 +122,7 @@ add_definitions(${DAWN_EXTERNAL_DEFINITIONS})
 yoda_report_result("Package summary" ${package_info})
 
 # Add clang-format target
-yoda_add_target_clang_format(DIRECTORIES ${CMAKE_SOURCE_DIR}/src EXTENSION ".h;.cpp")
+yoda_add_target_clang_format(DIRECTORIES ${CMAKE_SOURCE_DIR}/src ${CMAKE_SOURCE_DIR}/test EXTENSION ".h;.cpp")
```
lukasm91:
```suggestion
yoda_add_target_clang_format(DIRECTORIES ${CMAKE_PROJECT_DIR}/src ${CMAKE_PROJECT_DIR}/test EXTENSION ".h;.cpp")
```
I know, not your fault, but let's just fix it as we go

test/integration-test/GenerateInMemoryStencils.h
```diff
@@ -0,0 +1,25 @@
...
+//                     | (_| | (_| |\ V  V /| | | |
+//                      \__,_|\__,_| \_/\_/ |_| |_| - Compiler Toolchain
+//
+//
+//  This file is distributed under the MIT License (MIT).
+//  See LICENSE.txt for details.
+//
+//===------------------------------------------------------------------------------------------===//
+
+#pragma once
```
lukasm91: I prefer `#pragma once` but for consistency I would accept the `#ifdef` convention

test/integration-test/GenerateInMemoryStencils.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <memory>
```
lukasm91: Global includes should come after local includes (the rational for this convention is that you  are missing less includes in your own headers when you do that)

test/integration-test/GenerateInMemoryStencils.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "dawn/IIR/IIR.h"
+#include "dawn/IIR/StencilInstantiation.h"
+
+using namespace dawn;
```
lukasm91: remove

test/integration-test/GenerateInMemoryStencils.h
```diff
@@ -0,0 +1,25 @@
...
+
+void createCopyStencilIIRInMemory(std::shared_ptr<iir::StencilInstantiation>& target);
```
lukasm91: Is there a reason not to do 
```suggestion
std::shared_ptr<iir::StencilInstantiation> createCopyStencilIIRInMemory();
```

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,277 @@
...
+        }
+      }
+    }
+  }
+  const auto& lhsControlFlowStmts = lhs->getControlFlowDescriptor().getStatements();
+  const auto& rhsControlFlowStmts = rhs->getControlFlowDescriptor().getStatements();
+
+  ASSERT_EQ(lhsControlFlowStmts.size(), rhsControlFlowStmts.size());
+  for(int i = 0, size = lhsControlFlowStmts.size(); i < size; ++i) {
+    ASSERT_TRUE(lhsControlFlowStmts[i]->ASTStmt->equals(rhsControlFlowStmts[i]->ASTStmt.get()));
```
lukasm91:
```suggestion
    EXPECT_TRUE(lhsControlFlowStmts[i]->ASTStmt->equals(rhsControlFlowStmts[i]->ASTStmt.get()));
```
I think?

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    if(lhsControlFlowStmts[i]->StackTrace) {
+      if(rhsControlFlowStmts[i]->StackTrace) {
```
lukasm91: instead:
```suggestion
      ASSERT_TRUE(rhsControlFlowStmts[i]->StackTrace);
```

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,277 @@
...
+        for(int j = 0, jsize = lhsControlFlowStmts[i]->StackTrace->size(); j < jsize; ++j) {
+          EXPECT_EQ(*lhsControlFlowStmts[i]->StackTrace->at(j),
+                    *rhsControlFlowStmts[i]->StackTrace->at(j));
+        }
+      } else {
+        FAIL();
+      }
```
lukasm91: another else is needed if we don't have a stacktrace in lhs
```
} else {
  ASSERT_FALSE(rhsControlFlowStmts[i]->StackTrace);
}
```

test/integration-test/reference_iir/README.md
```diff
@@ -0,0 +1,7 @@
+## Updating References
+
+To update the references if iir was changed simply write them to disk, starting from the in-memory representation, e.g.
+```
+  createXXXStencilIIRInMemory(XXX_stencil_memory);
```
lukasm91: This can be updated now by pointing to the binary that should be called

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+          ASSERT_EQ(lhsDoMethod->getFields(), rhsDoMethod->getFields());
+          ASSERT_EQ(lhsDoMethod->getDependencyGraph(), rhsDoMethod->getDependencyGraph());
+        }
+      }
+    }
+  }
+}
+
+void readIIRFromFile(OptimizerContext& optimizer,
```
lukasm91:
```suggestion
std::shared_ptr<iir..StencilInstantiation> readIIRFromFile(OptimizerContext& optimizer, std::string const& fname);
```

test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -0,0 +1,277 @@
...
+        EXPECT_EQ(lhsStage->getGlobalVariables(), rhsStage->getGlobalVariables());
+        EXPECT_EQ(lhsStage->getGlobalVariablesFromStencilFunctionCalls(),
+                  rhsStage->getGlobalVariablesFromStencilFunctionCalls());
+        EXPECT_EQ(lhsStage->getExtents(), rhsStage->getExtents());
+        EXPECT_EQ(lhsStage->getRequiresSync(), rhsStage->getRequiresSync());
+
+        ASSERT_EQ(lhsStage->getChildren().size(), rhsStage->getChildren().size());
+
+        // checking each of the doMethods
+        for(int doMethodidx = 0, doMethodSize = lhsStage->getChildren().size();
```
lukasm91:
```suggestion
        for(int doMethodIdx = 0, doMethodSize = lhsStage->getChildren().size();
```

test/integration-test/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+dawn_add_integrationtest_impl(
+  NAME DawnIntegrationTest
+  SOURCES TestMain.cpp
+          TestIIRDeserializer.cpp
+          GenerateInMemoryStencils.cpp
+)
+
+yoda_add_executable(
+  NAME DawnRegenerateInMemoryStencils
```
lukasm91: Maybe `DawnUpdateReferences` or something like this? `RegenerateInMemoryStencils` doesn't sound like something that is written to disk.

test/unit-test/dawn/IIR/TestIIRSerializer.cpp
```diff
@@ -118,13 +123,17 @@ bool compareIIRs(iir::IIR* lhs, iir::IIR* rhs) {
             return false;
           }
         }
+      } else {
+        return false;
       }
-      return false;
+    } else {
+      IIR_EARLY_EXIT(!(rhsControlFlowStmts[i]->StackTrace));
```
lukasm91:
```suggestion
      return false;
```
for consistency

test/integration-test/GenerateInMemoryStencils.h
lukasm91: I would put the ifdefs

mroethlin:
launch pipeline

mroethlin:
## Technical Description 

This introduces testing for the `restoreIIR` function in `OptimizerContext` which computes the derived info of a stencil, given the stencils IIR. This is achieved by constructing a stencil in memory (A) and deserializing an equivalent stencil from disk (B). The in memory IIR, meta data and derived Info of A and B are then compared. 

Since those tests need to access the file system, they are considered integration tests instead of unit tests. These are the first integration tests for dawn. Build files have been adapted accordingly. 

Currently, to ensure equivalent ids between stencils from memory as well as from disk the UUID singleton needs to be reset. This is of course bad practice. The correct way to compare this would be to first try to find a bijective mapping between both id sets. To be discussed if this should go into a different PR are handled herein.

PR #289, opened 2019-09-30
Add ReduceOverNeighbor

havogt:


src/dawn/Optimizer/AccessComputation.cpp
```diff
@@ -256,6 +256,7 @@ class AccessMapper : public iir::ASTVisitor {
     // the block statements as the if/then/else block of the if-statement has been collapsed into
     // one single
     // vector of block statements
+    appendNewAccesses();
```
havogt: This breaks gtclang.

lukasm91: I am wondering why gtclang was green. I pushed a fix (I hope), but I admit that I don't fully understand how if statements and block statements interact. But block statements are broken in gtclang anyway, so I only fix them for iir as far as they compile and run now.

lukasm91:
launch jenkins

lukasm91:
launch pipeline

havogt:
launch pipeline

lukasm91:
launch pipeline

PR #290, opened 2019-09-30
Enable c++17

havogt:
substitute additions to <memory> in STLExtras.h with the actual c++14 std:: versions


havogt:
launch jenkins

PR #292, opened 2019-10-01
Switch to std::optional

havogt:
First c++17 feature to test compilers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,7 @@ include(DawnDefinitions)
 include(DawnMacros)
 
 # Set C++ standard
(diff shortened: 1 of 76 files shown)


havogt:
launch pipeline

havogt:
launch pipeline

twicki:
check the pipeline before the merge, seems to not be ok yet:
image

havogt:
I believe it should be ignored as there is a PR chain down to clang-gridtools for switching the compiler.

havogt:
Can you double-check? Then I would like to merge now.

twicki:
Ah yeah, seems good.

PR #302, opened 2019-10-04
Move accesses and AccessID metadata

### LINKED ISSUE #291: BlockStmt broken in Backend
The block statement is broken in the backend.

Trying to add a block statement  in https://github.com/MeteoSwiss-APN/dawn/blob/master/test/unit-test/dawn-c/TestCompiler.cpp#L77  fails:
```
  auto stencil_instantiation = b.build(
      "generated",
      b.stencil(b.multistage(dawn::iir::LoopOrderKind::LK_Parallel,
                             b.stage(b.vregion(dawn::sir::Interval::Start, dawn::sir::Interval::End,
                                               b.block(b.stmt(b.assignExpr(b.at(out_f), b.at(in_f)))))))));
```

with the following assertion:

```
Assertion failed: `access'       
Function: 'virtual void dawn::iir::DoMethod::updateLevel()'
Location: /home/lukas/documents/work/dawn/src/dawn/IIR/DoMethod.cpp:135
Aborted (core dumped)  
```

The problem could be fixed in this case by adding proper `appendNewAccesses` and `removeLastChildAccesses` for the `BlockStmt` in `AccessComputation`, but this again breaks the if-stmt which seems to abuse the `BlockStmt` in some way.

Relates https://github.com/MeteoSwiss-APN/gtclang/issues/164

Stagno:
## Technical Description

- Move information about accesses (`CallerAccesses` and `CalleeAccesses`) from `StatementAccessesPair` inside `iir::IIRStmtData` (GT-75)
	- Fix all usages
- Fix `StatementAccessesPair::clone()` which was not cloning the underlying statement (necessary now since the accesses information are inside the stmt)
- Since we now clone `Stmt`s we would also need to add records to `StencilMetaInformation::StmtIDToAccessIDMap_` and to `StencilMetaInformation::ExprDToAccessIDMap_` (children expressions are also cloned). Instead of doing that, those maps are removed from `StencilMetaInformation` and an `AccessID` data field is added to `VarDeclStmt`, `VarAccessExpr`, `FieldAccessExpr`, `LiteralAccessExpr` (GT-76).
    - Add `iir::VarDeclStmtData` (inherits from `iir::IIRStmtData`) as data object of `VarDeclStmt` (adds `AccessID` field).
    - Add abstract base`AccessExprData` as (initialized-on-the-fly) data object of `VarAccessExpr`, `FieldAccessExpr` and `LiteralAccessExpr`.
    - Add concrete `iir::IIRAccessExprData` with `AccessID` field.
    - Add helper function `iir::getAccessIDFromExpr` asserting on the correct type of expr passed and returning the AccessID from the expr's data
    - Fix all usages
- Switch from `clone` to `std::move` for `StatementAccessPair`/`DoMethod`/`Stage` in
	- `Stage::addDoMethod`
	- `DetectInlineCandiates::getNewStatementAccessesPairs`
	- `ReoderStrategyGreedy::reorder`
- Fixing various bugs related to cloning
- Fix python examples
- Add serialization/deserialization of some stmt/expr data
- Make `ProtoStmtBuilder` parametrized on the data type

### Resolves / Enhances

GT-75 and GT-76 (unplanned)
fixes #291 

### Dependencies

none


Stagno:
launch jenkins

dawn/src/dawn/IIR/ASTStmt.h
```diff
@@ -53,9 +72,24 @@ std::shared_ptr<ast::ReturnStmt> makeReturnStmt(Args&&... args) {
...
+struct VarDeclStmtData : public IIRStmtData {
+
+  VarDeclStmtData() = default;
+  VarDeclStmtData(const VarDeclStmtData&);
+
+  bool operator==(const VarDeclStmtData&);
+  bool operator!=(const VarDeclStmtData&);
+
+  /// ID of the variable declared by the statement
+  std::optional<int> AccessID;
```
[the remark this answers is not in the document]
Stagno: will write `iir::getAccessIDFromStmt()` ... I checked the naming conventions for other public fields of classes/structs and `AccessID` is correct (see structs in `SIR/SIR.h`)

dawn/src/dawn/SIR/proto/SIR/statements.proto
```diff
@@ -461,7 +478,8 @@ message VarAccessExpr {
   Expr index = 2;         // Is it an array access (i.e var[2])?
   bool is_external = 3;   // Is this an access to a external variable (e.g a global)?
   SourceLocation loc = 4; // Source location
-  int32 ID = 5;           // ID of the Expr
+  AccessExprData data = 5; // Access data
```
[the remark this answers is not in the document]
Stagno: in protobuf3 `data` is already optional ... since there won't be any SIR data for expressions, do you think it is still useful to put `OneOf` between `AccessExprData` and empty?

dawn/src/dawn/Optimizer/PassSSA.cpp
```diff
@@ -48,22 +48,25 @@ bool PassSSA::run(const std::shared_ptr<iir::StencilInstantiation>& stencilInsta
...
-        if(iir::ExprStmt* stmt =
-               dyn_cast<iir::ExprStmt>(stmtAccessesPair->getStatement().get()))
+        if(iir::ExprStmt* stmt = dyn_cast<iir::ExprStmt>(stmtAccessesPair->getStatement().get()))
           assignment = dyn_cast<iir::AssignmentExpr>(stmt->getExpr().get());
 
         std::vector<int> AccessIDsToRename;
 
         for(const std::pair<int, iir::Extents>& readAccess :
-            stmtAccessesPair->getAccesses()->getReadAccesses()) {
+            stmtAccessesPair->getStatement()
```
lukasm91: factor out common code here and beloow (getting `IIRStmtData`)

dawn/src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -248,17 +249,18 @@ class GlobalFieldCacher {
...
+      auto readAccessIterator = callerAccesses->getReadAccesses().find(AccessID);
+      if(readAccessIterator != callerAccesses->getReadAccesses().end()) {
         return true;
       }
       // If we did not find a read statement so far, we have  a write first and do not need to
       // fill the cache
-      auto wirteAccessIterator =
-          stmtAccessesPair->getCallerAccesses()->getWriteAccesses().find(AccessID);
-      if(wirteAccessIterator != stmtAccessesPair->getCallerAccesses()->getWriteAccesses().end()) {
+      auto wirteAccessIterator = callerAccesses->getWriteAccesses().find(AccessID);
```
lukasm91: typo (`writeAccessIterator`)

dawn/src/dawn/Optimizer/PassSetNonTempCaches.cpp
```diff
@@ -270,10 +272,12 @@ class GlobalFieldCacher {
...
         iterateIIROver<iir::StatementAccessesPair>(*multiStagePrt_)) {
 
+      const auto& callerAccesses =
+          stmtAccessesPair->getStatement()->getData<iir::IIRStmtData>().CallerAccesses;
+
       // If we find a write-statement we exit
-      auto wirteAccessIterator =
-          stmtAccessesPair->getCallerAccesses()->getWriteAccesses().find(AccessID);
-      if(wirteAccessIterator != stmtAccessesPair->getCallerAccesses()->getWriteAccesses().end()) {
+      auto wirteAccessIterator = callerAccesses->getWriteAccesses().find(AccessID);
```
lukasm91:
```suggestion
      auto writeAccessIterator = callerAccesses->getWriteAccesses().find(AccessID);
```

dawn/src/dawn/Optimizer/PassStageReordering.cpp
```diff
@@ -56,7 +56,7 @@ bool PassStageReordering::run(
     auto newStencil = strategy->reorder(stencilInstantiation.get(), stencilPtr, context_);
 
     stencilInstantiation->getIIR()->replace(stencilPtr, newStencil, stencilInstantiation->getIIR());
-
+    // TODO why is this pointer still used? (it was replaced in the previous statement)
```
lukasm91: is this TODO intended? This comment has no code changes around

dawn/src/dawn/Optimizer/PassTemporaryType.cpp
```diff
@@ -141,8 +142,12 @@ bool PassTemporaryType::run(const std::shared_ptr<iir::StencilInstantiation>& in
         }
       };
 
-      processAccessMap(statementAccessesPair->getAccesses()->getWriteAccesses());
-      processAccessMap(statementAccessesPair->getAccesses()->getReadAccesses());
+      processAccessMap(statementAccessesPair->getStatement()
```
lukasm91: factor out common part

dawn/src/dawn/Optimizer/Replacing.cpp
```diff
@@ -79,8 +80,7 @@ void replaceFieldWithVarAccessInStmts(
 
       iir::replaceOldExprWithNewExprInStmt(stmt, oldExpr, newExpr);
 
-      metadata.insertExprToAccessID(newExpr, AccessID);
-      metadata.eraseExprToAccessID(oldExpr);
+      newExpr->getData<iir::IIRAccessExprData>().AccessID = std::make_optional(AccessID);
```
lukasm91: I really don't like the interface. It might make sense internally, but acccessing doesn't make sense, because you wouldn't expect an optional (because you are never interested in whether accessID is set or not; if it is not set, it might be simply incorrect to access it, or it should be computed on the fly); and it doesn't make sense when setting the accessID, because it doesn't make sense to set nullopt

dawn/src/dawn/Optimizer/TemporaryHandling.cpp
```diff
@@ -70,19 +71,23 @@ void promoteLocalVariableToTemporaryField(iir::StencilInstantiation* instantiati
...
+    fieldAccessExpr->getData<iir::IIRAccessExprData>().AccessID = std::make_optional(accessID);
     auto assignmentExpr =
         std::make_shared<iir::AssignmentExpr>(fieldAccessExpr, varDeclStmt->getInitList().front());
     auto exprStmt = iir::makeExprStmt(assignmentExpr);
 
     // Replace the statement
     exprStmt->getData<iir::IIRStmtData>().StackTrace =
         oldStatement->getData<iir::IIRStmtData>().StackTrace;
+    exprStmt->getData<iir::IIRStmtData>().CallerAccesses =
+        oldStatement->getData<iir::IIRStmtData>().CallerAccesses;
```
lukasm91: factor out `exprData` somewhere above and access through that?

dawn/src/dawn/Unittest/IIRBuilder.h
```diff
@@ -91,11 +92,11 @@ class IIRBuilder {
     DAWN_ASSERT(si_);
     auto v_str = std::to_string(std::forward<T>(v));
     int acc = si_->getMetaData().insertAccessOfType(iir::FieldAccessType::FAT_Literal, v_str);
-    auto expr = std::make_shared<iir::LiteralAccessExpr>(
+    std::shared_ptr<iir::LiteralAccessExpr> expr = std::make_shared<iir::LiteralAccessExpr>(
```
lukasm91: isn't `auto` here a perfect fit, because the type is completely obvious from the RHS?

dawn/test/integration-test/TestIIRDeserializer.cpp
```diff
@@ -265,4 +267,4 @@ TEST(IIRDeserializerTest, LaplStencil) {
   compareIIRs(lap_stencil_from_file, lap_stencil_memory);
 }
 
-} // anonymous namespace
\ No newline at end of file
+} // anonymous namespace
```
lukasm91: Make sure that you properly use clang-format (and it should not be somehow interact with the IDE adding the new line). I thought it should remove the new line at the end of the file.

dawn/src/dawn/IIR/ASTStmt.h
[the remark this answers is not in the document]
lukasm91: Or those are also wrong, because the naming convention that was shared initially was different: https://meteoswiss-apn.github.io/dawn/coding.html#style-issues

dawn/src/dawn/SIR/proto/SIR/statements.proto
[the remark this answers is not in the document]
lukasm91: I agre maybe it is only useful if we had SIR data. Do you think it still makes sense to keep data in a separate message?

lukasm91:
Might fix #291  (currently it does). If it does, could you modify the test given in the link accordingly?

dawn/src/dawn/IIR/ASTStmt.h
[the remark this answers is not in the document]
Stagno: I am not sure if that's still up to date... It seems to me that there's this (unwritten) convention of having public members of classes starting with a capital letter ... for consistency I would keep `AccessID` until a decision is taken for all

dawn/src/dawn/SIR/proto/SIR/statements.proto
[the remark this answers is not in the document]
Stagno: To me it's still (kind of) useful to keep a logical separation between what is proper AST information and what is a "decoration" added by components processing it

dawn/src/dawn/Optimizer/Replacing.cpp
Stagno: So, this pattern is the same used for other data fields such as `CalleeAccesses`, `CallerAccesses` and `StackTrace` (even before the switch to data containers, where we used pointers instead of optional): if the datum is not yet calculated, the field is set to `nullptr`/`nullopt`. I agree that an interface around it would be nicer, but it would need to be public (can't `friend` every class that needs to write for example), and would necessarily consist of: `getAccessID()` (we already have `iir::getAccessIDFromExpr()`), `setAccessID(int)`, `unsetAccessID()` (needed by `AccessComputation.cpp:97` because a statement could be cloned and the clone brought out of a function, therefore `CalleeAccesses` must be unset), `isAccessIDSet()` ... which let's every user do the same things as exposing directly `AccessID`. Then I wonder if it is worth to switch to such interface and whether it is in the scope of this PR.

dawn/src/dawn/CodeGen/CXXNaive/ASTStencilBody.cpp
```diff
@@ -32,23 +33,21 @@ ASTStencilBody::~ASTStencilBody() {}
...
-    return metadata_.getFieldNameFromAccessID(metadata_.getAccessIDFromExpr(expr));
+    return metadata_.getFieldNameFromAccessID(iir::getAccessIDFromExpr(expr));
 }
 
 int ASTStencilBody::getAccessID(const std::shared_ptr<iir::Expr>& expr) const {
-  if(currentFunction_)
-    return currentFunction_->getAccessIDFromExpr(expr);
-  else
-    return metadata_.getAccessIDFromExpr(expr);
+  return iir::getAccessIDFromExpr(expr);
```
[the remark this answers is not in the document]
Stagno: What about removing `ASTStencilBody::getAccessID()` instead? Do we really need the interface? It made sense before, because the information could have come from different contexts, but now it is always inside the stmt/expr and we can use the global helper functions.

dawn/src/dawn/IIR/ASTStmt.cpp
```diff
@@ -13,16 +13,89 @@
...
+std::optional<Extents> computeMaximumExtents(Stmt& stmt, const int accessID) {
+  std::optional<Extents> extents;
+
+  const auto& callerAccesses = stmt.getData<IIRStmtData>().CallerAccesses;
+
+  if(callerAccesses->hasReadAccess(accessID) || callerAccesses->hasWriteAccess(accessID)) {
+    extents = std::optional<Extents>();
+
+    if(callerAccesses->hasReadAccess(accessID)) {
+      if(!extents)
```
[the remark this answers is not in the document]
Stagno: noticed now this comment, answered above

dawn/src/dawn/IIR/ASTStmt.cpp
```diff
@@ -13,16 +13,89 @@
...
+  }
+
+  for(auto const& child : stmt.getChildren()) {
+    auto childExtent = computeMaximumExtents(*child, accessID);
+    if(!childExtent)
+      continue;
+    if(extents)
+      extents->merge(*childExtent);
+    else
+      extents = childExtent;
```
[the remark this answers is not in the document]
Stagno: `merge` returns void, so can't assign to `extents` ... but it can become
```
if(auto childExtent = computeMaximumExtents(*child, accessID))
      extents ? extents->merge(*childExtent) : extents = childExtent;
```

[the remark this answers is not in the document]
Stagno: Tried it, but the compiler doesn't like it. Need if-else.

dawn/src/dawn/Optimizer/PassInlining.cpp
```diff
@@ -155,9 +156,11 @@ class Inliner : public iir::ASTVisitor {
 
       // Register the variable
       metadata_.addAccessIDNamePair(AccessID, returnVarName);
-      metadata_.addStmtToAccessID(newStmt, AccessID);
+      newStmt->getData<iir::VarDeclStmtData>().AccessID = std::make_optional(AccessID);
       // TODO recheck this
-      metadata_.insertExprToAccessID(newExpr_, AccessID);
+      std::dynamic_pointer_cast<iir::VarAccessExpr>(newExpr_)
+          ->getData<iir::IIRAccessExprData>()
```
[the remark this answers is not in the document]
Stagno: Mostly is something like `node->getData<DataType>().MemberField...` ... we could have helper functions for each member field of each data structure, so it would become `getMemberField(node)` ... but wouldn't it pollute the `dawn::iir` namespace with functions?

dawn/src/dawn/Optimizer/PassInlining.cpp
```diff
@@ -388,14 +384,14 @@ class DetectInlineCandiates : public iir::ASTVisitorForwarding {
...
+  void processStatment(std::unique_ptr<iir::StatementAccessesPair>& stmtAccesesPair) {
     // Reset the state
     inlineCandiatesFound_ = false;
-    oldStmtAccessesPair_ = stmtAccesesPair->clone();
+    oldStmtAccessesPair_ = &stmtAccesesPair;
     newStmtAccessesPairs_.clear();
 
     // Detect the stencil functions suitable for inlining
-    oldStmtAccessesPair_->getStatement()->accept(*this);
+    (*oldStmtAccessesPair_)->getStatement()->accept(*this);
```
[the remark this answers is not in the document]
Stagno: See line 406.
The idea here is to steal the `unique_ptr` (pointed by `oldStmtAccessesPair_`). Most of the time you don't want to steal, but sometimes you do.
This happens in 2 phases: in `processStatement` you identify a candidate and if `getNewStatementAccessesPairs` is executed you move the `unique_ptr`.

dawn/src/dawn/Optimizer/Replacing.cpp
lukasm91: Yes maybe not. If we add `iir::getAccessIDFromExpr`, we can hide most of them, so that's fine for me.

dawn/src/dawn/IIR/ASTStmt.cpp
```diff
@@ -13,16 +13,89 @@
...
+//===------------------------------------------------------------------------------------------===//
+//     computeMaximumExtents
+//===------------------------------------------------------------------------------------------===//
+
+std::optional<Extents> computeMaximumExtents(Stmt& stmt, const int accessID) {
+  std::optional<Extents> extents;
+
+  const auto& callerAccesses = stmt.getData<IIRStmtData>().CallerAccesses;
+
+  if(callerAccesses->hasReadAccess(accessID) || callerAccesses->hasWriteAccess(accessID)) {
```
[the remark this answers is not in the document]
lukasm91: 80 might go into the then case, if callerAccess has no ReadAccess. In that case, extents stays empty and we go into then.

Maybe something like this would make it clearer?
```
bool writeAccess = callerAccesses->hasWriteAccess(accessID);
bool readAccess = callerAccesses->hasWriteAccess(accessID);
if (readAccess && writeAccess)
    extents = callerAccesses->getReadAccess(accessID)->merge(callerAccesses->getWriteAccess(accessID));
else if (readAccess)
    extents = callerAccesses->getReadAccess(accessID);
else if (writeAccess)
    extents = callerAccesses->getWriteAccess(accessID);
```

dawn/src/dawn/IIR/ASTStmt.h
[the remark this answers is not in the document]
lukasm91: okay. Let's try to remember bringing it up :)

dawn/src/dawn/CodeGen/CXXNaive/ASTStencilBody.cpp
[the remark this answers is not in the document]
lukasm91: yes I think you're right. When you add the global function, you could probably also rename it to `iir::getAccessID`, because `FromStmt` and `FromExpr` does not add any information (as it is only a hint for the argument type that you pass anyway)

dawn/src/dawn/IIR/ASTStmt.cpp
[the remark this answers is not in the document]
lukasm91: Yes because types are not compatible; and I also wouldn't do it. In case you would want to do it (what you really shouldn't):
```
extents = extents ? (extents->merge(*childExtent), extents) : childExtent;
```
This should work.

But I think as long as merge is inplace, we need to have an if.  I would still put the `auto childExtent =` into the if, because then we don't need the `continue` which usually makes the code-flow harder to understand.

dawn/src/dawn/Optimizer/PassInlining.cpp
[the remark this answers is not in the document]
lukasm91: Hmm yes, we can also postpone this decision and ask others for their opinion. I am not strongly against global functions, especially if we are careful with namespaces. A global function might be a better abstraction in this case, because in this function chain there is significant potential that something changes in the future and we need to go through all of them.

dawn/src/dawn/Optimizer/PassInlining.cpp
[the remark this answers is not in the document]
lukasm91: I see the point. I think again, this is very complex code flow, but this is not your fault because it was already before. It looks like a weird pattern, that this class has three functions, because the three functions are always used together and in the same order (`c.process(); if (c.hasResult()) ret = c.getResult();`). For this three functions we have some internal state like `inlineCandiatesFound_`, `oldStmtAccessesPair_` and `newStmtAccessesPairs_`, but there is also some state about several calls to these three functions.

So *if* you want to clean it up a bit, I would try to merge the three functions. But you don't need to do it, because it was already like this before.

dawn/src/dawn/CodeGen/CXXNaive/ASTStencilBody.cpp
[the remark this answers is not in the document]
Stagno: ok

dawn/src/dawn/IIR/ASTStmt.cpp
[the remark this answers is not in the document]
Stagno: Problem is that we need to work on a copy of extents as we cannot use non-cost `merge` on `const` Extents returned by `getReadAccess(accessID)` ... also `extents` is `std::optional`...
Can't come up with nothing better than:
```
if(callerAccesses->hasReadAccess(accessID))
    extents = std::make_optional(callerAccesses->getReadAccess(accessID));

  if(callerAccesses->hasWriteAccess(accessID)) {
    if(extents)
      extents->merge(callerAccesses->getWriteAccess(accessID));
    else
      extents = std::make_optional(callerAccesses->getWriteAccess(accessID));
  }
```

dawn/src/dawn/Optimizer/PassInlining.cpp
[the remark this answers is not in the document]
Stagno: Somehow I like the idea of separating the pass into an analysis step that determines the operation to do (inline or don't) and an "apply decision" step. So i would still keep the separation, but find a way to improve the information passing from the first step to second ... in this case `std::unique_ptr<iir::StatementAccessesPair>*` could be substituted by some other way of referencing to the statement, such as an index over the children of the current `doMethod` or an iterator?

dawn/src/dawn/Optimizer/PassStageReordering.cpp
Stagno: It is a general TODO, it refers to the statement below: stencil pointed by `stencilPtr` has been replaced in the `IIR`, but it's `update` method is called anyways . What do you think?

dawn/src/dawn/Optimizer/TemporaryHandling.cpp
Stagno: Since `oldStatement` is going to be replaced, better to `std::move` the data:
```
exprStmt->getData<iir::IIRStmtData>() = std::move(oldStatement->getData<iir::IIRStmtData>());
```

dawn/src/dawn/Unittest/IIRBuilder.h
Stagno: yes

dawn/test/integration-test/TestIIRDeserializer.cpp
Stagno: Probably the IDE's clang-format configuration is messed up ... but these kind of problems should be solved with the pre-commit hook right?

dawn/src/dawn/Unittest/IIRBuilder.h
Stagno: compiler interprets `auto` as `std::shared_ptr<iir::Expr>` instead of `std::shared_ptr<iir::LiteralAccessExpr>` ... therefore cannot get data

Stagno:
launch jenkins

Stagno:
launch jenkins

dawn/src/dawn/Optimizer/PassInlining.cpp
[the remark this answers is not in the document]
lukasm91: I propose if we want to change the current implementation, we can do it in a new PR because this one is already pretty large. That's then also easier to discuss. Do you think that's a reasonable approach?

dawn/src/dawn/Optimizer/PassManager.cpp
```diff
@@ -13,6 +13,7 @@
 //===------------------------------------------------------------------------------------------===//
 
 #include "dawn/Optimizer/PassManager.h"
+#include "dawn/IIR/ASTVisitor.h"
```
lukasm91: weird that you need to add an include in a file that you did not touch

dawn/src/dawn/Unittest/IIRBuilder.h
lukasm91: hmm, why should it do that? `std::make_shared<X>` always returns `std::shared_ptr<X>`. If `X` is derived from `Y`, it can be cast to `std::shared_ptr<Y>`, but if you just put `auto` there is no reason why the compiler should convert it.

dawn/test/unit-test/dawn/IIR/TestIIRSerializer.cpp
```diff
@@ -333,13 +286,62 @@ TEST_F(IIRSerializerTest, IIRTests) {
...
+                                                   ->getChild(0)
+                                                   ->getChild(0)
+                                                   ->getStatement())
+          ->getExpr());
+  deserializedVarAccessExpr->getData<iir::IIRAccessExprData>().AccessID =
+      std::make_optional<int>(50);
+  IIR_EXPECT_NE(deserialized, referenceInstantiaton);
+  deserialized = serializeAndDeserializeRef();
+  auto deserializedVarDeclStmt = std::dynamic_pointer_cast<iir::VarDeclStmt>(deserialized->getIIR()
+                                                                                 ->getChild(0)
```
lukasm91: that looks very verbose... is this really what we want to do here?

dawn/src/dawn/Optimizer/PassInlining.cpp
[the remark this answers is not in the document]
Stagno: Yes that's reasonable, I'll open an issue on Github after the PR is merged

dawn/src/dawn/Unittest/IIRBuilder.h
Stagno: So ... When compiling with gcc the error wasn't clear, it somehow suggested that it didn't find the `getData` templated method, so I thought it interpreted auto as `std::shared_ptr<iir::Expr>` ... Now I tried compiling with clang and the error says:
```
/home/giacomo/dawn/dawn/src/dawn/Unittest/IIRBuilder.h:99:11: error: use 'template' keyword to treat 'getData' as a dependent template name
    expr->getData<IIRAccessExprData>().AccessID = std::make_optional(acc);
          ^
          template
```
... now I know what to do :)

dawn/src/dawn/IIR/Accesses.cpp
```diff
@@ -153,17 +138,23 @@ std::string Accesses::reportAccesses(const StencilFunctionInstantiation* stencil
...
+  std::stringstream ss;
+  std::string indent(initialIndent, ' ');
 
-std::string Accesses::toString(const StencilFunctionInstantiation* stencilFunc,
-                               std::size_t initialIndent) const {
-  return toStringImpl(
-      [&stencilFunc](int AccessID) { return stencilFunc->getNameFromAccessID(AccessID); },
-      initialIndent, writeAccesses_, readAccesses_);
+  auto printMap = [&](const std::unordered_map<int, Extents>& map) {
+    for(auto it = map.begin(), end = map.end(); it != end; ++it)
```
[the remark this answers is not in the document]
Stagno: That's really nice!

dawn/src/dawn/Optimizer/PassManager.cpp
Stagno: That's really weird ... I don't remember adding this, maybe it's a side effect of merging?

dawn/test/unit-test/dawn/IIR/TestIIRSerializer.cpp
Stagno: I think testing data serialization is nice to have ... but I couldn't come up with a shorter way to write that ugly chain: `getIIR()
                                                                                 ->getChild(0)
                                                                                 ->getChild(0)
                                                                                 ->getChild(0)
                                                                                 ->getChild(0)
                                                                                 ->getChild(1)
                                                                                 ->getStatement()`

Stagno:
launch jenkins

dawn/test/unit-test/dawn/IIR/TestIIRSerializer.cpp
Stagno: I've come up with this to get the first child: `iterateIIROver<iir::StatementAccessesPair>(deserialized->getIIR()).begin()` (`++` for the second) ... not sure if improves readability

lukasm91: No it's okay, I was just wondering whether this was intentional. What might improve readability, is to do it in several steps:
```
auto xyz = getIIR()->getChild(0);
auto abc = xyz->getChild(0);
...
```
because then, it would be clear what all these `getChild` are. What do you think?

Stagno: yes can do

Stagno:
launch jenkins

Stagno:
> Can you merge with master and re-run. The error seems to be a temporary slurm problem.

I get failing tests after I merged master. Will fix those and then re-run.

Stagno:
launch jenkins

Stagno:
launch jenkins

PR #307, opened 2019-10-05
add timers to GT-codegen

twicki:
## Technical Description

This PR adds start and pause instructions to the GT-Backend

### Resolves / Enhances

Resolves the fact that we currently do not  have daily profiling of the GT-Backend

### Dependencies

This PR is independent


twicki:
closed in favor of something simpler

PR #310, opened 2019-10-07
Add atlas interface

lukasm91:
- defines the current requirements of the interface (currently in `src/dawn/CodeGen/CXXNaive-ico/CXXNaiveCodeGen.cpp`)
- Adapts the `mylib` interface to this interface
- Implements the interface for `atlas` and adds a driver code
- Generalize the unstructured "CodeGenPlayground" to  take into account that different nodes might have different number of neighbours
- Fix initialization in the periodic unstructured example
- (unrelated) add bundle-folder to ignorelist of `clang-format-all.sh`


lukasm91:
launch jenkins

dawn/prototype/README
```diff
@@ -3,5 +3,7 @@ Compiling and running the prototype (until we have a proper setup)
 1. Compile Dawn
 2. cd <path/to/dawn> && ./bundle/build/dawn-prefix/src/dawn-build/bin/unittest/DawnCUnittest --gtest_also_run_disabled_tests 2>./prototype/generated.hpp && clang-format -i ./prototype/generated.hpp
    --> This will run the disabled CodeGenPlayground test that generated and outputs code
-3. cd prototype && g++ -o out driver.cpp grid.cpp -o out -std=c++17 -I <path/to/gtclang>/src; ./out
-4. paraview of_..vtk
+3. For "lukas_lib": cd prototype && g++ -o out driver.cpp grid.cpp -o out -std=c++17 -I<path/to/gtclang>/src; ./out
```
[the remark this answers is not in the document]
havogt: Also `my_interface.hpp` doesn't exist. How is it supposed to work?

dawn/prototype/atlas_interface.hpp
```diff
@@ -0,0 +1,97 @@
...
+} // namespace utility
+
+namespace atlasInterface {
+
+struct atlasTag {};
+
+template <typename T>
+class Field {
+public:
+  T operator()(int f) const { return atlas_field_(f, 0); }
```
havogt: Why not return `const&`

dawn/src/dawn/CodeGen/CodeGen.cpp
```diff
@@ -29,10 +29,12 @@ std::string CodeGen::generateGlobals(stencilInstantiationContext& context,
                                      std::string outer_namespace_, std::string inner_namespace_) {
 
   std::stringstream ss;
-  Namespace outerNamespace(outer_namespace_, ss);
   std::string globals = generateGlobals(context, inner_namespace_);
-  ss << globals;
-  outerNamespace.commit();
+  if(globals != "") {
```
havogt: I propose std::optional instead of an empty string

lukasm91: `globals` is generated code and we want to wrap it into a namespace. The namespace is of course not needed if there is no code - that would be just an empty namespace. So the code expresses quite well what we want to do.

lukasm91:
launch jenkins

PR #311, opened 2019-10-07
Refactor sir::Value using std::variant

### LINKED ISSUE #293: (2) Move away from our own value class to std::variant [2]
Now that we have c++17 support, the Value class in  https://github.com/MeteoSwiss-APN/dawn/blob/e37778538ef09a111f56aaf247cdd7e1cbf2b567/src/dawn/SIR/SIR.h#L280 is depricated

jdahm:
This removes the custom sir::Value in favor of using the std::variant container. Fixes #293.

### DIFF
--- a/dawn/src/dawn/SIR/SIR.h
+++ b/dawn/src/dawn/SIR/SIR.h
@@ -279,16 +279,26 @@ struct Stencil : public dawn::NonCopyable {
 ///
 /// @ingroup sir
 
-struct Value : NonCopyable {
+class Value : NonCopyable {
+public:
   enum TypeKind { Boolean = 0, Integer, Double, String };
 
+private:
+  template <typename T>
+  struct TypeInfo {};
+
+  std::optional<std::variant<bool, int, double, std::string>> value_;
+  bool is_constexpr_;
+  TypeKind type_;
+
+public:
   template <class T>
   explicit Value(T&& value)
       : value_{std::forward<T>(value)}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
 
   template <class T>
   explicit Value(T value, bool is_constexpr)
-      : value_{value}, is_constexpr_{is_constexpr}, type_{TypeInfo<T>::Type} {}
+      : value_{std::forward<T>(value)}, is_constexpr_{is_constexpr}, type_{TypeInfo<T>::Type} {}
 
   explicit Value(TypeKind type) : value_{}, is_constexpr_{false}, type_{type} {}
 
@@ -329,14 +339,6 @@ struct Value : NonCopyable {
     valueJson["value"] = toString();
     return valueJson;
   }
-
-  template <typename T>
-  struct TypeInfo {};
-
-private:
-  std::optional<std::variant<bool, int, double, std::string>> value_;
-  bool is_constexpr_;
-  TypeKind type_;
 };
 
 template <>
--- a/dawn/src/dawn/SIR/SIR.h
+++ b/dawn/src/dawn/SIR/SIR.h
@@ -25,11 +25,11 @@
 #include "dawn/Support/Type.h"
 #include <iosfwd>
 #include <memory>
+#include <optional>
 #include <string>
 #include <unordered_map>
-#include <vector>
 #include <variant>
-#include <optional>
+#include <vector>
 
 namespace dawn {
 
@@ -284,11 +284,11 @@ struct Value : NonCopyable {
 
   template <class T>
   explicit Value(T&& value)
-    : value_{value}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
+      : value_{std::forward<T>(value)}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
 
   template <class T>
   explicit Value(T value, bool is_constexpr)
-    : value_{value}, is_constexpr_{is_constexpr}, type_{TypeInfo<T>::Type} {}
+      : value_{value}, is_constexpr_{is_constexpr}, type_{TypeInfo<T>::Type} {}
 
   explicit Value(TypeKind type) : value_{}, is_constexpr_{false}, type_{type} {}
 
@@ -302,7 +302,7 @@ struct Value : NonCopyable {
   static BuiltinTypeID typeToBuiltinTypeID(TypeKind type);
 
   /// Convert the value to string
-  std::string toString() const { return std::string{typeToString(type_)}; }
+  std::string toString() const { return typeToString(type_); }
 
   /// @brief Check if value is set
   bool has_value() const { return value_.has_value(); }
@@ -330,7 +330,8 @@ struct Value : NonCopyable {
     return valueJson;
   }
 
-  template <typename T> struct TypeInfo {};
+  template <typename T>
+  struct TypeInfo {};
 
 private:
   std::optional<std::variant<bool, int, double, std::string>> value_;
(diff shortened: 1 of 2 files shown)


dawn/src/dawn/SIR/SIR.h
```diff
@@ -272,28 +274,26 @@ struct Stencil : public dawn::NonCopyable {
...
 
 struct Value : NonCopyable {
   enum TypeKind { Boolean = 0, Integer, Double, String };
 
   template <class T>
-  explicit Value(T&& value) : isConstexpr_(false) {
-    valueImpl_ = std::make_unique<ValueImpl<std::decay_t<T>>>(std::forward<T>(value));
-  }
+  explicit Value(T&& value)
+    : value_{value}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
```
lukasm91:
```suggestion
    : value_{std::forward<T>(value)}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
```

dawn/src/dawn/SIR/SIR.h
```diff
@@ -302,89 +302,59 @@ struct Value : NonCopyable {
   static BuiltinTypeID typeToBuiltinTypeID(TypeKind type);
 
   /// Convert the value to string
-  std::string toString() const;
+  std::string toString() const { return std::string{typeToString(type_)}; }
 
   /// @brief Check if value is set
```
lukasm91: clang-format in the end

dawn/src/dawn/SIR/SIR.h
```diff
@@ -302,89 +302,59 @@ struct Value : NonCopyable {
   static BuiltinTypeID typeToBuiltinTypeID(TypeKind type);
 
   /// Convert the value to string
-  std::string toString() const;
+  std::string toString() const { return std::string{typeToString(type_)}; }
```
lukasm91:
```suggestion
  std::string toString() const { return {typeToString(type_)}; }
```
this should also work, shouldn't it?

dawn/src/dawn/SIR/SIR.h
```diff
@@ -302,89 +302,59 @@ struct Value : NonCopyable {
...
-  static const Value::TypeKind Type = Value::Double;
+template <> struct Value::TypeInfo<double> {
+  static const TypeKind Type = Double;
 };
 
-template <>
-struct Value::TypeInfo<std::string> {
-  static const Value::TypeKind Type = Value::String;
+template <> struct Value::TypeInfo<std::string> {
+  static const TypeKind Type = String;
```
lukasm91:
```suggestion
  static constexpr TypeKind Type = String;
```

dawn/src/dawn/SIR/SIR.h
jdahm: Good catch 👍

dawn/src/dawn/SIR/SIR.h
jdahm: I was being explicit, but yes that may trigger a double copy, better to use the implicit construct.

dawn/src/dawn/SIR/SIR.h
jdahm: I was being explicit, but yes that may trigger a double copy, better to use the implicit construct.

dawn/src/dawn/SIR/SIR.h
jdahm: I had this originally, but in the context it's used the constexpr isn't helpful. I suppose it's still good form to have that.

dawn/src/dawn/SIR/SIR.h
```diff
@@ -284,11 +284,11 @@ struct Value : NonCopyable {
 
   template <class T>
   explicit Value(T&& value)
-    : value_{value}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
+      : value_{std::forward<T>(value)}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
 
   template <class T>
   explicit Value(T value, bool is_constexpr)
-    : value_{value}, is_constexpr_{is_constexpr}, type_{TypeInfo<T>::Type} {}
+      : value_{value}, is_constexpr_{is_constexpr}, type_{TypeInfo<T>::Type} {}
```
lukasm91: I think there's no reason not to use the forwarding constructor here too (also for consistency), at least if you don't plan to replace this anyway in the next PR.

dawn/src/dawn/SIR/SIR.h
```diff
@@ -330,28 +330,33 @@ struct Value : NonCopyable {
     return valueJson;
   }
 
-  template <typename T> struct TypeInfo {};
+  template <typename T>
+  struct TypeInfo {};
```
lukasm91: I think there's no need for the definition here (it just obscures the error message slightly)

jdahm: It needs to be forward declared before it can be specialized outside the class. I made a larger change in the latest commit.

jdahm:
launch jenkins

jdahm:
It looks like the failure is due to a skipped test, not an actual error.

dawn/src/dawn/SIR/SIR.h
lukasm91: Yes, but only declared, not defined
```
template <typename>
struct TypeInfo;
```
should do it I think.

lukasm91:
Reopened one of the comments above  (two characters ... ;))

dawn/src/dawn/SIR/SIR.h
```diff
@@ -272,28 +274,36 @@ struct Stencil : public dawn::NonCopyable {
...
+  bool is_constexpr_;
+  TypeKind type_;
+
+public:
   template <class T>
-  explicit Value(T&& value) : isConstexpr_(false) {
-    valueImpl_ = std::make_unique<ValueImpl<std::decay_t<T>>>(std::forward<T>(value));
-  }
+  explicit Value(T&& value)
+      : value_{std::forward<T>(value)}, is_constexpr_{false}, type_{TypeInfo<T>::Type} {}
```
havogt:
```suggestion
      : value_{std::forward<T>(value)}, is_constexpr_{false}, type_{TypeInfo<std::decay_t<T>>::Type} {}
```

havogt: will fix the test failure in gtclang

dawn/src/dawn/SIR/SIR.h
```diff
@@ -272,28 +274,36 @@ struct Stencil : public dawn::NonCopyable {
...
 
   template <class T>
-  explicit Value(T value, bool isConstexpr) : isConstexpr_(isConstexpr) {
-    valueImpl_ = std::make_unique<ValueImpl<std::decay_t<T>>>(std::forward<T>(value));
-  }
+  explicit Value(T value, bool is_constexpr)
+      : value_{std::forward<T>(value)}, is_constexpr_{is_constexpr}, type_{TypeInfo<T>::Type} {}
```
havogt:
```suggestion
      : value_{std::forward<T>(value)}, is_constexpr_{is_constexpr}, type_{TypeInfo<std::decay_t<T>>::Type} {}
```

dawn/src/dawn/SIR/SIR.h
jdahm: Ah, ok I see what you mean!

dawn/src/dawn/SIR/SIR.h
jdahm: Thanks 👍

jdahm:
launch jenkins

twicki:
launch jenkins

havogt:
There was still a problem here, I will relaunch jenkins and you need to check the issue.

havogt:
launch jenkins

jdahm:
launch jenkins

Issue #238, opened 2019-08-16
Refactor dependencies


havogt:
![DAA8D4D2-D495-46DA-ABE1-F6BB2BF3F478](https://user-images.githubusercontent.com/5136069/63158020-a6b2e200-c018-11e9-9239-91c37c4aef20.jpeg)

twicki:
This was done in a series of PR's listed here:
https://github.com/MeteoSwiss-APN/dawn/pull/264
https://github.com/MeteoSwiss-APN/dawn/pull/245
https://github.com/MeteoSwiss-APN/dawn/pull/244
https://github.com/MeteoSwiss-APN/dawn/pull/240

PR #313, opened 2019-10-08
Add check_modularization script

lukasm91:
## Technical Description

Adds the `check_modularization.sh`-script from GridTools. Complains if we have unexpected dependencies between the modules defined in #238


havogt:
I'd go for the githook as we already have it...

lukasm91:
done. I also added some commented out dependencies that would be ideal to have, but are not there. Can you check the dependencies too?

havogt:
Did you test the git hook? I tried to add a dependency in AST to IIR and didn't catch it.

havogt:
Sorry. Typo on my side.

PR #315, opened 2019-10-09
Add ReduceOverNeighbor concept to SIR

lukasm91:
- Add python constructors for concept
- Add concept to ProtoBuf
- Remove `ASTSimplifier` (this should have been removed in an earlier PR because it is not used; IIRBuilder is an enhanced replacement)
- `ReductionOverNeighbor` now properly supports `getChildren`.
- Add example in python


dawn/examples/python/data/unstructured_stencil_ref.cpp
```diff
@@ -0,0 +1,44 @@
+namespace dawn_generated{
+namespace cxxnaiveico{
+
+class unstructured_stencil {
+private:
+
+  struct stencil_13 {
+    Mesh const& m_mesh;
+    Field<double>& m_in;
```
twicki: there is a change coming in this API very soon - fields need to be specialized. Do you think it is worth it to have a reference here (even though it might be changing quite frequently in the coming weeks)

dawn/src/dawn/AST/ASTExpr.h
```diff
@@ -566,9 +566,10 @@ class LiteralAccessExpr : public Expr {
 /// @ingroup sir
 class ReductionOverNeighborExpr : public Expr {
 private:
+  enum OperandKind { Rhs = 0, Init };
+
   std::string op_ = "+";
-  std::shared_ptr<Expr> rhs_;
-  std::shared_ptr<Expr> init_;
+  std::array<std::shared_ptr<Expr>, 2> operands_;
```
twicki: not that I dislike it, I'm just curious why the change?

dawn/src/dawn/Optimizer/AccessComputation.cpp
```diff
@@ -317,8 +317,8 @@ class AccessMapper : public iir::ASTVisitor {
     removeLastChildAccesses();
   }
   virtual void visit(const std::shared_ptr<iir::ReductionOverNeighborExpr>& expr) override {
-    expr->getRhs()->accept(*this);
-    expr->getInit()->accept(*this);
+    for(auto& s : expr->getChildren())
```
twicki: we usually also put `{}` around single statement for loops. Worth for consistency?

dawn/src/dawn/Optimizer/StatementMapper.cpp
```diff
@@ -185,7 +185,10 @@ void StatementMapper::visit(const std::shared_ptr<iir::UnaryOperator>& expr) {
 }
 
 void StatementMapper::visit(const std::shared_ptr<iir::ReductionOverNeighborExpr>& expr) {
-  DAWN_ASSERT_MSG(0, "ReductionOverNeighborExpr not allowed in this context");
+  DAWN_ASSERT(initializedWithBlockStmt_);
+
+  for(auto& s : expr->getChildren())
```
twicki: we usually also put `{}` around single statement for loops. Worth for consistency?

dawn/src/dawn/Serialization/SIRSerializer.cpp
```diff
@@ -466,6 +466,11 @@ static std::shared_ptr<sir::Expr> makeExpr(const dawn::proto::statements::Expr&
     return std::make_shared<sir::LiteralAccessExpr>(
         exprProto.value(), makeBuiltinTypeID(exprProto.type()), makeLocation(exprProto));
   }
+  case dawn::proto::statements::Expr::kReductionOverNeighborExpr: {
+    const auto& exprProto = expressionProto.reduction_over_neighbor_expr();
+    return std::make_shared<sir::ReductionOverNeighborExpr>(
+        exprProto.op(), makeExpr(exprProto.rhs()), makeExpr(exprProto.init()));
+  }
```
twicki: did I miss the serialization part? seems like only deserialization is here. - Which leads me to the question: Do we want to have a tests-case for this also in Serializer and maybe even one in the (dawn)-integration tests? I'd suggest to at least move the latter to a separate task, but maybe the former does makes sense?

dawn/src/dawn/AST/ASTExpr.h
lukasm91: I don't really like it, but it's needed because I implement `getChildren` now and to implement that, I need the array (as it is done for `BinaryOperator`, for example).

dawn/src/dawn/Optimizer/AccessComputation.cpp
lukasm91: Are you sure? Because if you look at this file, exactly these two lines appear many times (without `{}`). I think currently, consistency is not really enforced and I am tempted to use GCC coding convention (just use the code convention that is used around the code that you're modifying) ;-)

dawn/examples/python/data/unstructured_stencil_ref.cpp
lukasm91: I agree. Removed the verification for this example for the moment.

dawn/src/dawn/Optimizer/StatementMapper.cpp
lukasm91: see above

dawn/src/dawn/Serialization/SIRSerializer.cpp
lukasm91: done

dawn/src/dawn/AST/ASTExpr.h
twicki: got it.

dawn/src/dawn/Optimizer/AccessComputation.cpp
twicki: :) Tendency is yes - Most likely it's not enforced but I think it was intended at some point. I can live with either

twicki:
launch jenkins

lukasm91:
launch jenkins

dawn/src/dawn/Optimizer/AccessComputation.cpp
lukasm91: I think, here I tend to skip the braces, because exactly this line is used many times in the same file. If we want to enforce the convention, we should do this in a separate PR and do all together.

PR #317, opened 2019-10-09
Bundle point to GT v1.0.3

havogt:
## Technical Description

Update to GT v1.0.3


havogt:
launch jenkins

havogt:
They are independent as far as I understand.

havogt:
And maybe I shouldn't tell you that jenkins wasn't using v1.0.3 because it was taking the GridTools from somewhere else.
I'd still merge, because clang-gridtools tested that it actually works.

havogt:
launch jenkins

PR #322, opened 2019-10-10
Fixes to compile on MacOS

jdahm:
This PR fixes a few undefined symbol and type definition errors found when compiling and linking on MacOS. It also sets DYLD_LIBRARY_PATH in the protobuf script instead of LD_LIBRARY_PATH when on MacOS.


dawn/src/dawn/Unittest/CMakeLists.txt
```diff
@@ -19,4 +19,5 @@ yoda_add_library(
           IIRBuilder.cpp
           IIRBuilder.h
   ARCHIVE
+  DEPENDS DawnIIRStatic DawnOptimizerStatic
```
jdahm: I am not sure this is the correct way to define the dependencies in the yoda framework, but Unittest needs to resolve symbols from DawnIIR and DawnOptimizer.

havogt: @cosunae

jdahm:
Both dawn and gtclang (with GTCLANG_ENABLE_GRIDTOOLS=OFF) now build.

gtclang/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ if(GTCLANG_HAS_CUDA)
 endif(GTCLANG_HAS_CUDA)
 
 # Set C++ standard
-set(CMAKE_CXX_STANDARD 17)
+yoda_set_cxx_standard(c++17)
```
havogt: Not 100% sure, but this might break cuda compilation.

jdahm: For some reason, CMake wasn't adding `-std=c++17` when I was compiling with clang-6 with `set(CMAKE_CXX_STANDARD 17)`, but this changed (for some reason) fixed that. Can we tell from the jenkins tests whether that breaks CUDA builds?

twicki: yes, jenkins will test the cuda back end. so if everything passes, you are fine

jdahm:
launch jenkins

gtclang/CMakeLists.txt
havogt: It breaks CUDA compilation because nvcc doesn't support c++17, therefore generated CUDA code must be c++11 compatible (or actually c++14 compatible, but today it is c++11) and `-std=c++17` must not be passed to the compile command. `yoda_set_cxx_standard()` is not really modern CMake and sets it globally for all targets. Adding it only to CMAKE_CXX_STANDARD will not apply it to CUDA files.
I am very surprised that `set(CMAKE_CXX_STANDARD 17)` doesn't work for MacOS. Where is the flag missing? Everywhere or only for some files?

twicki:
launch jenkins

jdahm:
launch jenkins

gtclang/CMakeLists.txt
jdahm: Ah yes, I remember now that CUDA currently only supports c++14. I'll try this again on mac soon and report back. Perhaps this was a symptom of another problem I already fixed elsewhere.

jdahm: This is not necessary after all.

jdahm:
Current status: dawn compiles and builds after `brew install cmake protobuf llvm@6`, and gtclang builds without gridtools support.

We need to add a few more references to DYLD_LIBRARY_PATH in yoda. I'll open a PR to do that soon.

After that, I know so far we need to fix detection of OpenMP on Mac (in `detect_features.cmake:42`) and a reference to
```
$ c++filt __ZN6google8protobuf8internal24proto3_preserve_unknown_E
google::protobuf::internal::proto3_preserve_unknown_
```
@twicki you mentioned you knew about the protobuf issue. Do you know the route to resolve it?

The OpenMP fix could be as easy as adding some logic specifically for MacOS+Homebrew to the `detect_openmp` macro.

jdahm:
@cosunae when you have a chance, could you look at the dependencies I added?

twicki:
preserve unknown is not an issue I've ever stumbled across, sorry!

jdahm:
launch jenkins

Issue #324, opened 2019-10-11
unstructured_stencil.py test failing


Stagno:
Test failing with the following error:
```
Assertion failed: `getIsUnstructuredFromAcessID(AccessID)' 
Function: 'dawn::ast::Expr::LocationType dawn::iir::StencilMetaInformation::getLocationTypeFromAccessID(int) const'
Location: /home/giacomo/dawn/dawn/src/dawn/IIR/StencilMetaInformation.cpp:434

```

lukasm91:
Introduced in  https://github.com/MeteoSwiss-APN/dawn/commit/d7f6a2e3a72149c236322cce792c8378c282af2e. I am wondering, why a commit with "[WIP]" is merged into master?

lukasm91:
@twicki @mroethlin

PR #325, opened 2019-10-11
Check return value in PostInstall

lukasm91:
## Technical Description

`PostInstall.cmake` should also check the return value of the process that runs the python script.

Relates #324


lukasm91:
launch jenkins

lukasm91:
ideally, jenkins should fail now.

PR #327, opened 2019-10-11
Support for clang 9

havogt:
## Technical Description

Support for clang 9.
`skipImplicitNodes()` is moved from Stmt down to Expr.

### DIFF
--- a/gtclang/src/gtclang/Support/ClangCompat/ImplicitNodes.h
+++ b/gtclang/src/gtclang/Support/ClangCompat/ImplicitNodes.h
@@ -1,0 +1,54 @@
+//===--------------------------------------------------------------------------------*- C++ -*-===//
+//                         _       _
+//                        | |     | |
+//                    __ _| |_ ___| | __ _ _ __   __ _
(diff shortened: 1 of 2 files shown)


havogt:
launch jenkins

gtclang/src/gtclang/Support/ClangCompat/ImplicitNodes.h
```diff
@@ -0,0 +1,54 @@
...
+// to reflect the change in clang 9
+
+namespace gtclang {
+namespace clang_compat {
+
+#if CLANG_VERSION_MAJOR < 9
+template <typename StmtT>
+typename std::enable_if<std::is_base_of<clang::Stmt, typename std::decay<StmtT>::type>::value,
+                        StmtT*>::type
+skipAllImplicitNodes(StmtT* e) {
```
lukasm91: when you modify this, can you also remove the SFINAE and replace it with a `static_assert`?

havogt: Makes sense.
I'll do it in a different branch, which I already have.

PR #337, opened 2019-10-14
Fix race condition in CMake between IIR and SIR generated protobuf

### LINKED ISSUE #323: Race condition in CMake
There seems to be a missing dependency from protobuf generator to the consuming compilation units which prevents us from building with more than 6 parallel jobs.

havogt:
## Technical Description

Sets a dependency between the SIR protobuf files and the IIR protobuf files which depend on the former. (Actually not for generating them, but for using the IIR header.)

### Resolves / Enhances

Fixes #323.


havogt:
launch jenkins

PR #353, opened 2019-10-16
Really fix the CMake race condition on protobuf

### LINKED ISSUE #323: Race condition in CMake

havogt:
## Technical Description

Looks like this fixes #323 (replacement for #337).


havogt:
launch jenkins

PR #372, opened 2019-10-21
Don't specify minor version for python

havogt:
## Technical Description

Removes the minor version from searching for the python executable.

### Resolves / Enhances

Resolves nothing as the build process is completely broken anyway. Enhances: finding a suitable python binary on daint. The problem was, with the fixed minor version we matched exactly the system python which was then preferred over the one from modules environment (which was a higher version). The system python was missing a package...


havogt:
launch jenkins

havogt:
launch jenkins

PR #373, opened 2019-10-21
Fix assert in unstructured IIR Builder

lukasm91:
## Technical Description

The location type is already added in `addField`. Doing this twice leads to an assert.


lukasm91:
launch jenkins

PR #374, opened 2019-10-21
Represent domain size as array of integers

### LINKED ISSUE #248: Change domain size argument to multiple ints instead of a string
Currently were parsing a <X>,<Y>,<Z> string which is obscure. We should have individual arguments for this to avoid this "arbitrary" format. Would mean no more parsing it in cuda-code gen, changing the options and modifying the tests / dycore examples using it

jdahm:
Replaces the string stored in `CudaCodeGenOptions` with an array of integer sizes. 
Fixes #248.

This PR depends on https://github.com/MeteoSwiss-APN/clang-gridtools/pull/133 to recover performance


havogt:
I'd make 3 arguments already in the frontend.

twicki:
I think we're already bloating the number of arguments quite a bit and it will never happen that you only want to specify just one of them. But i'm open. Why would you opt for three arguments?

havogt:
To avoid string parsing completely.

jdahm:
launch jenkins

jdahm:
This is ready for review again.

dawn/src/dawn/CodeGen/Cuda/CudaCodeGen.h
```diff
@@ -18,12 +18,9 @@
 #include "dawn/CodeGen/CodeGen.h"
 #include "dawn/CodeGen/CodeGenProperties.h"
 #include "dawn/CodeGen/Cuda/CacheProperties.h"
-#include "dawn/IIR/Interval.h"
```
jdahm: Not used anywhere here, as far as I can tell.

twicki:
launch jenkins

twicki:
This PR will create a performance hit since we're not updating https://github.com/MeteoSwiss-APN/clang-gridtools/blob/01c802cbc2a11bd8299dc3a807fa5cb55d60a0b8/benchmarks/CMakeLists.txt#L65 properly. I'd create a separate PR there and link the two somehow

PR #376, opened 2019-10-22
Default constructible Offsets and Extents

### LINKED ISSUE #369: Merge behaves differently for Extent and Extents with offsets
`Extent::merge(int offset)` checks 0:  https://github.com/MeteoSwiss-APN/dawn/blob/a8ded67b4a6f8b96e4d45051d31be0b206052896/dawn/src/dawn/IIR/Extents.h#L65

while as

`Extents::merge(Offset)` does not: 
https://github.com/MeteoSwiss-APN/dawn/blob/a8ded67b4a6f8b96e4d45051d31be0b206052896/dawn/src/dawn/IIR/Extents.cpp#L59

Beside the fact I doubt that this is used correctly, this inconsistency looks very dangerous to me and is very likely to cause bugs.

A similar behaviour happens with `add`, but this will be deleted because unused.

lukasm91:
## Technical Description

Extents and offsets are now default constructible. The default constructor creates a null offset/extent that can be used with and transformed into an unstructured and/or cartesian offset/extent. This is very helpful for generic passes that need to be able to create offsets/extents, but are not aware of the grid type.

- Adds default constructor for offset
- Adds default constructor for extent
- Adds tests for Offset class
- Improve tests for Extent class

resolves #369

### Examples

```
Extents my_extent{};
Extents cartesian_extent{ast::cartesian};
auto added_extent = my_extent + cartesian_extent;
auto const& hExtent = extent_cast<CartesianExtent const&>(added_extent.horizontalExtent());
// throws bad_cast when unstructured extents are implemented:
// auto const& hExtent = extent_cast<UnstructuredExtent const&>(added_extent.horizontalExtent());
```

Zero extents can be converted to any extent kind:
```
Extents my_extent{};
auto const& hExtent = extent_cast<CartesianExtent const&>(my_extent.horizontalExtent());
```

### Dependencies

No dependencies

### DIFF
--- a/dawn/src/dawn/AST/Offsets.h
+++ b/dawn/src/dawn/AST/Offsets.h
@@ -117,7 +117,11 @@ private:
 
 template <typename T>
 T offset_cast(HorizontalOffset const& offset) {
-  static std::remove_reference_t<T> nullOffset{};
+  using PlainT = std::remove_reference_t<T>;
+  static_assert(std::is_base_of_v<HorizontalOffsetImpl, PlainT>,
+                "Can only be casted to a valid horizontal offset implementation");
+  static_assert(std::is_const_v<PlainT>, "Can only be casted to const");
+  static PlainT nullOffset{};
   return offset.impl_ ? dynamic_cast<T>(*offset.impl_) : nullOffset;
 }
 
--- a/dawn/src/dawn/AST/Offsets.h
+++ b/dawn/src/dawn/AST/Offsets.h
@@ -71,25 +71,42 @@ private:
 
 class HorizontalOffset {
 public:
+  HorizontalOffset() = default;
+
   explicit HorizontalOffset(cartesian_) : impl_(std::make_unique<CartesianOffset>()) {}
   HorizontalOffset(cartesian_, int iOffset, int jOffset)
       : impl_(std::make_unique<CartesianOffset>(iOffset, jOffset)) {}
   HorizontalOffset(HorizontalOffset const& other) { *this = other; }
   HorizontalOffset& operator=(HorizontalOffset const& other) {
-    impl_ = other.impl_->clone();
+    if(other.impl_)
+      impl_ = other.impl_->clone();
+    else
+      impl_ = nullptr;
     return *this;
   }
   HorizontalOffset(HorizontalOffset&& other) = default;
   HorizontalOffset& operator=(HorizontalOffset&& other) = default;
 
-  bool operator==(HorizontalOffset const& other) const { return *impl_ == *other.impl_; }
+  bool operator==(HorizontalOffset const& other) const {
+    if(impl_ && other.impl_)
+      return *impl_ == *other.impl_;
+    else if(impl_)
+      return isZero();
+    else if(other.impl_)
+      return other.isZero();
+    else
+      return true;
+  }
   bool operator!=(HorizontalOffset const& other) const { return !(*this == other); }
 
   HorizontalOffset& operator+=(HorizontalOffset const& other) {
-    *impl_ += *other.impl_;
+    if(impl_ && other.impl_)
+      *impl_ += *other.impl_;
+    else if(other.impl_)
+      *this = other;
     return *this;
   }
-  bool isZero() const { return impl_->isZero(); }
+  bool isZero() const { return !impl_ || impl_->isZero(); }
 
   template <typename T>
   friend T offset_cast(HorizontalOffset const& offset);
@@ -100,11 +117,14 @@ private:
 
 template <typename T>
 T offset_cast(HorizontalOffset const& offset) {
-  return dynamic_cast<T>(*offset.impl_);
+  static std::remove_reference_t<T> nullOffset{};
+  return offset.impl_ ? dynamic_cast<T>(*offset.impl_) : nullOffset;
 }
 
 class Offsets {
 public:
+  Offsets() = default;
+
   Offsets(cartesian_, int i, int j, int k)
       : horizontalOffset_(cartesian, i, j), verticalOffset_(k) {}
   Offsets(cartesian_, std::array<int, 3> const& structuredOffsets)
@@ -125,6 +145,7 @@ public:
     return *this;
   }
 
+  // corresponds to the default initialized offset
   bool isZero() const { return verticalOffset_ == 0 && horizontalOffset_.isZero(); }
 
 private:
--- a/dawn/src/dawn/IIR/Extents.cpp
+++ b/dawn/src/dawn/IIR/Extents.cpp
@@ -67,9 +67,9 @@ bool Extents::isVerticalPointwise() const { return verticalExtent_.isPointwise()
 bool Extents::hasVerticalCenter() const {
   return verticalExtent_.minus() <= 0 && verticalExtent_.plus() >= 0;
 }
-Extents Extents::limit(int minus, int plus) const {
-  DAWN_ASSERT(minus <= 0 && plus >= 0);
-  return Extents{horizontalExtent_.limit(minus, plus), verticalExtent_.limit(minus, plus)};
+Extents Extents::limit(Extents const& other) const {
+  return Extents{horizontalExtent_.limit(other.horizontalExtent()),
+                 verticalExtent_.limit(other.verticalExtent())};
 }
 
 bool Extents::isPointwise() const {
--- a/dawn/src/dawn/IIR/Extents.cpp
+++ b/dawn/src/dawn/IIR/Extents.cpp
@@ -24,17 +24,13 @@ namespace iir {
 
 Extent operator+(Extent lhs, Extent const& rhs) { return lhs += rhs; }
 
+Extents::Extents() : Extents(HorizontalExtent{}, Extent{}) {}
 Extents::Extents(HorizontalExtent const& hExtent, Extent const& vExtent)
     : verticalExtent_(vExtent), horizontalExtent_(hExtent) {}
 
 Extents::Extents(const ast::Offsets& offset)
     : verticalExtent_(offset.verticalOffset(), offset.verticalOffset()),
-      horizontalExtent_(ast::cartesian) {
-  auto const& hOffset = ast::offset_cast<ast::CartesianOffset const&>(offset.horizontalOffset());
-
-  horizontalExtent_ = HorizontalExtent(ast::cartesian, hOffset.offsetI(), hOffset.offsetI(),
-                                       hOffset.offsetJ(), hOffset.offsetJ());
-}
+      horizontalExtent_(offset.horizontalOffset()) {}
 
 Extents::Extents(ast::cartesian_, int extent1minus, int extent1plus, int extent2minus,
                  int extent2plus, int extent3minus, int extent3plus)
@@ -72,6 +68,7 @@ bool Extents::hasVerticalCenter() const {
   return verticalExtent_.minus() <= 0 && verticalExtent_.plus() >= 0;
 }
 Extents Extents::limit(int minus, int plus) const {
+  DAWN_ASSERT(minus <= 0 && plus >= 0);
   return Extents{horizontalExtent_.limit(minus, plus), verticalExtent_.limit(minus, plus)};
 }
 
--- a/dawn/src/dawn/IIR/Extents.cpp
+++ b/dawn/src/dawn/IIR/Extents.cpp
@@ -23,6 +23,14 @@ namespace dawn {
 namespace iir {
 
 Extent operator+(Extent lhs, Extent const& rhs) { return lhs += rhs; }
+Extent merge(Extent lhs, Extent const& rhs) {
+  lhs.merge(rhs);
+  return lhs;
+}
+Extent limit(Extent lhs, Extent const& rhs) {
+  lhs.limit(rhs);
+  return rhs;
+}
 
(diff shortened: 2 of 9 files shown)


lukasm91:
launch jenkins

lukasm91:
launch jenkins

lukasm91:
launch jenkins

dawn/src/dawn/IIR/Extents.cpp
```diff
@@ -24,17 +24,13 @@ namespace iir {
 
 Extent operator+(Extent lhs, Extent const& rhs) { return lhs += rhs; }
 
+Extents::Extents() : Extents(HorizontalExtent{}, Extent{}) {}
 Extents::Extents(HorizontalExtent const& hExtent, Extent const& vExtent)
     : verticalExtent_(vExtent), horizontalExtent_(hExtent) {}
 
 Extents::Extents(const ast::Offsets& offset)
     : verticalExtent_(offset.verticalOffset(), offset.verticalOffset()),
```
Stagno: It looks to me that when you want to initialize a vertical extent you usually want it to correspond to an offset. Therefore it could make sense to add a constructor to `Extent` such as this:
```
Extent(int offset) : minus_(offset), plus_(offset) {}
```
and use it here

dawn/src/dawn/AST/Offsets.h
```diff
@@ -100,11 +117,14 @@ class HorizontalOffset {
 
 template <typename T>
 T offset_cast(HorizontalOffset const& offset) {
-  return dynamic_cast<T>(*offset.impl_);
+  static std::remove_reference_t<T> nullOffset{};
```
Stagno: I see a problem here: when you have no `impl_` and the user asked for a reference such as `T=CartesianOffset&`, this function will return a reference to a static object (`nullOffset`), which is not what we want, as the user will likely modify it.

Stagno: A solution would be to detect whether `T` is a reference and do something like :
```
// if T is a reference {
   if(!offset.impl_) {
      offset.impl_ = std::make_unique<std::remove_reference_t<T>>();
      return dynamic_cast<T>(*offset.impl_);
   }
// }
```

dawn/test/unit-test/dawn/IIR/TestExtent.cpp
```diff
@@ -38,6 +38,21 @@ TEST(ExtentsTest, Construction) {
...
+  EXPECT_EQ(Extents(), Extents());
+
+  Extents nullExtent;
+  Extents extents1{ast::cartesian, -1, 1, -1, 1, -1, 1};
+  extents1 = nullExtent;
+  EXPECT_EQ(extents1, nullExtent);
+
+  Extents extents2{ast::cartesian, -1, 1, -1, 1, -1, 1};
+  extents2 = std::move(nullExtent);
+  EXPECT_EQ(extents2, nullExtent);
```
Stagno: I'm not sure if this is what you want, as `nullExtent` has been `std::move`d

dawn/src/dawn/AST/Offsets.h
```diff
@@ -71,25 +71,42 @@ class CartesianOffset : public HorizontalOffsetImpl {
...
+      impl_ = other.impl_->clone();
+    else
+      impl_ = nullptr;
     return *this;
   }
   HorizontalOffset(HorizontalOffset&& other) = default;
   HorizontalOffset& operator=(HorizontalOffset&& other) = default;
 
-  bool operator==(HorizontalOffset const& other) const { return *impl_ == *other.impl_; }
+  bool operator==(HorizontalOffset const& other) const {
```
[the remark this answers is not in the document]
lukasm91: Are you sure that you would expect that? I feel like comparing a cartesian and an unstructured offset is always an error, because we cannot and we will never be able to handle that case. Just returning false, would mean that they are somewhat compatible, but they are not, in my opinion.

dawn/src/dawn/AST/Offsets.h
```diff
@@ -100,11 +117,14 @@ class HorizontalOffset {
 
 template <typename T>
 T offset_cast(HorizontalOffset const& offset) {
-  return dynamic_cast<T>(*offset.impl_);
```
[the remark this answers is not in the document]
lukasm91: good point

dawn/src/dawn/IIR/Extents.h
```diff
@@ -85,27 +85,27 @@ class HorizontalExtentImpl {
...
   }
-  std::unique_ptr<HorizontalExtentImpl> clone() const { return clone_impl(); }
+  std::unique_ptr<HorizontalExtentImpl> clone() const { return cloneImpl(); }
 
-  void merge(HorizontalExtentImpl const& other) { merge_impl(other); }
-  void merge(ast::HorizontalOffset const& other) { merge_impl(other); }
-  bool equals(HorizontalExtentImpl const& other) const { return equals_impl(other); }
-  bool isPointwise() const { return isPointwise_impl(); }
+  void merge(HorizontalExtentImpl const& other) { mergeImpl(other); }
+  void addCenter() { addCenterImpl(); }
```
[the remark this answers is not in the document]
lukasm91: The problem is in the implementation of `merge`. You can do:
```
HorizontalExtent my_extent{cartesian, 1, 2, 3, 4, 5, 6};
HorizontalExtent zero_extent{};
my_extent->merge(zero_extent);
```
This adds zero to `my_extent`. For `my_extent`, `impl_` is a `CartesianExtent`, for `zero_extent`, `impl_` is `nullptr`.

`merge` now needs to merge those two extents. `HorizontalExtentImpl` has `HorizontalExtentImpl::merge(HorizontalExtentImpl const& other);`, but this I cannot call because the RHS does not exist.

Another option would be to have `HorizontalExtentImpl::merge(HorizontalExtentImpl const*)`, but then we would just duplicate the logic into the `merge` functions of `CartesianExtent` and `UnstructuredExtent`.

dawn/src/dawn/IIR/Extents.h
```diff
@@ -164,48 +160,79 @@ class CartesianExtent : public HorizontalExtentImpl {
 
 class HorizontalExtent {
 public:
+  // the default constructed horizontal extents creates a null-extent that can be compared to all
+  // kind of grids
+  HorizontalExtent() = default;
+
+  HorizontalExtent(ast::HorizontalOffset const& offset) {
```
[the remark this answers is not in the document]
lukasm91: Yes. This is a switch in the version that comes with `UnstructuredOffset`. https://github.com/MeteoSwiss-APN/dawn/blob/d9291379882933cca6e3530ec8f9855b6aa1882d/dawn/src/dawn/IIR/Extents.h#L211

dawn/src/dawn/IIR/Extents.cpp
lukasm91: I see your point. My point against it is that I would not be 100% sure what this constructor does, because it could be either
```
Extent(int offset) : minus_(offset), plus_(offset) {}
```

or

```
Extent(int offset) : minus_(-offset), plus_(offset) {}
```

And both version kind of make sense to me. What do you think?

dawn/src/dawn/AST/Offsets.h
lukasm91: Mhh he shouldn't be able to call this with `T=CartesianOffset&`. The idea is that the user cannot modify the offset through `offset_cast` (that's ok because this function is only needed for CodeGen. Your version doesn't work because you cannot return a valid reference to that object; it will go out of scope immediately.  I think currently it would be possible to cast to non-const (because the pointer is non-const) even though the signature suggests that it's not possible; but I add a static_assert.

dawn/test/unit-test/dawn/IIR/TestExtent.cpp
lukasm91: Good point! Thanks.

dawn/src/dawn/IIR/Extents.h
```diff
@@ -116,39 +116,35 @@ class CartesianExtent : public HorizontalExtentImpl {
...
                                              extents_[1].minus(), extents_[1].plus());
   }
 
-  bool isPointwise_impl() const override {
+  bool isPointwiseImpl() const override {
     return extents_[0].isPointwise() && extents_[1].isPointwise();
   }
 
-  std::unique_ptr<HorizontalExtentImpl> limit_impl(int minus, int plus) const override {
+  std::unique_ptr<HorizontalExtentImpl> limitImpl(int minus, int plus) const override {
```
lukasm91: A final change I am going to do  here is changing `limit` to 
```
Extents::limit(Extents const& other);
```
because `minus` and `plus` don't make sense for unstructured, while when RHS is `Extents`, it makes much more sense.

dawn/src/dawn/IIR/Extents.h
```diff
@@ -164,48 +160,79 @@ class CartesianExtent : public HorizontalExtentImpl {
...
-    return *this;
-  }
-
   template <typename T>
   friend T extent_cast(HorizontalExtent const&);
-  template <typename T>
-  friend T extent_cast(HorizontalExtent&&);
 
-  bool operator==(HorizontalExtent const& other) const { return impl_->equals(*other.impl_); }
+  bool operator==(HorizontalExtent const& other) const {
```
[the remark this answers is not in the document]
lukasm91: see above

dawn/src/dawn/IIR/Extents.cpp
lukasm91: discussed and done.

dawn/src/dawn/AST/Offsets.h
Stagno: What would go out of scope? If you `make_unique` your implementation and then return a reference to it, this will persist, because you hold it as a member.
But if we don't need it, it's ok to disallow `T` being a reference... in case, instead, we realize we could need it, let's just keep in mind that it is still possible

dawn/src/dawn/IIR/Extents.h
Stagno: I agree. Would it make sense if `other` was an rvalue instead? It seems to me that you would never reuse your limit. Btw, I think there's only one use case for `limit()` (in cuda codegen).

dawn/src/dawn/IIR/Extents.h
```diff
@@ -40,6 +40,7 @@ class Extent {
   /// @name Constructors and Assignment
   /// @{
   Extent(int minus, int plus) : minus_(minus), plus_(plus) {}
+  Extent(int extent) : Extent(extent, extent) {}
```
Stagno: Is the correct terminology for the parameter here `extent` or `offset`? I would say offset, as it reflects this https://github.com/MeteoSwiss-APN/dawn/blob/25bc24b899dc8eb4436a455af772b0924a6df3c0/dawn/src/dawn/IIR/Extents.h#L219

dawn/src/dawn/AST/Offsets.h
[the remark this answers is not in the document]
Stagno: Mixing them should crash. What I meant was slightly unrelated: sometimes it may not be clear from the context whether we are dealing with an unstructured/cartesian stencil, therefore we would "ask" `HorizontalOffset` which is its type and run different code accordingly.

dawn/src/dawn/IIR/Extents.h
lukasm91: No, binding an rvalue to `const&` comes without a cost, but this allows us to also call it with an lvalue.
Regarding the other point: Yes there is only one use-case. I think it is a valid addition to the interface as it is the opposite of merge. But there is another caveat:
```
void Extent::merge(Extent const& other);
Extent Extent::limit(Extent const& other);
```

I propose symmetry:
```
void Extent::merge(Extent const& other);
void Extent::limit(Extent const& other);
Extent merge(Extent const&, Extent const&);
Extent limit(Extent const&, Extent const&);
```

dawn/src/dawn/AST/Offsets.h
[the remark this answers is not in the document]
Stagno: To wrap up our discussion around this: we should wait if there's an actual usecase for this "getType" on an offset, but probably there shouldn't be any.

dawn/src/dawn/IIR/Extents.h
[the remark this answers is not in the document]
Stagno: Ok.

lukasm91:
launch jenkins

lukasm91:
launch jenkins

lukasm91:
launch jenkins

lukasm91:
launch jenkins

dawn/src/dawn/IIR/Extents.h
```diff
@@ -223,8 +222,11 @@ class HorizontalExtent {
   }
   void merge(ast::HorizontalOffset const& other) { merge(HorizontalExtent{other}); }
   bool isPointwise() const { return !impl_ || impl_->isPointwise(); }
-  HorizontalExtent limit(HorizontalExtent const& other) const {
-    return impl_ ? impl_->limit(*other.impl_) : *this;
+  void limit(HorizontalExtent const& other) {
+    if(impl_ && other.impl_)
+      impl_->limit(*other.impl_);
+    else if(other.impl_)
+      *this = other;
```
Stagno: Why did the behavior change here? Is this correct? This line will not limit but rather enlarge.

lukasm91:
launch jenkins

PR #344, opened 2019-10-15
Support for clang 10 aka master

havogt:
## Technical Description

Fixes some API changes in clang 10 (master branch as of Oct 14 2019).

PR #380, opened 2019-10-24
Fix compilation for GTClang with LLVM 9

lukasm91:
## Technical Description

Fix compilation with clang 9. I can still not use gtclang built with llvm 9, but at least it compiles.

### Resolves / Enhances

Regression from #344

### DIFF
--- a/gtclang/src/gtclang/Support/ClangCompat/CompilerInvocation.h
+++ b/gtclang/src/gtclang/Support/ClangCompat/CompilerInvocation.h
@@ -22,7 +22,7 @@
 #include "llvm/Option/Option.h"
 
 namespace gtclang::clang_compat::CompilerInvocation {
-#if CLANG_VERSION_MAJOR < 9
+#if CLANG_VERSION_MAJOR <= 9
 inline bool CreateFromArgs(clang::CompilerInvocation& Res, llvm::opt::ArgStringList& ccArgs,
                            clang::DiagnosticsEngine& Diags) {
   return clang::CompilerInvocation::CreateFromArgs(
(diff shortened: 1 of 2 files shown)


lukasm91:
launch jenkins

lukasm91:
@havogt Are you sure?
https://github.com/llvm-mirror/clang/blob/release_90/include/clang/Frontend/CompilerInvocation.h#L154

havogt:
No ;)
Probably I added this while being on master...

gtclang/src/gtclang/Support/ClangCompat/CompilerInvocation.h
```diff
@@ -22,7 +22,7 @@
 #include "llvm/Option/Option.h"
 
 namespace gtclang::clang_compat::CompilerInvocation {
-#if CLANG_VERSION_MAJOR < 9
+#if CLANG_VERSION_MAJOR <= 9
```
havogt: Can you change to < 10? I think that's the pattern I usually used.

lukasm91:
```suggestion
#if CLANG_VERSION_MAJOR < 10
```

gtclang/src/gtclang/Support/ClangCompat/FileUtil.h
```diff
@@ -21,7 +21,7 @@
 #include "clang/Basic/Version.h"
 
 namespace gtclang::clang_compat::FileUtil {
-#if CLANG_VERSION_MAJOR < 9
+#if CLANG_VERSION_MAJOR <= 9
```
lukasm91:
```suggestion
#if CLANG_VERSION_MAJOR < 10
```

gtclang/src/gtclang/Support/ClangCompat/CompilerInvocation.h
lukasm91: you're right

lukasm91:
launch jenkins

PR #381, opened 2019-10-24
Optimizer gt104

mroethlin:
## Technical Description

Turns out that the functionality to mark stencils as "required for correct functionality" is already there. Passes may register themselves as required for correctness in their constructors by setting debug = true. Unfortunate name but I guess this refers to a "Debug" build. However, the not all stencils required for correctness were setting the flag correctly. This is now corrected.

### Resolves / Enhances

GT-104


mroethlin:
launch jenkins

PR #403, opened 2019-10-28
Substitute cloning with std::move

Stagno:
## Technical Description

Removes unnecessary cloning of IIR nodes from the implementations of 2 passes (`PassIntervalPartitioner` and `PassStencilSplitter`) substituting with `std::move`.

### Dependencies
Currently `PassIntervalPartitioner` is not tested, therefore there's no way to check whether the new implementation is correct.


twicki:
closed for now as priority is not high

PR #342, opened 2019-10-14
Getting started

mroethlin:
## Technical Description

This Pull Request introduces a new `README` that serves as a tutorial for people who are new to **dawn** / **GTClang**. Besides simply starting from a **GTClang** two other options are discussed:
1) start from in memory SIR generated in python, then call the C interface from python to generate the code
2) deserialize bytecode SIR from disk in a C++ program, then use the C interface again to generate code

PR #406, opened 2019-10-28
Getting started

jdahm:
## Technical Description

This PR adds a tutorial for dawn and gtclang. It compiles and executes the same stencil three times: once starting from a stencil written with the **GTClang** DSL, once starting by using Python to write SIR, and once handing over SIR to **Dawn** using C++.

The files are found in `gtclang/examples/tutorials`.

### Resolves / Enhances

Extends #342. 

### Example

A set of three examples are being created. See above.

### Testing

Tests are not changed.

### Dependencies

It should not rely on merging the dawn and gtclang CMake systems, but it will be simplified once that is merged.


dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
...
+## Generate code from SIR using dawn from C++
+
+As a final exercise, the C interface to dawn is again used to compile the same example. This time, however, the interface is called from a C++ file. This example will use the SIR written to disk by the preceding example, so please make sure that you followed along beforehand. Switch to the cpp example and build the `dawn_standalone` binary:
+
+```
+cd cpp
+mkdir build
+cd build
+cmake .. && make
+./dawn_standalone
```
mroethlin: should be `./dawn_standalone ../laplacian_stencil_from_python.sir`

dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
...
+globals {
+  double dx;  //grid spacing
+};
+
+stencil laplacian_stencil {
+  storage_ij out_field;
+  storage_ij in_field;
+  Do() {
+    vertical_region(k_start, k_end) {
+	    out_field[i,j] = (-4*in_field + in_field[i+1] + in_field[i-1] + in_field[j-1] + in_field[j+1])/(dx*dx);
```
twicki: if we're using `[i,j]` in outfield we should probably be consistent and use it everywhere so also write `[i,j-1]` etc

twicki: and if we do that, maybe add a sentence in the end saying there is a simplification

dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
...
+mkdir build && cd build && cmake .. && make
+```
+
+This will place an executable called `laplacian_driver` in the tutorial directory. When run, two `vtk` files will be written. Those can be viewed using (ParaView)[https://www.paraview.org/]. `in.vtk` shows the initial conditions. If `out.vtk` is loaded on top, the inversion of phase and twicefold increase in amplitude can clearly be seen, as well as the halos around the domain, which would overlap with a "neighboring" MPI rank in practical implementations.
+
+<img src="img/in.png" width="425"/> <img src="img/out.png" width="425"/> 
+
+## Use Python to generate SIR 
+
+Another option to use **dawn** without having to rely on the **GTClang** DSL is to use the Python interface provided. As a preliminary step, a config file that was generated during the **dawn** install step needs to be copied and the proper `PYTHONPATH` needs to be set. Assuming you built dawn using the bundle this is:
```
twicki: does this not also require dawn to be built in a specific way? I thought `DAWN_BUNDLE_PYTHON` was required?

dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+```
+cd <path/to/dawn>/dawn/examples/
+cp ../bundle/install/examples/python/config.py .
```
twicki: instead of copying the config file into the src directory we could also go to the install/examples directory to run it there - as it was intended.

It is up for discussion if this should be in build/examples or if install is good enough, but that is for another PR

dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
...
+#include "laplacian_stencil_from_python.cpp"
+```
+
+then re-compile and re-run the driver
+
+```
+make && ./laplacian-driver
+```
+
+The python file should be quite easy to follow. The the bulk of the AST of the stencil is generated in function `create_vertical_region_stmt`.
```
twicki: it's bad practice to write easy / hard in tutorials since it makes the user feel bad if they don't progress fast enough on "easy" tasks.
I understand that you don't want to elaborate on it but I'd run it down quickly. Say that the vertical region provides the equivalent info as the gtclang file (and leave it up to the reader to figure it out). It might be helpful to provide the info that the functions are using the functionality provided by the builder in \<insert path\> and then quickly elaborate that lines 99+ are basically setting up options and are here to run the code

dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+```
+make && ./laplacian-driver
+```
+
+The python file should be quite easy to follow. The the bulk of the AST of the stencil is generated in function `create_vertical_region_stmt`. 
+
+## Generate code from SIR using dawn from C++
+
+As a final exercise, the C interface to dawn is again used to compile the same example. This time, however, the interface is called from a C++ file. This example will use the SIR written to disk by the preceding example, so please make sure that you followed along beforehand. Switch to the cpp example and build the `dawn_standalone` binary:
```
twicki: Good practice would be to try to explain what you want to teach in each example. This was "trivial" for the previous two but might not be here, Maybe it would help to elaborate that if you're coming from your front end and you write SIR (through the means of protobuf which is provied \<insert path\>) then you can use the back end in the way that we teach here

dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
+# Getting Started using GTClang & dawn
+
```
twicki: Start a tutorial by explaining what the user will learn at each stage

dawn/examples/tutorial/README.md
```diff
@@ -0,0 +1,106 @@
...
+In this tutorial the basic usage of **GTClang** will be demonstrated using a simple example. To follow this tutorial, please make sure that you compiled **GTClang** with the `GTCLANG_ENABLE_GRIDTOOLS=ON` flag. The readme in the gtclang subdirectory has instructions on how to do that. We will compile and execute the same stencil three times: once starting from a stencil written with the **GTClang** DSL, once starting by using Python to write SIR, and once handing over SIR to dawn using C++.
```
twicki: as written later, I think `DAWN_BUNDLE_PYTHON` is also required?

dawn/examples/tutorial/cpp/introDawnStandalone.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <dawn-c/Options.h>
+#include <dawn-c/TranslationUnit.h>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <stdio.h>
+
+int main(int argc, char* argv[]) {
+  if(argc != 2) {
+    std::cerr << "Usage: " << argv[0] << " <file>" << std::endl;
```
twicki: file is not very descriptive - I'd replace it by at least `SIR file`

dawn/examples/tutorial/laplacian_driver.cpp
```diff
@@ -0,0 +1,102 @@
...
+  auto in_v = make_host_view(in);
+  for(int i = 0; i < N; i++) {
+    for(int j = 0; j < N; j++) {
+      auto pos = at(positions, i, j);
+      in_v(i, j, 0) = sin(std::get<0>(pos)) * sin(std::get<1>(pos));
+    }
+  }
+
+  // visualize in field
+  to_VTK("in.vtk", positions, in, N);
```
twicki: do we want to set the halo to zero before we write the input since it might be confusing to the user if it is nonzero before the computation and zero afterwards?

jdahm:
If we have a single "in-source" CMake for the tutorial, then there should be a `make clean-all` target that removes the CMake files as well.

dawn/examples/tutorial/README.md
jdahm: Good idea. I added that.

dawn/examples/tutorial/README.md
jdahm: This is the only point I did not address in the clean up. At some point, we should discuss installing the tutorial.

twicki: yeah, I'm fine with that

dawn/examples/tutorial/README.md
```diff
@@ -16,12 +16,14 @@ stencil laplacian_stencil {
   storage_ij in_field;
   Do() {
     vertical_region(k_start, k_end) {
-	    out_field[i,j] = (-4*in_field + in_field[i+1] + in_field[i-1] + in_field[j-1] + in_field[j+1])/(dx*dx);
+	    out_field[i,j] = (-4*in_field + in_field[i+1,j] + in_field[i-1,j] + in_field[i,j-1] + in_field[i,j+1])/(dx*dx);
```
twicki: ... `(-4*in_field[i,j]` ...

PR #407, opened 2019-10-28
Merge cmake projects

jdahm:
## Technical Description

Merges the two separate CMake projects `dawn/{dawn,gtclang}` into a single top-level `project()`.


gtclang/CMakeLists.txt
```diff
@@ -89,8 +45,8 @@ make_config_string("Asserts" ${GTCLANG_ASSERTS})
 yoda_report_result("Configuration summary" ${config_info})
 
 # Include the packages (and set the correct, libraries, includes etc.)
-foreach(package bash ccache Python3 Boost clang-format Clang Dawn OpenMP Threads)
```
jdahm: Are Boost and OpenMP direct dependencies here, or are these packages that GridTools depends on?

havogt:
I agree we should refactor CMake, but I propose we discuss the direction first.

dawn/CMakeLists.txt
```diff
@@ -12,51 +12,11 @@
...
-if(NOT BUILD_SHARED_LIBS)
-  set(BUILD_SHARED_LIBS ON CACHE BOOL "Build shared libraries." FORCE)
-endif()
-
-if(NOT CMAKE_INSTALL_PREFIX)
-  set(CMAKE_INSTALL_PREFIX "${CMAKE_SOURCE_DIR}/install" CACHE STRING
-      "Install path prefix, prepended onto install directories." FORCE)
-endif()
-
-project(dawn C CXX)
```
havogt: I believe they should still behave as independent projects. Therefore the preamble should be kept.

dawn/CMakeLists.txt
```diff
@@ -132,10 +77,8 @@ if(DAWN_PYTHON)
 endif()
 
 # Build dawn
-include_directories(${CMAKE_SOURCE_DIR}/src)
-include_directories(${CMAKE_BINARY_DIR}/src)
-include_directories(SYSTEM ${CMAKE_SOURCE_DIR}/test/utils/googletest/include)
-include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/src/dawn/Support/External)
+include_directories(${DAWN_SOURCE_DIR}/src)
```
havogt: If you do as suggested above, then this should be
```suggestion
include_directories(${PROJECT_SOURCE_DIR}/src)
```
and there is no need for `DAWN_SOURCE_DIR`, `GTCLANG_SOURCE_DIR` etc.

twicki:
seems like this came to a halt

PR #426, opened 2019-10-30
Fix dangling reference

havogt:
## Technical Description

In [StencilFunctionAsBCGenerator](https://github.com/MeteoSwiss-APN/dawn/blob/462107cf58910e1734e20a671214452c6ad5c13a/dawn/src/dawn/CodeGen/StencilFunctionAsBCGenerator.cpp#L51) we took a reference from the member of a temporary. For some compilers this resulted in garbage values in the generated code.

Now we return a reference instead of a value.


havogt:
launch jenkins

PR #440, opened 2019-10-31
extended atlas interface to include all possible reductions

mroethlin:
## Technical Description

Quick PR that extends the ATLAS interface to include the reductions which were missing. 

### Testing

Currently not tested. To be tested along with the naive atlas codegen


dawn/prototype/atlas_interface.hpp
```diff
@@ -107,6 +107,91 @@ std::vector<int> const& edgeNeighboursOfCell(atlas::Mesh const& m, int const& id
   return neighs[idx];
 }
 
+std::vector<int> const& nodeNeighboursOfCell(atlas::Mesh const& m, int const& idx) {
+  // note this is only a workaround and does only work as long as we have only one mesh
+  static std::map<int, std::vector<int>> neighs;
+  if(neighs.count(idx) == 0) {
+    const auto& conn = m.cells().node_connectivity();
+    neighs[idx] = std::vector<int>{};
```
twicki: this is the same for every function but `nodeNeighborOfNode`. Can we abstract it to accept a connector instead of duplicating it?

PR #447, opened 2019-11-04
add reference update scripts

lukasm91:
## Technical Description

Adds update scripts for all generated references in dawn and gtclang. 

All scripts try to guess the proper paths from yoda and normal build, otherwise, the options can be provided.

### Testing

Try to change some references and then run the update scripts.

### DIFF
--- a/dawn/test/unit-test/dawn/Optimizer/Passes/update_references.sh
+++ b/dawn/test/unit-test/dawn/Optimizer/Passes/update_references.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+usage()
+{
+    echo "usage: update_references.sh [-g path_to_gtclang] [-h]"
+}
+
+while [ "$1" != "" ]; do
+    case $1 in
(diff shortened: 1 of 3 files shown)


dawn/test/unit-test/dawn/Optimizer/Passes/update_references.sh
```diff
@@ -33,6 +33,10 @@ if [[ "$gtclang_path" == "" ]]; then
         echo "Cannot find gtclang."
         exit 1
     fi
+else
+    if [[ ! -f "$gtclang_path" ]]; then
```
twicki: how can we ever reach this?

lukasm91: if we are given an invalid path

twicki: :+1:

lukasm91:
launch jenkins

PR #451, opened 2019-11-05
Move IIRBuilder and AccessComputation to IIR directory

jdahm:
## Technical Description

Moves `Unittest/IIRBuilder.{cpp,h}` to `IIR/IIRBuilder.{cpp,h}` so that it can be used elsewhere in the code. To keep a consistent dependency tree, `Optimizer/AccessComputation.{cpp,h}` had to be moved as well.

### Resolves / Enhances

 This allows the `IIRBuilder` to be used elsewhere.- I am planning to use it in the `PassFixVersionedInputFields` #441.


jdahm:
Some people think `IIR` may not be the correct place for this class. Let's hold off on merging this until we can talk about it again as a group.

twicki:
I can see the IIRBuilder living in the IIR

Stagno:
I also agree that they fit very well in `IIR`. I think that `IIRBuilder` could even be used to simplify some parts of the code that need to construct IIR trees with stencils/multistages/stages/... Regarding `AccessComputation`, it is reasonable that it should go in the most upstream component possible and that is `IIR`.

jdahm:
launch jenkins

PR #452, opened 2019-11-05
Add unstructured integration tests and required CMake infrastructure

### LINKED ISSUE #435: Make test infrastructure ready for atlas on jenkins
As preliminary step for having regression tests for unstructured, we should prepare our cmake/scripts for the link to atlas. This should work both locally and on jenkins.
As reference, unstructured prototype currently does:
```
g++ -o out atlas[TestCase]Driver.cpp -std=c++17 -I../../gtclang/src -I<path/to/atlas>/include -I<path/to/eckit>/include -L</path/to/atlas>/lib -latlas -L<path/to/eckit>/lib -leckit
```
DoD: have a test in dawn running atlas

### LINKED ISSUE #436: Unstructured regression tests with atlas
Add regression tests for simple unstructured cases that cover all the way from SIR (or maybe just from IIR?) to execution with atlas. Tests should compare outputs against outputs computed by a reference code that uses our atlas interface.

Stagno:
## Technical Description

- Adds infrastructure to 
   1. Code generate the unstructured stencils that are going to be tested
   2. Compile drivers + stencils
   3. Run them and compare outputs with either hardcoded values or with outputs of reference stencils (which also need to be compiled and run)
- Adds `atlas_interface.hpp` (from prototype) into `src/interface`
- Adds `AtlasVerifyer` class to compare outputs (atlas' `ArrayView`s) which should be equal within some tolerance.
- Add tests from code already present in prototype: `CopyCell`, `CopyEdge`, `accumulateEdgeToCell`, `diffusion`

### Resolves / Enhances

resolves #435 resolves #436


dawn/prototype/atlasDiffusionDriverIntegrationTest.cpp
```diff
@@ -0,0 +1,91 @@
+#include <fstream>
```
twicki: remove prototype

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
```diff
@@ -0,0 +1,196 @@
...
+  size_t nb_levels = 1;
+
+  atlas::functionspace::CellColumns fs_cells(mesh, atlas::option::levels(nb_levels));
+  atlas::Field in{fs_cells.createField<double>(atlas::option::name("in"))};
+  atlas::Field out{fs_cells.createField<double>(atlas::option::name("out"))};
+
+  atlasInterface::Field<double> in_v = atlas::array::make_view<double, 2>(in);
+  atlasInterface::Field<double> out_v = atlas::array::make_view<double, 2>(out);
+
+  for(int cell_idx = 0; cell_idx < mesh.cells().size(); ++cell_idx)
```
twicki: should we, for consistency sake initialize out to -1?

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
```diff
@@ -0,0 +1,196 @@
...
+//                   \__, |\__\___|_|\__,_|_| |_|\__, | - GridTools Clang DSL
+//                    __/ |                       __/ |
+//                   |___/                       |___/
+//
+//
+//  This file is distributed under the MIT License (MIT).
+//  See LICENSE.txt for details.
+//
+//===------------------------------------------------------------------------------------------===//
+#include <fstream>
```
twicki: is this needed?

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
```diff
@@ -0,0 +1,196 @@
...
+    using std::swap;
+    swap(in_ref, out_ref);
+    swap(in_gen, out_gen);
+  }
+
+  {
+    auto out_v_ref = atlas::array::make_view<double, 2>(out_ref);
+    auto out_v_gen = atlas::array::make_view<double, 2>(out_gen);
+    AtlasVerifyer v;
+    EXPECT_TRUE(v.compareArrayView(out_v_gen, out_v_ref)) << "while comparing output (on cells)";
```
twicki: why do we use the verifier here and have "hard" coded asserts in the other two tests?

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
```diff
@@ -0,0 +1,196 @@
...
+    in_v(cell_idx, 0) = 1.0;
+
+  dawn_generated::cxxnaiveico::copyCell<atlasInterface::atlasTag>(mesh, static_cast<int>(nb_levels),
+                                                                  in_v, out_v)
+      .run();
+
+  for(int cell_idx = 0; cell_idx < mesh.cells().size(); ++cell_idx)
+    ASSERT_EQ(out_v(cell_idx, 0), 1.0);
+}
+} // namespace
```
twicki: why is each test in it's own private namespace?

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
```diff
@@ -0,0 +1,196 @@
+//===--------------------------------------------------------------------------------*- C++ -*-===//
```
twicki: there is generally no documentation on what's going on in this file - can we have some broad categories? Most likely not everyone will have a good atlas syntax understanding

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
```diff
@@ -0,0 +1,196 @@
...
+//                         _       _
```
twicki: why is this file in the test folder and not in `src/dawn/integrationtest`?

gtclang/test/atlas-integration-test/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+##
+##
+##  This file is distributed under the MIT License (MIT). 
+##  See LICENSE.txt for details.
+##
+##===------------------------------------------------------------------------------------------===##
+
+if(eckit_DIR AND atlas_DIR)
+
+    find_package(eckit)
```
twicki: why is this insde the `if(eckit_DIR`?

gtclang/test/atlas-integration-test/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+            message(FATAL_ERROR "Could not find eckit")
+    endif()
+
+    find_package(atlas)
+    if(NOT atlas_FOUND)
+            message(FATAL_ERROR "Could not find atlas")
+    endif()
+
+    # Need to specify here the names of the stencil codes that are going to be generated.
+    set(generated_stencil_codes generated_copyCell.hpp generated_copyEdge.hpp generated_accumulateEdgeToCell.hpp generated_diffusion.hpp)
```
twicki: for now the convention was to have them be named *.cpp even though we include them

gtclang/test/atlas-integration-test/GenerateUnstructuredStencils.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  {
+    using namespace dawn::iir;
+    using LocType = dawn::ast::Expr::LocationType;
+
+    UnstructuredIIRBuilder b;
+    auto in_f = b.field("in_field", LocType::Cells);
+    auto out_f = b.field("out_field", LocType::Cells);
+
+    auto stencil_instantiation = b.build(
```
twicki: why did you chose to test the code-gen only and not have unstructured SIR and run it trough the full toolchain?

gtclang/test/atlas-integration-test/GenerateUnstructuredStencils.cpp
```diff
@@ -0,0 +1,139 @@
...
+  for(auto const& macroDefine : tu->getPPDefines())
+    ss << macroDefine << "\n";
+
+  ss << tu->getGlobals();
+  for(auto const& s : tu->getStencils())
+    ss << s.second;
+  os << ss.str();
+}
+
+int main() {
```
twicki:
```suggestion
int main(int argc, char* argv[]) {
```

gtclang/test/atlas-integration-test/reference/reference_diffusion.hpp
```diff
@@ -0,0 +1,63 @@
+#define GRIDTOOLS_CLANG_GENERATED 1
```
twicki: what was the idea behind this? Would the goal not be a pure atlas diffusion as the reference?

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
Stagno: I don't think we should have a precondition on out, just the postcondition that after the stencil is run its values are 1.0

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
Stagno: nope, good catch

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
Stagno: To not pollute the global namespace. Also need to close it and open again to have the test-specific includes near the test.

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
Stagno: The purpose of `AtlasVerifyer` is to allow non-exact comparison of `ArrayView`s of fields written by generated and reference stencils, while hard coded checks are for simple copies and reductions for which we expect the results to be exact. Surely we could use `AtlasVerifyer` to do also that by setting the precision to 0 and by forging a "reference" field filled with the expected values, but I'd argue that the code would become less readable.

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
Stagno: Yes, documentation would be nice. What do you mean by broad categories? Could you comment with an example?

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
Stagno: I'm not sure i understand. We cannot move to dawn right now because it depends on `gridtools/clang_dsl.hpp` which is in gtclang.

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
twicki: ok

gtclang/test/atlas-integration-test/AtlasIntegrationTestCompareOutput.cpp
twicki: wrong file and there was an atlas dependency

gtclang/test/atlas-integration-test/CMakeLists.txt
Stagno: Will change to checking if `ENABLE_ATLAS_INTEGRATION_TESTS` is set to on

gtclang/test/atlas-integration-test/CMakeLists.txt
Stagno: will change

gtclang/test/atlas-integration-test/GenerateUnstructuredStencils.cpp
Stagno: but we don't need the arguments, unused variable warnings would pop up

gtclang/test/atlas-integration-test/CMakeLists.txt
Stagno: There's a problem: if they are cpp files, the compiler is going to try to also compile them as separate translation units (and it will fail as other includes are missing), because they are listed as sources of build target `AtlasIntegrationTestCompareOutput` in order to make the target dependent on them.

Stagno: @twicki After having discussed this with you, I tried to do the same thing as gtclang `integration-test`'s cmake does which is adding an additional target that would depend on the source file and then make the test target depend on this additional target (seems a bit hacky btw, but I guess cmake comes with some limitations). Unfortunately, you cannot tell `yoda_add_unittest` to depend on another target unless it's a linkable library (`DEPENDS` keyword).
Another thing that I noticed is that probably the `.hpp` extension was assigned to these files (which come from the prototype) because they contain a stencil class which is templated (with respect to the library tag) ... don't know if this affects something here, though.

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

PR #456, opened 2019-11-06
Fix dangling reference

havogt:
## Technical Description

Another dangling reference was introduced. Caught by clang-tidy.

### DIFF
--- a/dawn/src/dawn/IIR/Extents.h
+++ b/dawn/src/dawn/IIR/Extents.h
@@ -209,16 +209,16 @@ public:
   HorizontalExtent() = default;
 
   HorizontalExtent(ast::HorizontalOffset const& hOffset) {
-    *this = offset_dispatch(hOffset,
-                            [](ast::CartesianOffset const& cOffset) {
-                              return HorizontalExtent(ast::cartesian, cOffset.offsetI(),
-                                                      cOffset.offsetI(), cOffset.offsetJ(),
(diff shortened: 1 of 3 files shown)


dawn/src/dawn/IIR/Extents.h
```diff
@@ -209,16 +209,16 @@ class HorizontalExtent {
   HorizontalExtent() = default;
 
   HorizontalExtent(ast::HorizontalOffset const& hOffset) {
```
havogt: No change in this file. Was it properly formatted before?

dawn/src/dawn/Support/ArrayRef.h
```diff
@@ -77,13 +78,17 @@ class ArrayRef {
   constexpr ArrayRef(const T (&Arr)[N]) : data_(Arr), length_(N) {}
 
 /// @brief Construct an ArrayRef from a std::initializer_list
+#if DAWN_GNUC_PREREQ(9, 0, 0)
```
havogt: fixes a warning in older compilers about -Winit-list-lifetime doesn't exist

dawn/src/dawn/IIR/Extents.h
jdahm: That may have been my enum PR. I don't think VSCode was configured correctly to auto-apply `clang-format` on save.

havogt:
launch jenkins

PR #458, opened 2019-11-08
Introduce deref() in mesh library interface

havogt:
## Technical Description

Adds an extra customization point in the library which allows a cleaner solution for the getEdges() hack in mylib.

### DIFF
--- a/dawn/prototype/mylib.hpp
+++ b/dawn/prototype/mylib.hpp
@@ -1,17 +1,9 @@
 #pragma once
 
+#include <algorithm>
 #include <cmath>
 #include <iostream>
 #include <vector>
-
-namespace wstd {
-struct identity {
-  template <class T>
-  constexpr T&& operator()(T&& t) const noexcept {
-    return std::forward<T>(t);
-  }
-};
-} // namespace wstd
 
 namespace mylib {
 class Vertex;
@@ -106,7 +98,7 @@ private:
 
 class Grid {
 public:
-  Grid(int nx, int ny, bool periodic = false)
+  explicit Grid(int nx, int ny, bool periodic = false)
       : faces_(2 * nx * ny), vertices_(periodic ? nx * ny : (nx + 1) * (ny + 1)),
         edges_(periodic ? 3 * nx * ny : 3 * (nx + 1) * (ny + 1)), nx_(nx), ny_(ny) {
     auto edge_at = [&](int i, int j, int c) -> Edge& {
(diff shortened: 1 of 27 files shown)


dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
```diff
@@ -73,11 +73,11 @@ void ASTStencilBody::visit(const std::shared_ptr<iir::ReductionOverNeighborExpr>
     }
   };
   std::string typeString = getLocationTypeString(expr->getRhsLocation());
-  ss_ << std::string(indent_, ' ') << "reduce" + typeString + "ToCell(LibTag{}, m_mesh, t, ";
+  ss_ << std::string(indent_, ' ') << "reduce" + typeString + "ToCell(LibTag{}, m_mesh, loc, ";
   expr->getInit()->accept(*this);
-  ss_ << ", [&](auto& lhs, auto const& t) { return lhs " << expr->getOp() << "= ";
+  ss_ << ", [&](auto& lhs, auto const& red_loc) { return lhs " << expr->getOp() << "= ";
```
havogt: makes the generated code a bit more readable by having a different name inside the lambda (instead of shadowing the form `t` outside).

dawn/prototype/mylib.hpp
```diff
@@ -276,11 +277,18 @@ class Grid {
           v.add_face(face_at(i - 1, j, face_color::downward));
         }
       }
+
+    for(auto& e : edges_) {
```
lukasm91: could it be const?

dawn/prototype/mylib_interface.hpp
```diff
@@ -21,6 +22,16 @@ inline decltype(auto) getCells(mylibTag, mylib::Grid const& m) { return m.faces(
 inline decltype(auto) getEdges(mylibTag, mylib::Grid const& m) { return m.edges(); }
 inline decltype(auto) getVertices(mylibTag, mylib::Grid const& m) { return m.vertices(); }
 
+// TODO Should be moved to a place common to all interfaces
```
lukasm91: would that be `gtclang/src/gridtools/clang/interface.hpp`? I know it's a weird path, because that folder should not be in gtclang, but this is true for the whole folder.

havogt: Probably, but I would introduce this new dependency in the mylib prototype. Maybe we should wait until we sorted out this problem?

dawn/prototype/mylib.hpp
havogt: done

dawn/prototype/mylib_interface.hpp
havogt: Ah, the dependency is already there.

dawn/src/dawn/IIR/ASTStmt.h
```diff
@@ -24,7 +24,7 @@
 namespace dawn {
 namespace iir {
 
-struct Extents;
+class Extents;
```
havogt: I sneaked in this fix...

havogt:
launch jenkins

havogt:
launch jenkins

PR #466, opened 2019-11-12
Temporarily disable IIR Optimizer no-dependency check

Stagno:
## Technical Description

Since `IIRBuilder` has been moved to IIR, the `check_modularization` pre-commit hook blocks us from committing as `IIRBuilder` depends on Optimizer (it uses `restoreIIR()`) and on CodeGen.
This is a temporary fix to allow `IIRBuilder` to stay in IIR until it's moved back to tests (or another decision is taken). It removes the dependency on CodeGen (not needed in any case) and it disables the no-dependency check IIR->Optimizer.

### Dependencies

None


Stagno:
launch jenkins

PR #496, opened 2019-11-15
Fix pre-commit ignore list

Stagno:
## Technical Description

Fix `pre-commit` which was not using the ignore list (regular expressions) defined by `ignore_list.sh`. Also fixes such list which was incorrect as it assumed that the file paths outputted by `git diff-index` were of the form `./dawn/something`, while they actually are like `dawn/something`.


Stagno:
launch jenkins

Issue #409, opened 2019-10-29
Use modern CMake in Dawn and GTClang


cosunae:

jdahm:
We discussed some sub-tasks to make this large effort more digestible. These are:
1. Merge CMake projects together (#407) (blocks 6)
2. Switch to target-based include directories, compilation flags, etc. (depends on 3, blocks 6)
3. Remove the bundle (blocks 6)
4. Remove yoda from main project (depends on 1)
5. Move `gridtools::clang` to top-level directory in project
6. Create a modern CMake bundle (depends on 3)

cosunae:
thanks @jdahm 
it is not clear how to interpret the combination of 3 & 6. We remove the bundle and create a new bundle?
Could you clarify it? 
I would also like to point out that even if gtclang was merged into dawn, and there is less need now for a bundle we still have the clang-gridtools project with all the dycore examples that still makes use of the bundle. So in case of 3) we would need to find an equivalent solution

jdahm:
Hi @cosunae - The idea behind (3) and (6) was, as you said, ensure there's a bundle-like build option left after the refactor. As a start, that bundle could use yoda, even though yoda is planned to be removed from the dawn/gtclang build process. Later we could incorporate CMake's ExternalProject: https://cmake.org/cmake/help/latest/module/ExternalProject.html

PR #497, opened 2019-11-17
Replace dawn bundle by a simple CMake script

jdahm:
## Technical Description

Replaces dawn bundle by a CMake script that downloads Protobuf.

### Resolves / Enhances

Part of #409.


jdahm:
This is put into the other CMake PR.

PR #505, opened 2019-11-28
fix to reordering strategy

twicki:
## Technical Description

This fixes the greedy reorder strategy to do what we expect in the write after write case

###  Enhances

Fixes illegal execution order

### Example

```
stencil Test {
  storage in, out;
  Do {
      out = in;
    }
      out = 0;
    }
  }
};
```


twicki:
launch jenkins

PR #507, opened 2019-11-29
Rename GRIDTOOLS_CLANG_* defines

Stagno:
## Technical Description

- Rename GRIDTOOLS_CLANG_* defines to either GRIDTOOLS_DAWN_* or just DAWN_*
- Simplifies defines regarding precision
- Enforces correct dependencies between includes

### Testing

On clang-gridtools https://github.com/MeteoSwiss-APN/clang-gridtools/pull/144

### Dependencies

https://github.com/MeteoSwiss-APN/clang-gridtools/pull/144


Stagno:
launch jenkins

dawn/src/dawn/CodeGen/CodeGen.cpp
```diff
@@ -346,8 +346,7 @@ void CodeGen::addMplIfdefs(std::vector<std::string>& ppDefines, int mplContainer
 
   ppDefines.push_back(makeIfNotDefined("BOOST_RESULT_OF_USE_TR1", 1));
   ppDefines.push_back(makeIfNotDefined("BOOST_NO_CXX11_DECLTYPE", 1));
-  ppDefines.push_back(
-      makeIfNotDefined("GRIDTOOLS_CLANG_HALO_EXTEND", codeGenOptions.MaxHaloPoints));
+  ppDefines.push_back(makeIfNotDefined("GRIDTOOLS_DAWN_HALO_EXTEND", codeGenOptions.MaxHaloPoints));
```
lukasm91: shouldn't it be `DAWN_GRIDTOOLS`? And can you rename it to `EXTENT` with a `T`?

Stagno: The `GRIDTOOLS_DAWN` prefix is to mirror the namespace `gridtools::dawn`. Having `dawn::gridtools` would have made things much more complicated, as to access objects within `::dawn` we would have needed to prefix them with `::dawn::`.
Will fix the D instead of T typo.

PR #510, opened 2019-12-03
Use `make_grid` in the generated code

lukasm91:
## Technical Description

The constructor of `gridtools::grid` is not part of the public API (indicated by the fact that it changed from v1.0 to v1.1). Thus, dawn should use `gridtools::make_grid` which is public API to create the grid.

### Resolves / Enhances

GT4Py uses Gridtools 1.1.


lukasm91:
launch jenkins

lukasm91:
still some tests failing... ;-)

lukasm91:
launch jenkins

PR #517, opened 2019-12-04
Coding Conventions

mroethlin:
## Technical Description

This is the PR to update our coding conventions. Some points left open before are now further clarified. 

* east vs. west const was left open for now, since there seems to be a `clang-format` rule for this [in the near future](https://reviews.llvm.org/D69764), making the point obsolete.
* I agree that a naming convention for template parameters would be nice to improve legibility. Currently most of the template parameters we use (which aren't simply `T`) start with an uppercase letter, clashing with the convention for class names. Maybe it would be reasonable to have all template parameters start with `T`? This is already respected in parts of the code, e.g. in `dawn/IIR/IIRNode.h` which uses  `template <typename TNodeType>`, `template <typename TParent, typename TChild>` etc.

### DIFF
--- a/_sources/coding.txt
+++ b/_sources/coding.txt
@@ -21,6 +21,8 @@ Naming of Types, Functions, Variables, and Enumerators
 
   * **Members** should start with a lower-case letter and contain an `_` suffix (e.g ``myMember_``) 
 
+  * **Namespaces** are all lower case and are in snake_case (e.g. `dawn_generated`)
+
 For types, functions and enumerators you should follow the LLVM style.
 
 Source Code Formatting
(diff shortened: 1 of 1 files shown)


_sources/coding.txt
```diff
@@ -80,13 +84,65 @@ while the following is considered bad
...
+# Headers in the same library
+# Other libraries
+# `std` headers
+
+For headers in the same folder, do not put the absolute path.
+
+Miscellaneous Guidelines
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+** Do not define member functions which are longer than one line in header files (except for templated functions). E.g.
```
jdahm:
```suggestion
** Do not define non-templated member functions which are longer than one line in header files (except for templated functions). E.g.
```

PR #549, opened 2019-12-20
Gradient test case

### LINKED ISSUE #546: Make weighted reduce test a proper gradient computation
Currently, the test in the gradient computation using the weighted reduce over neighbor concept does not really compute a gradient. Fix the computation using periodic boundaries, either employing the atlas periodic halo facilities or by manually fixing the edge neighbor lists to wrap around the domain

mroethlin:
## Technical Description

Fixed the gradient test case
* Our testing w.r.t atlas was fundamentally broken. We cache the neighbor lists as static local variables in a function. This crashes and burns in the worst possible ways if different mesh sizes are used for the individual tests. For now, I deleted the caching in the atlas interface for testing. The atlas interface in prototype still uses caching. 
* No caching is _very_ slow, I had to reduce the number of iterations in the diffusion example from 500 to 5. We need to find another solution
* I made a Wrapper for Atlas that translates from the lat long stuff to Cartesian coordinates and offers debug facilities like querying the orientation of an edge, 
* Fixed order of operations in the atlas integration testing (function spaces on edges where defined before said edges were constructed, this fails silently and quite horribly)
* Made a function that "wraps" around the neighbor lists of edges to get periodic boundaries. Using the Wrapper described above this seemed simpler than trying to emulate the behavior using Atlas halos 

### Resolves / Enhances

Resolves #546 

### Example / Testin

There is an example of a working gradient in the prototype folder as well as in the atlas-integration tests


mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

dawn/prototype/AtlasCartesianWrapper.cpp
```diff
@@ -0,0 +1,100 @@
...
+//
+//===------------------------------------------------------------------------------------------===//
+
+#include "AtlasCartesianWrapper.h"
+
+Point AtlasToCartesian::cellMidpoint(const atlas::Mesh& mesh, int cellIdx) const {
+  const atlas::mesh::HybridElements::Connectivity& cellNodeConnectivity =
+      mesh.cells().node_connectivity();
+
+  const int missingVal = cellNodeConnectivity.missing_value();
```
Stagno: Sorry for the probably stupid question: what is the missing value here? I was looking through the atlas code and couldn't figure it out ... it's just a curiosity

dawn/prototype/atlasGradientDriver.cpp
```diff
@@ -11,8 +11,53 @@
...
+      // second index
+      if(atlasMapper.edgeOrientation(mesh, edgeIdx) == Orientation::Vertical) {
+        assert(cellI == nx - 1 || cellI == 0);
+        cellI = (cellI == nx - 1) ? 0 : nx - 1;
+      } else { // Orientation::Horizontal
+        assert(cellJ == ny - 1 || cellJ == 0);
+        cellJ = (cellJ == ny - 1) ? 0 : ny - 1;
+      }
+      int oppositeIdx = cellI + cellJ * nx;
+      // ammend the neighbor list
```
Stagno: This won't "merge together" boundary edges which are at the exact opposite side of the grid. I guess just linking to the correct cell neighbors is sufficient for the gradient example, right?

dawn/src/interface/atlas_interface.hpp
```diff
@@ -99,101 +99,65 @@ std::vector<int> getNeighs(const atlas::Mesh::HybridElements::Connectivity& conn
...
-  // note this is only a workaround and does only work as long as we have only one mesh
-  static std::map<int, std::vector<int>> neighs;
-  if(neighs.count(idx) == 0) {
-    neighs[idx] = getNeighs(m.edges().cell_connectivity(), idx);
-  }
-  assert(neighs[idx].size() == 1 || neighs[idx].size() == 2); // boundary edges may have 1 cell
-  return neighs[idx];
+std::vector<int> const cellNeighboursOfEdge(atlas::Mesh const& m, int const& idx) {
+  auto neighs = getNeighs(m.edges().cell_connectivity(), idx);
+  assert(neighs.size() == 2); // boundary edges may have 1 cell
```
Stagno: Should we keep this assert? === Will it always be the case that we have periodic boundaries?

gtclang/test/atlas-integration-test/GenerateUnstructuredStencils.cpp
```diff
@@ -48,10 +49,14 @@ int main() {
...
-                        b.stage(b.vregion(dawn::sir::Interval::Start, dawn::sir::Interval::End,
-                                          b.stmt(b.assignExpr(b.at(out_f), b.at(in_f))))))));
+        "copyCell",
+        b.stencil(
+            b.multistage(LoopOrderKind::Parallel,
+                         b.stage(b.vregion(dawn::sir::Interval::Start, dawn::sir::Interval::End,
+                                           b.stmt(b.assignExpr(b.at(out_f), b.at(in_f)))))),
+            b.multistage(LoopOrderKind::Parallel,
+                         b.stage(b.vregion(dawn::sir::Interval::Start, dawn::sir::Interval::End,
+                                           b.stmt(b.assignExpr(b.at(out_f), b.at(in_f))))))));
```
Stagno: Popped up again :)

gtclang/test/atlas-integration-test/GenerateUnstructuredStencils.cpp
```diff
@@ -15,6 +15,7 @@
 //===------------------------------------------------------------------------------------------===//
 
 #include "dawn/CodeGen/CXXNaive-ico/CXXNaiveCodeGen.h"
+#include "dawn/CodeGen/CXXNaive/CXXNaiveCodeGen.h"
```
Stagno: Popped up again :)

PR #550, opened 2019-12-27
fix pip calls to use the venv

twicki:
## Technical Description

Jenkins used to use the standard pip instead of the one in the virtual environment which led to permission errors

### Resolves

Fixes the currently broken plans on master

### Dependencies

This is independent


twicki:
launch jenkins

twicki:
launch jenkins

Issue #470, opened 2019-11-12
Formalize the parallel model 


twicki:
Currently we have the notion of Stages in our parallel model that do not serve a direct cause since there can be stages without sync (meaning they do not enforce synchronization). The only strong requirement they bring is that there can not be two overlapping do-methods in one stage.

There is also the issue with a offset read after write in a MS to fields that can lead to illegal execution that should be specified. 

We should clarify what exactly is meant by any concept. This will probably not affect the user code since there is no notion of parallelism there but we should be precise about our model

cosunae:
I would say the model is the following, 
computations are group into a stage if they do not require synchronization (as you said) and they require the same extent of redundant computations (i.e. the extent of the Stage). 
Indeed two stages might not require synchronization because the separation was triggered by the second condition. 
The first condition is unavoidable, the second is only for performance reasons, but It can be force by activate the StageMerger pass. 

Does this help?

PR #551, opened 2019-12-27
First draft for a parallel model description

twicki:
## Technical Description

A formal description of the parallel model in dawn on both the SIR and IIR level.
There are still some open points (in the line-separated parts of the text) that need to be discussed as we're converging.

### Resolves

Fixes https://github.com/meteoswiss-apn/dawn/issues/470

### Dependencies

This PR is independent


dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+   index ``k`` will be executed after each statement that it depends on
+   that appears previously in the current vertical region or in any
+   previous vertical region was executed on every vertical level with
+   indices ``k-1`` and smaller.
+-  If the loop order is backward, any statement with a vertical
+   dependency on another statement executed on a vertical level with
+   index ``k`` will be executed after each statement that it depends on
+   that appears previously in the current vertical region or in any
+   previous vertical region was executed on every vertical level with
+   indices ``k+1`` and bigger.
```
lukasm91: Can you rephrase this sentence? I had trouble parsing this sentence... 
1. "after each statement that it depends on": If we say that we can implement the toolchain in any way if the code runs _as if it would obey this rules_, we could just skip this part and say "after each statement that appears".
2. "all previous statements were executed on every vertical level with indices ``k-1`` and smaller". Now think about
```
    vertical_region(k_start, k_end)
      one = zero[k]
    vertical_region(k_start, k_end)
      two = one[k + 1]
```
If I understand correctly from what you write, this would behave the same way as
```
    vertical_region(k_start, k_end)
      one = zero[k]
      two = one[k + 1]
```
But it should not, because it should execute the whole first vertical_region first (but because of the "as-if" rule, the toolchain might do something like
```
    vertical_region(k_start, k_end)
      one = zero[k + 1]
      two = one[k]
``

File: dawn/docs/parallel_model.rst [hunk 2]
```diff
@@ -0,0 +1,251 @@
...
+
+An example of this is the following
+
+.. code:: c++
+
+   1 vertical_region(k_start, k_end)
+   2   one = 1
+   3   two = one[i - 1]
+
+For every vertical level, we guarantee that line two will be executed on
```
lukasm91: Not sure if this is true. You say here that we allow vertical-diagonal dependencies:
```
   1 vertical_region(k_start, k_start)
   2   one = 1
   3 vertical_region(k_start + 1, k_end)
   4   one = 1
   5   two = one[i - 1, k - 1]
```
Is this a real pattern we should support? This are not "pure vertical solvers" anymore.

dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+intermediate representation. It's specification can be found `here`_.
+
+Dependencies
+~~~~~~~~~~~~
+
+Generally, stencil computation in dawn differentiates between horizontal
+and the vertical dimension. The assumption is that individual statements
+of a stencil are embarrassingly parallel in the horizontal. For that
+reason, there is no guarantee that any horizontal point's computation
+will be finished before any other in each user statement.
```
lukasm91: > The assumption is that individual statements of a stencil are embarrassingly parallel in the horizontal.
This implies that you cannot do self-assignment in general (`out = out[i+1]`), but is there anything else?

dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+Here, for every level with index ``k``, line two will be executed before
+line 4.
+
+Horizontal Dependencies
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Irrespective of the loop order every statement with a horizontal
+dependency on other statements will be executed after all statements it
+depends on are executed on the full horizontal plane in the same
+vertical level.
```
lukasm91: Could we skip the part with "with a horizontal dependency on other statements"? Say that each statement will be executed on the full horizontal plane, but it has the same problem anyway (because of the write-after-read-after-write problem).

dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+
+A Stage (St) specifies a specific compute-domain for different
+computations. These can be either global computation bounds or extended
+compute domains due to larger required input data.
+
+Each stage has the option to require (block-wide) synchronization.
+
+-  in the CUDA back end they are equivalent to conditionals that check
+   if the block / thread combination is inside the compute domain and a
+   potential ``__syncthreads()`` afterwards.
```
lukasm91: is there a reason to have it only "potentially"? Because if you don't need synchronization you could put them into the same Do?

dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+-  in the naive back end every storage will be synchronized on the host
+   at the beginning of each multistage
+
+--------------
+
+   Can we discuss why we still have the loop order on this level? Would
+   it potentially be interesting to promote this to the Do level and
+   change the fusion strategy if we use the GridTools back end? Or do we
+   know that we are losing performance in the CUDA back end if we mix
+   forward and backward loops versus new kernel calls?
```
lukasm91: I think I agree that we could put it into the Do-level. Then the only purpose of the multistage would be to resolve the write-after-read-after-write problem?

dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+   Non-temporary fields can cause problem in the following case: write
+   center, read offset, write center. This is currently translated to
+   one multistage with three stages. The problem is that on input /
+   output fields we duplicate the writes which can have 2 side-effects
+   if neighboring blocks are out of sync: Block 1 writes in its extended
+   domain (line 2), block 2 writes (line 2) (same value), then reads
+   offset (line 3), then writes in the center (line 4). Now Block 1
+   reads (line 3) and gets the wrong value. This means that if we want
+   to solve the problem we need a global sync in that case (meaning
+   breaking the kernel call).
```
Stagno: Yes, this looks like the only possible solution. It should mean, if I'm correct, that line 4 must go in a new multistage. So we should fix the current implementation.
If we decide to remove the loop order concept from the multistage, than the multistage would still be useful because of this.

dawn/docs/parallel_model.rst
Stagno: I don't think we risk loosing performance if we remove the loop order from the multistage, or, at least, I couldn't come up with an example for which we would miss possible optimizations with such parallel model. Just having the loop order on the do-methods should do fine. (I assume the various reordering/fusing/splitting algorithms are correct and smart)
An example to motivate is:
```
1 one = 1
2 zero = 0
3 two = one[k - 1]       // forward
4 three = one[k + 1] 	   // backwards
5 one = 2
6 four = zero[k - 1]     // forward
```
here we would need a multistage split between lines 4 and 5. But this shouldn't prevent line 6 to be "put together" with line 3 (loops fused because of same order), as it has no dependency on lines 4 and 5.

dawn/docs/parallel_model.rst
twicki: maybe we should rephrase it then. What it means is there can not be loop carried dependencies (reductions) in the horizontal.

Since we do field-versioning, we should probably explain it here?

dawn/docs/parallel_model.rst
lukasm91: Actually, for the frontend (and the high-level representation) we can easily accept read-write-read, because the problem only happens with temporary fields. As we only describe how the program behaves, we can easily think that each non-temporary field is first copied into a temporary field, and in the end copied out of the temporary field. So we can assume that each field has redundancy because there is a simple starting point from which we can optimize the problem. Do you agree?

dawn/docs/parallel_model.rst
lukasm91: I see. Yes field versioning sounds like a good thing to mention here. So what should happen if I write
```
out = out[i+1] + out[i - 1]
```
If we do field versioning, this will become
```
out' = out[i+1] + out[i - 1]
```
Which looks perfectly legal. So the statement is not about a missing guarantee, but it is a statement that each statement is executed double-buffered
```
out' = out[i+1] + out[i-1]
out = out'
```

dawn/docs/parallel_model.rst
twicki: 1) sounds good

2) we need to define a target here, our current implementation creates the same result (one backward MS with both lines in the same do).

dawn/docs/parallel_model.rst
twicki: I don't follow. If you have two pointwise statements
```
a = b
c = a
```
there is no guarantee that `a=b` is executed on the full plane before `c=a` is executed. Of course this guarantee is there on a point by point basis.

dawn/docs/parallel_model.rst
twicki: That is a pattern we never encountered in all of comso. Unclear if we need to support it

dawn/docs/parallel_model.rst
twicki: yes, currently the reason is that stages cannot have overlapping do-methods. so you might need to generate two stages to fulfil that requirement even though they do not require synchronisation.

dawn/docs/parallel_model.rst
twicki: we could let the do-method hold all the information about the k-loop, right? Then the concept of the MS would just be global sync.

Does that make sense?

dawn/docs/parallel_model.rst
lukasm91: what I mean is: We can tell the user in the high level representation that each statement is executed on the *full plane* one after the other. Internally, we can still do an analysis of what data he is actually accessing and do it on a point-by-point basis internally, but to the user, it still feels like the full plane is updated one after the other, because the outcome is the same.

dawn/docs/parallel_model.rst
lukasm91: Yes, I know. This is why I bring this point (it was already brought up many times before). This simplifies a lot when thinking about the problems and we can fully split between horizontal stencils and vertical solvers.

dawn/docs/parallel_model.rst
lukasm91: Yes. This doesn't change the fact that the multistage is at the point where the k-loop(s) happen, but the actual indices this is going to iterate over is inside the do-method. But I like the idea to have one responsibility per concept (global sync).

dawn/docs/parallel_model.rst
lukasm91: I see. But there is nothing fundamental which prohibits us doing this, right?

dawn/docs/parallel_model.rst
cosunae: I would propose something like
```
Generally, stencil computation in dawn differentiates between horizontal
and the vertical dimension. The assumption is that individual statements
of a stencil are embarrassingly parallel in the horizontal. For that
reason, there is no guarantee that any horizontal point's computation
will be finished before any other in each user statement.
Computations with data dependencies like 
tmp = field[i-1]
field = tmp[j+1] + tmp[j-1]
will be resolved by field versioning and adding synchronizations between statements with data dependencies. 
That means that iterative solvers where a computation accumulates values over one of the horizontal dimensions are not possible
```

dawn/docs/parallel_model.rst
cosunae: I didnt understand Lukas point 2, but description with the point 1 Lukas sounds good.

dawn/docs/parallel_model.rst
twicki: @lukasm91 
the current implementation does a different thing:
we currently have 
```
NewMS:
 out' = out
MS:
  out = out'[i+1] + out'[i-1]
```
so snyc is implied if versioning is required. 

@cosunae 
should we maybe touch on all the ways this can happen?
the`out = out[offset] + something` case is a "different one" that is also interesting, right?

dawn/docs/parallel_model.rst
twicki: got it. Can be done

dawn/docs/parallel_model.rst
twicki: I would say no

dawn/docs/parallel_model.rst
lukasm91: Let me explain again point 2:
The user can have
```
    vertical_region(k_start, k_end)
      one = zero
    vertical_region(k_start, k_end)
      two = one[k + 1]
```
What will be in variable two in the end? It might be `zero[k + 1]` or it might be the original `one[k + 1]`.

I personally might be tempted to think that first, `one = zero[k]` is applied to the whole vertical region, and after that, the `two = one[k + 1]` is applied to the whole vertical region, because we have two vertical regions (and they have a direction, a start, and an end).

But as far as I understand, the text says (with some parentheses for my own understanding ...):

> any [statement with a vertical dependency on another statement executed on a vertical level with index ``k``] will be executed [after each statement that it depends on that appears ({previously in the current vertical region} or {in any previous vertical region}) was executed on every vertical level with indices ``k-1`` and smaller].

So it says that all statements in a previous vertical region are executed on every vertical level with indices `k-1` and smaller. Which is equivalent to write:

```
    vertical_region(k_start, k_end)
      one = zero
      two = one[k + 1]
```
So for each level, both statements are applied, and then we move to the next level.

twicki: Is that the intended behaviour?
If the loop order is forward, any statement executed on a vertical level with index ``k`` will be executed after each statement that appears in the current vertical region on every vertical level with indices ``k-1`` and smaller.
Each previous vertical region was executed on its full iteration space.

dawn/docs/parallel_model.rst
lukasm91: Agreed that the implementation might be different, but it doesn't matter (both interpretations are ok). The important thing is that it is double buffered.
```
out' = out[i+1] + out[i-1]
out = out'
```
and
```
out' = out
out = out'[i+1] + out'[i-1]
```
They should be equivalent (of course statement by statement, applied to the full ij-plane).

dawn/docs/parallel_model.rst
lukasm91: What should happen here:
```
vertical_region(k_start, k_end):
    one = 0
vertical_region(k_start, k_end):
    one = 1
    two = one[k + 1] (1) + one (2) + one[k - 1] (3)
    one = 2
    two = two + one[k + 1] (4) + one (5) + one[k - 1] (6)
```
What is happening in dawn:
```
Stencil_0
    MultiStage_0
        Stage_0
            Do
                one = 0
        Stage_1
            Do
                one = 1
    MultiStage_1
        Stage_0
            Do
                two = one[k + 1] (1) + one (2) + one[k - 1] (3)
    MultiStage_2
        Stage_0
            Do
                one = 2
    MultiStage_3
        Stage_0
            Do
                two = two + one[k + 1] (4) + one (5) + one[k - 1] (6)
```
So if I understand correctly, it means that (1), (2) and (3) refer to `one = 1`, while (4), (5) and (6) refer to `one = 2`.

Is this really what we expect our parallel model to do?

dawn/docs/parallel_model.rst
lukasm91: Just an addition here from the discussion with @twicki : The global sync is needed for the write-after-read-after-write problem, but it is also needed for vertical-diagonal offsets (i+1, k+1), if `ij` and `k` loops may be exchanged. So if we have vertical-diagonal offsets, we need to split into two multistages.

dawn/docs/parallel_model.rst
lukasm91: I agree that we never encounter it. Might it appear indirectly, because we do something like
```
vertical_region(k_start, k_end)
    zero = 0
    one = zero[k - 1]
    two = one[i + 1]
```
I think the consequence is the same: At some point we need to split into two multistages (because two accesses `zero[k - 1, i + 1]`. 

If I am right and this can also happen indirectly, it might be better to support the pattern, because it's too hard for the user to know when he uses this pattern. But it is not yet fully thought through.

dawn/docs/parallel_model.rst
twicki: @cosunae ?

dawn/docs/parallel_model.rst
twicki: true. And in the contract we have with the SIR it should not matter since the user only needs to know the effect (his field is versioned, his input is read properly and the output is in the correct field)

dawn/docs/parallel_model.rst
mbianco: I need some help in understanding the statements (sorry, but I'm slow). Taking Lukas's example:

```
vertical_region(k_start, k_end)
     one = zero
vertical_region(k_start, k_end)
      two = one[k + 1]
```
If the two vertical_regions are just parallel ranges, and if I read "the first vertical statement is executed before the second one" the loop could be transformed into
```
vertical_region(k_end, k_start)
      one = zero[k+1]
```
What's wrong with my reasoning?

dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+reason, there is no guarantee that any horizontal point's computation
+will be finished before any other in each user statement.
+
+Vertical Dependencies
+^^^^^^^^^^^^^^^^^^^^^
+
+Each vertical region of a stencil specifies a loop order of execution.
+Dawn will uphold the following rules:
+
+-  If the loop order is forward, any statement with a vertical
```
mbianco: I guess my main concern here is that there are only two loop orders in the vertical. I think we should keep "parallel" in the IIR level, too.

dawn/docs/parallel_model.rst
```diff
@@ -0,0 +1,251 @@
...
+   if neighboring blocks are out of sync: Block 1 writes in its extended
+   domain (line 2), block 2 writes (line 2) (same value), then reads
+   offset (line 3), then writes in the center (line 4). Now Block 1
+   reads (line 3) and gets the wrong value. This means that if we want
+   to solve the problem we need a global sync in that case (meaning
+   breaking the kernel call).
+
+.. code:: c++
+
+   1 vertical_region(k_start, k_end){
```
mbianco: One should be able to read the code meaning without knowing if the fields are temporary or not. If I read this it is very clear what this means: two will contains `1`, except at the halo.

dawn/docs/parallel_model.rst
twicki: so the goal was to not express parallelism in the front end (and hence the SIR). On an easy level this means 
```
vertical_region(k_start, k_end)
  a = b
vertical_region(k_start, k_end)
  c = d
```
should yield to the same output code as 
```
vertical_region(k_start, k_end)
  a = b
  c = d
```
This can be achieved. 
The question is really what we expect to happen with k-dependencies. Here we can change the behaviour of the parallelizer if we have a clear idea what we promise the user. 
There are two different computations you can express:

```
  vertical_region(k_start, k_end)
    one = zero
  vertical_region(k_start, k_end)
    two = one[k + 1]
````
can mean set all of one to zero and then set all of two to the updated one. The result of this would be `two=zero` and `one = zero`. Where 
```
  vertical_region(k_start, k_end)
    one = zero
    two = one[k + 1]
````
can be interpreted as update my current one and write the old one to two. The result of this would be `one = zero` `two=one`. 

The current implementation of dawn does not distinguish between the two and always goes for the second option. We need to clarify if these should be different things though.

And if they are different (my opinion is they should be since there are two motifs to express) we need to think very carefully about how Lukas' last example should be interpreted. 

I hope this clarifies things for you @mbianco?

File: dawn/docs/parallel_model.rst (hunk 8 above)
Comment (twicki) [reply dawn/docs/parallel_model.rst]: IIR does have parallel. SIR does not since the parallelization happens in the back end where we traverse accesses to clarify if parallel is a viable strategy for the loop order

Comment (twicki) [reply dawn/docs/parallel_model.rst]: and this means that the contract with the user is clearly defined but if our analysis shows that parallel execution can uphold that contract we can switch. 

This again boils down to the fact that parallelism is not expressed in the front end / SIR. The user can choose if the numerics dictate a loop order and specify that one and we will respect it if we need to (to make sure the results are the expected ones)

File: dawn/docs/parallel_model.rst (hunk 9 above)
Comment (twicki) [reply dawn/docs/parallel_model.rst]: so you agree that we need to kernel calls here?

File: dawn/docs/parallel_model.rst (hunk 1 above)
Comment (mbianco) [reply dawn/docs/parallel_model.rst]: Yes, thanks. I think these semantics questions should be answered before we talk about the parallel model.

Comment (Stagno) [reply dawn/docs/parallel_model.rst]: By reading the last code that you wrote:
```
1  vertical_region(k_start, k_end)
2    one = zero
3    two = one[k + 1]
```
I would expect the same thing that I expect when reading horizontal stencil code, that is isolation between statements: line 3 should get executed after line 2 has fully executed, resulting in `two = one = zero`. As comparison, look at the horizontal version of it:
```
1  vertical_region(k_start, k_end)
2    one = zero
3    two = one[i + 1]
```
... since we expect statement isolation we stage-split between 2 and 3.

I don't see the benefit of complicating the intended meaning of `vertical_region`, as you can always obtain the result `one = zero` and `two = one(prev)` by simply writing:
```
1  vertical_region(k_start, k_end)
2    two = one
3    one = zero
```
We gain in simplicity of the analysis and readability of the code if we keep `vertical_region` as just the specifier of the vertical interval in which its statements execute.
What I say we should do to fix the problem of the current implementation, which appears in this case
```
1  vertical_region(k_start, k_end)
2    one = one[k+1]
3    one = one * 2.0
```
(2 and 3 are erroneously put in the same do-method), is to disallow fusion (or introduce a split) when detecting that a statement (line 3) writes on a field accessed vertically offset by a statement that precedes it (line 2). Which is, again, analogous to what we currently do in the horizontal.

Comment (lukasm91) [reply dawn/docs/parallel_model.rst]: So what you basically say is that in a vertical region, you want to forbid accessing the "old values", because this can always be done by versioning manually, right? That would mean, that a field, which is read and then written in the same vertical region, it is always a solver. SIR should not have a notion of forward/backward then. The implied solver directions would need to match, so you cannot have a forward read-write dependency and a backward read-write dependency in the same vertical region. 
```
vertical_region(k_start, k_end):
    one = 1
    one = one[k - 1] + one
```
This will be a solver.

```
vertical_region(k_start, k_end):
    one = 1
    one = one[k - 1] + one[k + 1]
```
This is invalid.

Comment (Stagno) [reply dawn/docs/parallel_model.rst]: @lukasm91
> So what you basically say is that in a vertical region, you want to forbid accessing the "old values", because this can always be done by versioning manually, right? 

Yes, because statement-level atomicity would imply that once line 2 (of my last example) has been fully executed, the old values of `one` are lost and line 3 will operate on the new values. But why would you need the old values in any case? Could you provide an example?

> That would mean, that a field, which is read and then written in the same vertical region, it is always a solver. SIR should not have a notion of forward/backward then. The implied solver directions would need to match, so you cannot have a forward read-write dependency and a backward read-write dependency in the same vertical region.

To me having a vertical offset access on a statement implies that the statement is (part of) a vertical solver. Whether we want to switch the order of `k_start` and `k_end` or not in the vertical region declaration, I don't care, because the direction is always clear from the accesses.
The only ambiguous case is when you don't have neither vertical nor horizontal offset accesses:
```
one = one + 1
```
which could be interpreted as either part of a solver or an horizontal stencil. Operationally it shouldn't matter. But we could we could improve DSL code readability by putting vertical solvers inside a `vertical_solver` block.

> 
> ```
> vertical_region(k_start, k_end):
>     one = 1
>     one = one[k - 1] + one
> ```
> 
> This will be a solver.

Yes, and I would code generate it to:
```
for (k=k_start; k<k_end; k++) {
   one(k) = 1;
}
for (k=k_start; k<k_end; k++) {
   one(k) = one(k-1) + one(k);
}
```

> ```
> vertical_region(k_start, k_end):
>     one = 1
>     one = one[k - 1] + one[k + 1]
> ```
> 
> This is invalid.

This would still be valid and I would code generate it to:

```
for (k=k_start; k<k_end; k++) {
   one(k) = 1;
}
for (k=k_start; k<k_end; k++) {
   one(k) = one(k-1) + one(k+1);
}
```

> 
> ```
> vertical_region(k_start, k_end):
>     one = 1
> vertical_region(k_start, k_end):
>     one = one[k - 1] + one[k + 1]
> ```
> 
> This are two 3D stencils (not a solver)

This, to me, it's the exact same as before (same generated code).

Comment (twicki) [reply dawn/docs/parallel_model.rst]: Vertical offsets are not the same as solvers though. You can also just have normal stencils in the vertical

File: dawn/docs/parallel_model.rst (hunk 7 above)
Comment (Stagno) [reply dawn/docs/parallel_model.rst]: I agree that the frontend shouldn't change, but I don't understand the rest. Could you provide an example, please?

File: dawn/docs/parallel_model.rst (hunk 1 above)
Comment (Stagno) [reply dawn/docs/parallel_model.rst]: > Vertical offsets are not the same as solvers though. You can also just have normal stencils in the vertical

@twicki right. Then I think it is important to fix the syntax both on the frontend and on the backend as soon as possible.
From what I understand, writing `one = one[k - 1]` in the context of vertical solvers means accumulating, but in the context of stencils means a stencil access (accessing the "old values" of `one` in the rhs, like you would access an array). I can see two alternative solutions:

- Always having vertical solvers in a `vertical_solver` block, where the meaning of `one[k-1]` is that of an array access. Outside a vertical solver, of course, we would mean a stencil access.
- Use a different notation: `one{k-1}` for an array-like access and `one[k-1]` for a stencil access.

Comment (cosunae) [reply dawn/docs/parallel_model.rst]: Let me summarize and see if understood the discussion here. There original problem was with this code:
```
vertical_region(k_start, k_end)
     one = zero
vertical_region(k_start, k_end)
      two = one[k + 1]
```
In the current semantics (I am going to name it **Semantic 0**), any access that is counter loop order, means a stencil (i.e. access old data and not solver).
That is `two = one[k+1]` should be a stencil. The problem is that currently dawn is merging the two vertical regions into 
```
vertical_region(k_start, k_end)
     one = zero
      two = one[k + 1]
```
which clearly shouldnt do since the semantics of the stencil access is to access values written in previous vertical_regions to `one`. But if we fix the merging pass, adding the rule that a solver stmt can not be moved into a vertical_region where we write into the same field, we would fix the problem, and the behaviour would be semantically correct, isnt? 

Then I understood Lukas/Giacomo were proposing to simplify the semantics (**Semantic 1**)by not using the loop order to identify whether accesses are solvers or stencils. Instead we assume that any read access to a field that has been written into in the same vertical_region is a solver. Then the loop order will be defined by the read access of the solver. 
In `out = out[k-1]` should be forward, and the opposite for `out = out[k+1]`. 
Using this semantic, my interpretation diverges though from the examples of Lukas and Giacomo, so here it comes mine:
```
vertical_region(k_start, k_end):
    one = 1
    one = one[k - 1] + one
```
I agree with both here, this is a solver. Though is a strange example since stmt `one=1` is not used in most of the region. I think the following is more readable for me (though it does not change the result)
```
vertical_region(k_start, k_start)::
    one = 1
vertical_region(k_start+1, k_end)::
    one = one[k - 1] + one
```
The example:
```
vertical_region(k_start, k_end):
    one = 1
    one = one[k - 1] + one[k + 1]
```
I agree with @lukasm91, this should be invalid. In principle is a solver but we can not define the loop order of the solver since there are two contradictory accesses. I guess @Stagno was making valid here assuming **Semantic 0** ?

For the example: 
```
vertical_region(k_start, k_end):
    one = 1
vertical_region(k_start, k_end):
    one = one[k - 1] + one[k + 1]
```
I disagree with both, to me this is the way I would actually write a solver, because of the stmt
```    one = one[k - 1] + one[k + 1] ```. Notice that the pre-condition `one = 1` can always go into a separate `vertical_region(k_start,k_start)` as a defined in the example above and is not what defines the solver but the self -update.

Then to conclude we have the possibility of **Semantic 2**, using a `vertical_solver`. I would like to add the caveat that this probably makes difficult to combine solvers accesses (for example to different fields) even in the same statement, which could happen, isnt?
Finally we have the **Semantic 3** of using different notation `{}` an `[]` (i.e. different field access nodes).

Did I get it right? Is there now about choosing or do we still lack some understanding or non covered cases?

File: dawn/docs/parallel_model.rst [hunk 10]
```diff
@@ -0,0 +1,251 @@
...
+An example of this would be the following
+
+.. code:: c++
+
+   1 vertical_region(k_start, k_end)
+   2   one = 1
+   3 vertical_region(k_start, k_end)
+   4   two = one[k - 1]
+
+Here, for every level with index ``k``, line two will be executed before
```
cosunae: Following the discussions above, what is ambiguous in the text I think is that both:
```
vertical_region(k_start, k_end)
   one = 1
vertical_region(k_start, k_end)
   two = one[k - 1]
```
and 
```
vertical_region(k_start, k_end)
   one = 1
   two = one[k - 1]
```
are compatible with the sentence `line two will be executed before`. Do we need to introduce the concept of a solver and add
`in case of a solver access, all writes into that field of previous vertical_regions are guaranteed to have been processed for the entire k domain before computing the solver.` ?

dawn/docs/parallel_model.rst
cosunae: yes, that is another interesting example. I think this is a real use case. 
What about this in the VerticalAdvectionFunctions ? 
`Average<Direction, DirectionFactor, NestedShift<Z, 1>>(eval, wcon());`
As far as we have staggering, I think is normal to do stencils (averages, etc) in the i,j dimension for a displaced position in k. But I think all these cases are stencils and not solvers. I think solvers should be always accessing a column. It looks weird that a solver is accessing an intermediate value of the diagonals of the solve of a neighbour column, or at least I dont know what that means. Therefore I would say that we need to support this, directly and/or indirectly (which as pointed out by lukas does not make any difference). So I guess adding the requirement that solvers are purely vertical to any of the semantic options would do the job. 
However @lukasm91 added the example that would indirectly violate this condition. If I can also reformulate the example: 
```
var one
vertical_region(k_start, k_start)
    zero = 0
vertical_region(k_start+1, k_end)
    zero = zero[k - 1]
    one = zero
    two = one[i + 1]
```
So what we could do is to add a check (or a pass) that for any vertical region, any horizonal stencil access is not accessing and field that has have a vertical solver access. This is probably not trivial since we need to traverse backwards the DAG of dependencies. 
Additionally, we could have the following: 
```
var one
vertical_region(k_end, k_end)
    zero = 0
vertical_region(k_end-1, k_start+1)
    zero = zero[k + 1]
vertical_region(k_start, k_start)
    zero = zero[k + 1]
    one= zero[i + 1]
```
that means we have a forward (not in the example)/backward solver, and at the end of the solver for level k_start we still have a solver access (`zero=zero[k+1]` ). After that stmt, actually the solution is in place in the fields and we could access `one = zero[i+1]`. When the solver is finished is up to the user, it could happen at a boundary layer (k_end, k_start) or at any other level. 
So 
* option 1: we dont add any protection/check and let the responsibility to the user.
* option 2: we do check, and force the user to close the vertical_region and open a new one before he can access the solutions of a solver. However, that is still complicated, because as output of some optimizations, the example I made before where we access the output of a solver at the last level should be legal. So even if we force the user to split vertical_regions, we can not check this state internally anymore after some optimizations happen.

dawn/docs/parallel_model.rst
cosunae: Let me summarize and see also if I understood here, since I was confused about a couple of statements in the comments. 
@Stagno I think splitting into different multistages will not solve the problem, since multistages are not synchronizing among blocks which is the essential problem here. If splitting is the solution, it should be split into different stencils. 
However I think @lukasm91 gives the best solution, that is field version this into temporaries. A correction on his comment so that make sense to me, "because the problem only happens with **non-**temporary fields". 
That means we should have a mode of the field versioning that checks for this patterns and upon detection put the field into a temporary. Temporaries will have private allocations of redundant halos so that the current organization of a single multistage and multiple stages will work with the parallel model. Finally the temporary needs to be copied back into the field. 
Therefore the action here would be we need to support this in the field versioning pass, but the example is totally valid and we dont need to do anything concerning the parallel model. Correct?

dawn/docs/parallel_model.rst
Stagno: @cosunae 
In my first comment (https://github.com/MeteoSwiss-APN/dawn/pull/551/files#r366916356) I erroneously assumed that all vertical offset accesses are to be intended as solver accesses. This could be still a viable option because if the user wants a stencil-like access he/she can always manually field-version:

> ```
> vertical_region(k_start, k_end):
>    one = 1;
>    one_out = one[k-1] + one[k+1];
>    one = one_out
> ```

but it's definitely not the one I'd prefer.

In the second comment (https://github.com/MeteoSwiss-APN/dawn/pull/551/files#r366935001), as you said we have Semantic 2 and Semantic 3.
Answering your question on semantic 2: it is important that we thoroughly check if that's a possible use case that we need to support (horizontal stencil + solver access in the same statement). If it is, we can't have semantic 2.

dawn/docs/parallel_model.rst
lukasm91: They will do the same, but they are not the same (if you have `one[k + 1]`, they do different things in the formulation I am aiming for). But yes, I agree, that we should probably aim for a better formulation which explicitly specifies solvers; but first we might want to fix the current formulation (whether the full fix should be implemented or not - I don't know).

dawn/docs/parallel_model.rst
lukasm91: > Then I understood Lukas/Giacomo were proposing to simplify the semantics (Semantic 1)by not using the loop order to identify whether accesses are solvers or stencils. Instead we assume that any read access to a field that has been written into in the same vertical_region then is a solver. 
No, I don't think we should do that (my statements were just about trying to understand proposed solutions)! The user always have a direction in mind when he is writing a solver. So we should accept that, and I think there is no reason to reverse engineer what the user might have thought when writing the code.

dawn/docs/parallel_model.rst
lukasm91: Option 3: We just generate the right code. This might lead to very slow codes (for CUDA, a kernel call in each k-iteration), but at least we have legal code, and we can diagnose to the user why his code is small.

PR #531, opened 2019-12-04
Front end implementation for global indexing

twicki:
## Technical Description

Implementation of code-generation for globally indexed stages

### Enhances

Fixes #492

### Dependencies

This is built on top of https://github.com/MeteoSwiss-APN/dawn/pull/530

PR #554, opened 2020-01-02
stages with defined extents do not extend their computational domain

### LINKED ISSUE #472: Ensure that IIR global indexing is working with all passes
in #399 we add the concept to the IIR and we should ensure that these concepts go through all passes without deteriorating performance

twicki:
## Technical Description

Discussions showed that Stages that have a specified horizontal extent should never be extended to fit for additional computation

### Resolves / Enhances

Fixes #472

### Example

```
#include "gtclang_dsl_defs/gtclang_dsl.hpp"
using namespace gtclang::dsl;

stencil Test {
  storage in, out, out2;
  void Do() {
    iteration_space() {
      // computation
      out = in;
    }
    iteration_space(i_start, i_start + 1) {
      // correction term
      out = 3;
    }
      // use of conputed value
      out2 = out[i + 1] + out[i - 1] + out[j - 1] + out[j + 1];
    }
  }
};
```

### Dependencies
This PR is independent, for the test, use the front end branch https://github.com/MeteoSwiss-APN/dawn/pull/531

### DIFF
--- a/dawn/src/dawn/Optimizer/PassComputeStageExtents.cpp
+++ b/dawn/src/dawn/Optimizer/PassComputeStageExtents.cpp
@@ -14,6 +14,7 @@
 
 #include "dawn/Optimizer/PassComputeStageExtents.h"
 #include "dawn/IIR/DependencyGraphStage.h"
+#include "dawn/IIR/Extents.h"
 #include "dawn/IIR/IIRNodeIterator.h"
 #include "dawn/IIR/StencilInstantiation.h"
 #include "dawn/Optimizer/OptimizerContext.h"
@@ -36,6 +37,12 @@ bool PassComputeStageExtents::run(
     // backward loop over stages
     for(int i = numStages - 1; i >= 0; --i) {
       iir::Stage& fromStage = *(stencil.getStage(i));
+      for(auto globalIterationSpace : fromStage.getIterationSpace()) {
+        if(globalIterationSpace) {
+          fromStage.setExtents(iir::Extents());
+          continue;
+        }
+      }
 
       iir::Extents const& stageExtent = fromStage.getExtents();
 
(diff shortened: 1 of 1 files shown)


twicki:
launch jenkins

dawn/src/dawn/Optimizer/PassComputeStageExtents.cpp
```diff
@@ -36,6 +37,12 @@ bool PassComputeStageExtents::run(
     // backward loop over stages
     for(int i = numStages - 1; i >= 0; --i) {
       iir::Stage& fromStage = *(stencil.getStage(i));
+      for(auto globalIterationSpace : fromStage.getIterationSpace()) {
+        if(globalIterationSpace) {
+          fromStage.setExtents(iir::Extents());
+          continue;
```
lukasm91: not sure what this statement is supposed to do, but certainly it does not what it should as it will only skip the rest of the current iteration (which is nothing).

twicki: you are right, fixed

twicki:
launch jenkins

dawn/src/dawn/Optimizer/PassComputeStageExtents.cpp
```diff
@@ -36,6 +37,13 @@ bool PassComputeStageExtents::run(
     // backward loop over stages
     for(int i = numStages - 1; i >= 0; --i) {
       iir::Stage& fromStage = *(stencil.getStage(i));
+      // If the stage has a global interationspace set, we should never extend it since is is user
```
lukasm91:
```suggestion
      // If the stage has a global interationspace set, we should never extend it since it is user
```

lukasm91:
```suggestion
      // If the stage has a global iterationspace set, we should never extend it since it is user
```

twicki:
launch jenkins

PR #555, opened 2020-01-02
Fix compilation warnings

jdahm:
## Technical Description

The new CMake branch removes certain flags that ignored warnings. This fixes most of those warnings. This relies on first merging the new CMake branch.


jdahm:
Retargeting this to the CMake refactor branch

PR #566, opened 2020-01-08
Fix k-Caching in presence of iteration spaces

mroethlin:
## Technical Description

Glocal Iteration spaces were introduced. This has consequences w.r.t to caching. If a stage with an iteration space caches a field, subsequent reads in another stage may "miss" the cache if this other stage has no iteration space or an iteration space outside the first stages'. In these cases, caching needs to be disabled. 

### Testing

There are various simple tests with two stages introduced which check the correct behavior of the caching in presence of iteration spaces.


dawn/src/dawn/IIR/Stage.cpp
```diff
@@ -347,5 +347,26 @@ void Stage::setIterationSpace(const IterationSpace& value) { iterationSpace_ = v
 
 const Stage::IterationSpace& Stage::getIterationSpace() const { return iterationSpace_; }
 
+bool Stage::hasIterationSpace() const {
+  return iterationSpace_[0].has_value() || iterationSpace_[1].has_value();
```
twicki:
```suggestion
return std::any_of(iterationSpace_.cbegin(),iterationSpace_.cend(), [](const auto& p) { return p.has_value(); })
```

dawn/src/dawn/Optimizer/PassSetCaches.cpp
```diff
@@ -227,6 +227,7 @@ bool PassSetCaches::run(const std::shared_ptr<iir::StencilInstantiation>& instan
         const auto& fields = ms.getFields();
         for(const auto& AccessIDFieldPair : fields) {
           const iir::Field& field = AccessIDFieldPair.second;
+          // printf("%s: %d", ms.getMetadata().getNameFromAccessID(field.getAccessID()).c_str(), field.getIntend());
```
twicki: rm

dawn/src/dawn/Unittest/IIRBuilder.cpp
```diff
@@ -183,6 +183,15 @@ IIRBuilder::Field CartesianIIRBuilder::field(std::string const& name, FieldType
   return {id, name};
 }
 
+IIRBuilder::Field CartesianIIRBuilder::tmpField(std::string const& name, FieldType ft) {
```
twicki:
```suggestion
IIRBuilder::Field CartesianIIRBuilder::tmpField(const std::string& name, FieldType ft) {
```

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassCaching.cpp
```diff
@@ -0,0 +1,291 @@
...
+              1); // we expect the instantiation to contain one stencil...
+  ASSERT_TRUE(stencil->getStencils().at(0)->getChildren().size() == 1); //... with one multistage
+
+  // check if the cache was set correctly (we expect a k cache fill on tmp) 
+  const auto& multiStage = *stencil->getStencils().at(0)->getChildren().begin();
+  ASSERT_TRUE(multiStage->isCached(tmp.id));
+  ASSERT_TRUE(multiStage->getCache(tmp.id).getIOPolicy() == iir::Cache::IOPolicy::fill);
+  ASSERT_TRUE(multiStage->getCache(tmp.id).getType() == iir::Cache::CacheType::K);
+
+  //   std::ofstream of("check_kcache.cpp");
```
twicki: rm

dawn/src/dawn/Optimizer/PassSetCaches.cpp
```diff
@@ -248,6 +249,32 @@ bool PassSetCaches::run(const std::shared_ptr<iir::StencilInstantiation>& instan
...
+              break;
+            }
+            stageIter++;
+          }
+
+          //check if subsequent uses are compatible
+          auto firstUsageWithGlobalIndex = ms.getChildren().begin();
+          std::advance(firstUsageWithGlobalIndex, stageIter);
+          bool incompatibleIterationSpaces = false;
+          for (auto it = firstUsageWithGlobalIndex; it != ms.getChildren().end(); it++) {
```
twicki: is this loop over all other stage? It seems like the check that the field in question is used is missing?

mroethlin: loop is over all subsequent stages. that check that field is in use was indeed missing, good catch!

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #583, opened 2020-01-10
Consistency grid

### LINKED ISSUE #423: Consistency check for extents and offsets
We need to check whether all extents and offsets are of the grid type that the user promises us.

We already know which grid type we have, so we need to traverse the AST, check all offsets and make sure that they can be casted to that grid type.

This consistency check should be unit tested!

mroethlin:
## Technical Description

This PR adds consistency type checking for *Grid* types. That is, it is ensured that once the grid type is decided, classes with specializations regarding the grid type can only be constructed using the correct grid type. If an SIR is constructed it sets the global grid type, this grid type is forwarded to the IIR. If an IIR is constructed only (e.g. using the IIR Builder) the IIR sets the global grid type. No other class may set the global grid type. If neither IIR nor SIR is constructed no grid type is set and the checks are not performed (this is the case in some unit tests). 

Constructing a class with incosistent grid type simply asserts, there is no "graceful" failure or forwarding of error messages to the user in this case. 

## Testing

No additional tests were introduced. It is assumed that the tests currently in dawn test this functionality implicitly. 

## Resolves / Enhances

Fixes #423


mroethlin:
launch jenkins

mroethlin:
launch jenkins

dawn/src/dawn/AST/GridType.h
```diff
@@ -1,10 +1,50 @@
...
 #include <iosfwd>
+#include <optional>
 
 namespace dawn {
+namespace iir {
+class IIR;
+}
+} // namespace dawn
+
+namespace dawn {
```
jdahm: Merge into namespace above

PR #584, opened 2020-01-11
Correctly detect externally-installed protobuf python bindings

jdahm:
# Technical Description

Correctly detect externally-installed protobuf python bindings.


jdahm:
launch jenkins

PR #587, opened 2020-01-13
Increase CMake test discovery timeouts

jdahm:
## Technical Description

Increase discovery timeout.


jdahm:
Yes, `GTClangUnittestFrontend` timed out in debug mode on my laptop.

PR #588, opened 2020-01-13
Fix installation of dawn

havogt:
## Technical Description

Fixes installation of dawn
- Add missing headers to install
- Fix driver-includes include path


havogt:
launch jenkins

PR #597, opened 2020-01-15
Put test output in build directory

jdahm:
## Technical Description

Put GTClang integration test output in the build directory.

### DIFF
--- a/dawn/src/dawn/Compiler/DawnCompiler.cpp
+++ b/dawn/src/dawn/Compiler/DawnCompiler.cpp
@@ -88,6 +88,16 @@ struct ComputeEditDistance<std::string> {
   }
 };
 
+std::pair<std::string, std::string> dirnameBasename(const std::string& filename) {
+
+  const char sep = '/';
+
+  const auto i = filename.rfind(sep);
+  return (i != std::string::npos)
+             ? std::make_pair(filename.substr(0, i), filename.substr(i + 1, filename.length() - i))
+             : std::make_pair(filename, filename);
+}
+
 std::string filenameSansExtension(const std::string& fullName, const std::string& extension) {
   const std::size_t pos = fullName.rfind(extension);
   return pos < std::string::npos ? fullName.substr(0, pos) : fullName;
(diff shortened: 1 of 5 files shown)


dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -88,6 +88,16 @@ struct ComputeEditDistance<std::string> {
   }
 };
 
+std::pair<std::string, std::string> dirnameBasename(const std::string& filename) {
+
+  const char sep = '/';
+
+  const auto i = filename.rfind(sep);
```
havogt: I saw a lot of path manipulation in prs recently. We should consider using std::filesystem.

jdahm: Ah, yes I’m glad you said that. There are different parts of the code all using their own parsing methods. I will change that in this PR before we merge.

gtclang/test/utils/gtclang-tester/gtclang_tester/runner.py
```diff
@@ -123,14 +131,19 @@ def run(self):
             # Check expected files
             if test.get_expected_file() and not Config.generate_reference:
                 for files in test.get_expected_file():
-                    for outputfile, referencefile in zip(files.get_output_files(), files.get_reference_files()):
+                    for outputfile, referencefile in zip(
+                        files.get_output_files(), files.get_reference_files()
+                    ):
 
                         try:
+                            print("HERE ", outputfile, referencefile)
```
jdahm: Delete this line before merging!

dawn/src/dawn/Compiler/DawnCompiler.cpp
jdahm: I used it in this place. We'll have to track down the other places later.

jdahm:
I would like to merge this at the same time as #598.

jdahm:
launch jenkins

jdahm:
@havogt This works locally but is breaking on Travis. Can you attempt a quick fix tomorrow? It looks like it's missing the c++ filesystem library in the link step. From a quick search online I found out that this may be due to a difference in clang vs. gcc and cmake.

jdahm:
Closed in favor of #598.

PR #599, opened 2020-01-16
Fix IIR Deserialization of the weighted Reduce

mroethlin:
## Technical Description

The IIR deserialization of the weighted reduce concept in the IIR was broken (weights simply were not de-serialized from the protobuf). This PR fixes that.

### Testing

New tests have been added to `unit-test/dawn/IIR/TestIIRSerializer.cpp`, testing both the weighted and unweighted versions of the ReduceOverNeighbor Concept, for all supported types of the weights. This was realized using the IIRBuilder in two separate tests.


mroethlin:
launch jenkins

PR #612, opened 2020-01-16
Update pre-commit hook and clang-format

jdahm:
## Technical Description

Replaces our use of custom clang-formatting scripts with the LLVM standard git-clang-format and makes it work on Mac. Note that this is still using clang-format version 6, but perhaps we should update that soon.

**Important Notes:**
* We are no longer using an ignore file. Entire directories (e.g. reference code for CodeGen tests) are ignored by adding another `.clang-format` file with `--- DisableFormat: true ---` inside.
* Current developers will be prompted with instructions download and setup `git-clang-format` on the first time using the pre-commit hook.


scripts/clang_format_all.sh
```diff
@@ -1,29 +0,0 @@
-#!/bin/bash
```
twicki: should we not have this file still around and make it check for `git-clang-format` and run it on the root?

jdahm: Good idea. I added a cleaned up version back in 👍

PR #614, opened 2020-01-17
Interval partition test

### LINKED ISSUE #439: Add tests for PassIntervalPartitioner
Add unit tests for `PassIntervalPartitioner`, which is not currently tested.
`PassIntervalPartitioner`, for each multistage, will first calculate a partition of the (vertical) intervals contained in the multistage (see documentation of https://github.com/MeteoSwiss-APN/dawn/blob/8b80b189eaee39df9a71b54e4333aa9bf6f1d085/dawn/src/dawn/IIR/Interval.cpp#L184 for how the partition is constructed). Then it applies this partitioning of the intervals on the DoMethods of the multistage so that at the end, each DoMethod's interval corresponds to one interval of the computed partition (check code in https://github.com/MeteoSwiss-APN/dawn/blob/master/dawn/src/dawn/Optimizer/PassIntervalPartitioner.cpp which is self-explanatory).
Could start writing one test from following gtclang code snippet (thx to @twicki):
```
#pragma gtclang merge_stages
stencil test {
  storage in, out, out2, out3, out4;
  var tmp;
  void Do() {
    vertical_region(k_start, k_end) { out = in + 1; }
    vertical_region(k_start, k_start + 2) { out2 = in + 2; }
    vertical_region(k_end - 3, k_end) { out3 = in + 3; }
    vertical_region(k_start + 1, k_end - 2) { out4 = in + 4; }
  }
};
```

eddie-c-davis:
## Technical Description

This PR adds a test class TestIntervalPartitioner that tests the PassIntervalPartitioner pass using the example provided in issue #439. 

### Resolves / Enhances

Resolves #439

### Testing

The unit test TestIntervalPartitioner.test_interval_partition reads the corresponding SIR and asserts that the expected partitions are produced.


eddie-c-davis:
launch jenkins

dawn/test/unit-test/dawn/Optimizer/Passes/TestIntervalPartitioner.cpp
```diff
@@ -0,0 +1,94 @@
+//===--------------------------------------------------------------------------------*- C++ -*-===//
```
twicki: What was the decision process to start from SIR here?

It feels like we should be able to start form a generated / deserialized IIR and just call the one pass on it?

eddie-c-davis:
launch jenkins

PR #621, opened 2020-01-20
Fix python proto copy target

jdahm:
## Technical Description

Fix the copy target for generated python proto files.


jdahm:
launch jenkins

jdahm:
launch jenkins

PR #637, opened 2020-01-21
Optimizer pass options

### LINKED ISSUE #561: Allow for individual enabeling / disabling of passes
Allow the optimizer to enable / disable independent groups of passes one by one

jdahm:
## Technical Description

Add options to turn on individual passes, which can be enabled after passing `-default-none`.

### Resolves / Enhances

Resolves #561.

### DIFF
--- a/dawn/src/dawn/Compiler/DawnCompiler.cpp
+++ b/dawn/src/dawn/Compiler/DawnCompiler.cpp
@@ -91,6 +91,10 @@ struct ComputeEditDistance<std::string> {
   }
 };
 
+bool shouldRunPass(const Options& options, bool runSpecificPass) {
+  return !options.DefaultNone || runSpecificPass;
+}
+
 } // anonymous namespace
 
 /// @brief Report a diagnostic concering an invalid Option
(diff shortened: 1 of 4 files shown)


jdahm:
launch jenkins

jdahm:
@twicki @havogt this is failing because it's building against the wrong clang-gridtools branch. As far as I can tell, "build with parameters" only let's me build branches (or sha hashes) that exist in the MeteoSwiss-APN/dawn repo. The problem is that the branch for the PR is in my fork.

jdahm:
@havogt told me that hashes from PRs work, even if the branch isn't in the main dawn repo.

jdahm:
This passes on Jenkins with a manual launch at `job/dawn_gtclang_clang-gridtools_PR/118/`

jdahm:
@twicki Could you take a look at this PR before I squash and merge to `optimizer_refactor`?

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -91,6 +91,10 @@ struct ComputeEditDistance<std::string> {
   }
 };
 
+bool shouldRunPass(const Options& options, bool runSpecificPass) {
```
twicki: if we use it like this there is no real reason to do a `checkAndPushBack` right? Would we not want to push back directly then?

jdahm: True, and checkAndPushBack and the whole PassManager is being removed in another PR :)

PR #640, opened 2020-01-22
Use experimental::fs if fs is not supported

havogt:
## Technical Description

Use std::experimental::filesystem if std::filesystem is not supported, i.e. gcc < 8.


havogt:
launch jenkins

PR #642, opened 2020-01-22
Move clang-tidy and remove 'User' key

jdahm:
## Technical Description

Clang 9 complains about the 'User' key not having a value. For example, from the vscode log:

```YAML:8:1: error: unexpected scalar
CheckOptions:
^
Error parsing /Users/johannd/Code/dawn/dawn/.clang-tidy: Invalid argument
```

Is there a reason not to have this in the root directory? This PR moves it there.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -4,7 +4,6 @@ WarningsAsErrors: ''
 HeaderFilterRegex: ''
 AnalyzeTemporaryDtors: false
 FormatStyle:     none
-User:
 CheckOptions:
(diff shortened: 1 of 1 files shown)


havogt:
> Is there a reason not to have this in the root directory? This PR moves it there.

I think about it usually as: will dawn still be complete if we just keep the dawn subdirectory. (Of course this is not fully the case e.g. with the jenkins build script.)
But I don't care much, we can move it back easily anyway...

PR #643, opened 2020-01-22
Cleanup dawn compiler

jdahm:
## Technical Description

Quick cleanup of the DawnCompiler without changing the interfaces much. Expect a follow-up PR that digs deeper.


jdahm:
I tried to make the commits each make sense and compile, so it may be easier to look at each individually when reviewing

jdahm:
launch jenkins

jdahm:
launch jenkins

PR #645, opened 2020-01-23
Fix python tests

mroethlin:
## Technical Description

This re-enables the dawn4py tests. 

## Testing

Two kinds of tests are run
* the `run.sh` compiles some stencils starting from the python SIR builder. A subset of which is compared against a reference on disk
* the python tests in `dawn/test/unit-test/test_dawn4py`using `pytest`


mroethlin:
launch jenkins

havogt:
It is possible to use the python bindings without having to pip install? The disadvantage of the proposed approach is that we, the developers, will never run these tests. They will only fail on jenkins.

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

twicki:
@havogt why should we not run the tests locally? Updating the `.so` file that is used by the installation is trivial (from the readme:)
> Changes in Python sources are instantly available in your environment. Changes in C++ sources require a recompilation of the CPython extension, which you can do by running the `develop` command of the `setup.py` script: `cd ./dawn/dawn && python setup.py develop`

so all you need is a venv in any folder where dawn is installed

havogt:
Basically I am asking for: which keys do I need to press in my IDE to run this test?

PR #655, opened 2020-01-24
Gtclang include refactor

### LINKED ISSUE #501: Refactor gtclang's processing of includes
gtclang preprocessor should add `#include "gtclang_dsl_defs/gtclang_dsl.hpp"` to the code to be parsed. Such include should not be put in the source DSL file and should not propagate to the generated code. 

We should also find a way to allow forwarding some includes of the source DSL code directly to generated code, as we may not want gtclang parser to go through them (only needed for compiling generated code).
For instance, if the user includes `driver-includes/math.hpp` in the source DSL code, we don't want it to be parsed by gtclang, but to forward it to the generated code.

eddie-c-davis:
## Technical Description

This PR allows DSL source files to be implemented without C/C++ include files with the required include and using statements added automatically before compiling.

### Resolves / Enhances

Resolves #501


eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

PR #656, opened 2020-01-24
Remove PassManager

### LINKED ISSUE #639: Remove PassManager
It doesn't do anything useful.

jdahm:
## Technical Description

Removes the PassManager, and adds a few iterator adaptors.

### Resolves / Enhances

Resolves #639.


jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
This is not my favorite PR, but I think we should accept it for now because it
1. gives the developer a sense of the order the passes are run in. It disables Pass dependency checking, but that was broken anyway. We will recover that with dependencies between pass groups once the refactor is finished.
2. is the first step toward isolating the OptimizerContext to the Optimizer, which will make the code easier to understand and dependencies cleaner. Our idea is to make Pass::run a void method, and allow it to throw if an error occurs. The optimizer can then choose to continue passes, or itself throw a CompileError, which the calling method can handle. This means that sentinel nullptrs won't be returned from the optimizer any more.

Please let me know if you disagree with the vision in (2) above so we can discuss concerns.

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -109,6 +110,24 @@ struct ComputeEditDistance<std::string> {
   }
 };
 
+template <typename P, typename... Args>
+void runPass(OptimizerContext& optimizer, std::shared_ptr<iir::StencilInstantiation> instantiation,
```
jdahm: This will be temporary.

jdahm:
I think we need a bit more time to structure this. I'm going to close for now.

PR #660, opened 2020-01-27
Validation pass

### LINKED ISSUE #638: Consider moving validation into a pass
Is there a benefit to doing so? If it's easy to do, make a PR that does this and we can discuss the pros and cons.

eddie-c-davis:
## Technical Description

This PR abstracts the IR validation tests, i.e., AST integrity, grid types, and unstructured dimensions into a validation pass. The pass manager is also modified to ensure the validation pass is performed after all other passes, to guarantee IR consistency between passes.

### Resolves / Enhances

Resolves #638


eddie-c-davis:
launch jenkins

dawn/src/dawn/Optimizer/PassValidation.cpp
```diff
@@ -0,0 +1,67 @@
...
+PassValidation::PassValidation(OptimizerContext& context) : Pass(context, "PassValidation") {}
+
+bool PassValidation::run(const std::shared_ptr<iir::StencilInstantiation>& instantiation) {
+  return run(instantiation, "");
+}
+
+bool PassValidation::run(const std::shared_ptr<iir::StencilInstantiation>& instantiation,
+                         const std::string& description) {
+  IntegrityChecker checker(instantiation.get());
+  checker.run();
```
twicki: This throws, we should catch it

dawn/src/dawn/Optimizer/PassValidation.cpp
```diff
@@ -0,0 +1,67 @@
+//===--------------------------------------------------------------------------------*- C++ -*-===//
```
twicki: validation now has a bunch of ways to do things:
-Throwing from `IntegrityChecker`
- `DAWN_ASSERT_MESSAGE`s from the grid-checkers
- returning `nullptrs` in validation

I think we should make an effort do unify things.
The idea of the `DAWN_ASSERT`s was that they could be disabled. What is the advantage of throwing?

-

dawn/src/dawn/Optimizer/PassValidation.cpp
eddie-c-davis: Exceptions are currently caught in the PassManager and added to the diagnostics.

dawn/src/dawn/Optimizer/PassValidation.cpp
eddie-c-davis: Standardized PassValidation to throw exceptions. Benefits are that we can throw/catch errors and aggregate them in the diagnostics, without halting execution and gather as many errors as possible.

eddie-c-davis:
launch jenkins

dawn/src/dawn/Optimizer/PassValidation.cpp
eddie-c-davis: Reverted grid/dimension type checks to asserts per PR #583

eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

PR #663, opened 2020-01-28
Use compatibility header for filesystem

havogt:
## Technical Description

- use compatibility header for filesystem
- move includes from header to source


havogt:
launch jenkins

PR #659, opened 2020-01-24
Add iterator adapters -- easy_iterators v1.3

jdahm:
## Technical Description

Add useful adapters and test.

Issue #651, opened 2020-01-23
Testing of individual groups of Optimization


twicki:
In #400 we found groups of independent optimization passes.
They are currently split in https://github.com/MeteoSwiss-APN/dawn/pull/637 with different flags.

DOD:
- test empty vs (group of pass1)
- test empty vs (all groups until group of pass1)
 
for all group of passes

twicki:
https://docs.google.com/spreadsheets/d/1vaQa65jZB8QfyL7i9TL4LQvxa4FOIbo_CFunUlkGQsE/edit?usp=sharing

PR #671, opened 2020-01-31
Compare IIRs with boolean operators

jdahm:
## Technical Description

Compares IIRs, ignoring the IDs. Will be useful when refactoring tests.

### Resolves / Enhances

Part of #651.

### Testing

Other test functions are replaced by this. See Dawn IIR unit tests.

### Dependencies

This is built on top of #659.


jdahm:
This is ready, but I need to clean up comments I made to myself in the code as I was reviewing it.

mroethlin:
Does this simply ignore the IDs or also attempt a remapping? While the IDs are arbitrary the relationship they impose between objects is still relevant

twicki:
I agree that ignoring the ID's completely might be the wrong approach

jdahm:
I think we need to ignore the IDs (to some extent) to be able to finish the optimizer refactoring, since _any_ amount of turning off passes or reordering groups changes the IDs. The IDs are a broken concept, but I do agree that _something_ like IDs in a few places need to exist.

Checking IDs should be all-or-nothing since anything in-between will be difficult to maintain. We've seen this already with other pass metadata getting out of sync.

I have been researching ID usage in Dawn, and I think we can replace it in most cases with a hash, or locally use IDs. As an example, the dependency graph node IDs are created from Stage's IDs, even though these could be done internally to that class.

At the Stencil, Multistage, and Stage level the IDs are essentially only used in code generation, so could be removed. At the AST level, we will need a way to refer to Fields, but name mangling will work there. For the rest of it, we can use hashes or pointers internally.

mroethlin:
I think getting rid of the IDs is quite a daunting task. However, to compare IIRs all one needs to do is find a bijective map between two sets of IDs. Maybe I'm not seeing it but that shouldn't be _too_ hard?

havogt:
My 2 cents: I would go with completely ignoring the ID's in testing (at least for now). However, I would not put this functionality into the comparison operator. The comparison operator should do the true comparison. I am undecided if the compare functions should be part of the objects or put in the test folder.

dawn/src/dawn/AST/ASTStmt.h
```diff
@@ -219,6 +219,7 @@ class BlockStmt : public Stmt {
   /// @brief removes all stmts from the block
   void clear() { statements_.clear(); }
 
+  // Make this a list!
```
havogt: Why?

jdahm:
I'm closing this for now until we determine if symbol tables are going to be the way of the future.

PR #674, opened 2020-02-04
Fix build

mroethlin:
## Technical Description

Forgot to add newly generated stencils to the appropriate `CMakeLists`


mroethlin:
launch jenkins

PR #676, opened 2020-02-04
Fix Codegen for Sparse Index 

mroethlin:
## Technical Description

The sparse index was not correctly reset between multiple reductions in the same stage. This PR fixes that and puts a test to ensure that this functionality works. 

### Testing

A new test was introduced in the `AtlasIntegrationTest`s


mroethlin:
launch jenkins

PR #677, opened 2020-02-04
Python tests and dawn4py setup

jdahm:
## Technical Description

Fix python bindings and update Dockerfile.

Many updates from my week-or-so deep dive on dawn4py and gt4py. What this does:
- Updates Dockerfile
- Rename `DAWN_REQUIRE_PYTHON_TESTING` to `DAWN_REQUIRE_PYTHON`
- Correct HINTS -> PATHS for protobuf detection in DawnConfig.cmake.in
- Moves top-level CMake files into both `{dawn,gtclang}/cmake/` directories, since Python `pip` needs self-contained directories
- Cleans up `dawn/examples/python/run.sh` and rename to `generate_and_diff`
- Cleanup `dawn/setup.py` to allow for out-of-source (non `-e`) `pip install dawn/dawn4py`, full wheel support, cleanup sub-project CMakeCache.txt files, use self.spawn correct, etc.
- Corrects placement of SIR and IIR python protobuf files in the `pip install` process (this was not happening before)
- Remove QUIET from GridTools detection. It can happen that find_package(GridTools)` fails if OpenMP is not detected, so this now prints out a reason for GridTools not being detected
- Cleans up `dawn/scripts/jenkins/build.sh`
- Python protobuf files are generated in the build directory, not in source


dawn/examples/python/run.sh
```diff
@@ -1,15 +0,0 @@
-#!/bin/bash
```
jdahm: We need to update the Jenkins tests to use `generate_and_diff` instead of this. See `Dockerfile` for specific command.

jdahm:
launch jenkins

dawn/CMakeLists.txt
```diff
@@ -123,7 +124,7 @@ endfunction()
 
 # Protobuf
 if(NOT DAWN_BUILD_PROTOBUF)
-  find_package(Protobuf 3.4.0 CONFIG)
```
twicki: what are the implications of this?

dawn/setup.py
```diff
@@ -48,119 +89,107 @@ class CMakeBuild(build_ext):
...
-        cmake_executable = os.getenv("CMAKE_EXECUTABLE", default="cmake")
-        cmake_args = os.getenv("CMAKE_ARGS", default="").split(" ") or []
-        try:
-            out = subprocess.check_output([cmake_executable, "--version"])
-        except OSError:
-            raise RuntimeError(
-                "CMake must be installed to build the following extensions: "
-                + ", ".join(e.name for e in self.extensions)
+        dawn_build_dir = os.getenv("DAWN_BUILD_DIR", default=None)
+        py_protos = [x + "_pb2.py" for x in ("IIR/IIR", "SIR/SIR", "SIR/statements", "SIR/enums")]
```
twicki: is this future-proof? can we not find files in the `SIR` / `IIR` subfolders?

scripts/jenkins/build.sh
```diff
@@ -38,60 +38,54 @@ done
...
   PARALLEL_BUILD_JOBS=8
 fi
 
-echo "Building with ${PARALLEL_BUILD_JOBS} jobs."
-cmake --build . --parallel ${PARALLEL_BUILD_JOBS}
+echo "Building with $PARALLEL_BUILD_JOBS jobs."
+cmake --build $build_dir --config $build_type --parallel $PARALLEL_BUILD_JOBS
 
 # Run unittests
-ctest -C ${build_type} --output-on-failure --force-new-ctest-process
```
twicki: now we only have the python tests running and not eh other ones?

dawn/CMakeLists.txt
jdahm: It is necessary to remove CONFIG if we want to have protobuf found from system packages. If you want to use a custom protobuf version that's still fine, but pass `Protobuf_DIR` appropriately. It should break nothing. @havogt do you see an issue with this?

dawn/setup.py
jdahm: Sure, I can find files there instead. I'll fix in another commit.

scripts/jenkins/build.sh
jdahm: It looks like I kept this line. However, I removed the `-C ${build_type}` since that is taken care of in the build command on line 71.

jdahm:
@havogt and @twicki 

First of all, thank you so much for looking at this large PR. I'm a bit embarrassed that all this went into one giant PR instead of a series of PRs as it should have been, but c'est la vie.

> Do you already have a recommended setup for developing dawn including dawn4py?

As long as `dawn4py` is in the same repo as `dawn`, and since `dawn4py` is exposing the interfaces of `dawn`, it should be tested alongside `dawn`.

Testing `gt4py` and `dawn` is more complicated. Ideally once the dawn interface is simplified, we almost don't need python bindings, but `gt4py` and `dawn` should be tested independently. There should be tested added that ensure GT4py is generating valid SIR, but that should be independent of the dawn project itself.

All that above is to say, long term I think we should split dawn into three projects (could be all in the same repo at first) `{dawn-irs, dawn-opt, dawn-codegen}`. What do you think of this proposal?

> Currently we are also running the setup of dawn4py in the jenkins plan after testing dawn+gtclang, which adds another 4-5 minutes, because it builds dawn again. I would move this into a separate plan, what do you think?

I've exposed `DAWN_BUILD_DIR` that makes this process faster: if set and `setup.py` finds all python-related files built, the python module build time is very fast.

An example is in the Dockerfile:
```
ENV DAWN_BUILD_DIR /usr/src/dawn/build/dawn
RUN python -m pip install /usr/src/dawn/dawn
```

jdahm:
launch jenkins

havogt:
>As long as dawn4py is in the same repo as dawn, and since dawn4py is exposing the interfaces of dawn, it should be tested alongside dawn.

Should I continue on my branch where I include the python examples into CMake or do you have a better solution? What I wouldn't like is to have to execute 2 commands to cover the standard tests (ctest + some separate python command).

> All that above is to say, long term I think we should split dawn into three projects (could be all in the same repo at first) {dawn-irs, dawn-opt, dawn-codegen}. What do you think of this proposal?

I am always pro merging and against splitting repositories, unless there is a standalone use-case for one of the repositories.

> I've exposed DAWN_BUILD_DIR that makes this process faster: if set and setup.py finds all python-related files built, the python module build time is very fast.

The setup.py would just re-use the build files and it will just copy things around?

jdahm:
launch jenkins

scripts/jenkins/build.sh
```diff
@@ -38,60 +38,48 @@ done
...
-
-python -m pytest -v ${base_dir}/dawn/test/unit-test/test_dawn4py/
-echo "PYTHON PYTEST SUCCESFUL!"
-
-base_dir=$(pwd)
-rm -rf $build_dir
-
+build_dir=$base_dir/build
 mkdir -p $build_dir
-cd $build_dir
```
twicki: is this still working?

scripts/jenkins/build.sh
```diff
@@ -38,60 +38,48 @@ done
...
+echo "Building with $PARALLEL_BUILD_JOBS jobs."
+cmake --build $build_dir --config $build_type --parallel $PARALLEL_BUILD_JOBS
+
+# Run tests -- this also runs python tests
+ctest --output-on-failure --force-new-ctest-process
 
-# Run unittests
-ctest -C ${build_type} --output-on-failure --force-new-ctest-process
+# Test installation
+cmake --build $build_dir --config $build_type --parallel $PARALLEL_BUILD_JOBS --target install
```
twicki: rm config

dawn/CMakeLists.txt
```diff
@@ -43,9 +43,10 @@ unset(__DAWN_VERSION)
 option(BUILD_EXAMPLES "Build examples" ON)
 option(DAWN_BUILD_TESTING "Override BUILD_TESTING if part of a multi-project build." OFF)
 option(DAWN_BUILD_EXAMPLES "Override BUILD_EXAMPLES if part of a multi-project build." OFF)
-option(DAWN_REQUIRE_PYTHON_TESTING "Require enabling python." OFF)
-set(PROTOBUF_PYTHON_MODULE_DIR "" CACHE PATH "Path to protobuf python module, if not at $Protobuf_INCLUDE_DIR/..")
-mark_as_advanced(PROTOBUF_PYTHON_MODULE_DIR)
+option(DAWN_REQUIRE_PYTHON "Require enabling python." OFF)
+set(PROTOBUF_PYTHON_MODULE_DIR "" CACHE PATH "Path to protobuf python module. Use this if CMake is not finding the module.")
+set(DAWN4PY_MODULE_DIR "${PROJECT_SOURCE_DIR}/src" CACHE PATH "Path to dawn4py python module being built. Change this to self.build_lib for pip out-of-source builds.")
```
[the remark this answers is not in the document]
jdahm: See the cache property description.

dawn/cmake/DawnProtobufGenerate.cmake
```diff
@@ -111,16 +120,16 @@ function(dawn_protobuf_generate)
     unset(generated_srcs)
     foreach(ext ${extensions})
       if(${ARG_LANGUAGE} STREQUAL "java" )
-        list(APPEND generated_srcs "${CMAKE_CURRENT_BINARY_DIR}/dawn/sir/${basename}${ext}")
+        list(APPEND generated_srcs "${output_directory}/sir/${basename}${ext}")
```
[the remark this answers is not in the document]
jdahm: Not sure. For the sake of time I'm going to put `dawn/` back and put off checking into this.

scripts/jenkins/build.sh
jdahm: Updated. Thanks for catching this!

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

dawn/setup.py
```diff
@@ -18,23 +18,65 @@
...
+AUTHOR = "MeteoSwiss / ETH Zurich / Vulcan"
+
+# Note that DAWN_DIR below may be the version pip copies over before building
 DAWN_DIR = os.path.join(os.path.dirname(__file__))
-DAWN_ABS_DIR = os.path.abspath(DAWN_DIR)
 BUILD_JOBS = 4
 
 # Select protobuf version
-# TODO: avoid parsing python files and adapt to new CMake
 with open(os.path.join(DAWN_DIR, "cmake", "FetchProtobuf.cmake"), "r") as f:
```
egparedes: This is hardcoding the python protobuf package version to be installed to the version selected in `FetchProtobuf.cmake`. This creates a potential problem if there is a valid protobuf version installed in the system with a different version which gets used by CMake to build the binaries. In that case, the python protobuf package and binary protobuf will have different versions.

A simple solution would be to always use the FetchProtobuf version when building the `python` target, which would also fix the other problem I was experiencing, when a too old version of protobuf installed in the system messes up with the CMake build configuration.

dawn/src/dawn4py/CMakeLists.txt
```diff
@@ -37,8 +69,6 @@ pybind11_add_module(_dawn4py MODULE _dawn4py.cpp)
 target_compile_features(_dawn4py PUBLIC cxx_std_17)
 target_link_libraries(_dawn4py PUBLIC pybind11::pybind11 Dawn)
 target_compile_options(_dawn4py PUBLIC -Wno-shadow)
-
-# Depending on DawnSIRProto and DawnIIRProto ensures the python files are
-# generated and moved into the correct location
-# TODO Do we need to depend on Dawn or DawnC here?
-add_custom_target(python DEPENDS _dawn4py)
+set_target_properties(_dawn4py PROPERTIES
```
havogt: Why did you changes this? The default seems to be the better choice to me. To me `pybind11_add_module` should behave like `add_library`.

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
Fix python bindings and update Dockerfile.

Many updates from my week-or-so deep dive on dawn4py and gt4py. What this does:

    Updates Dockerfile
    Rename DAWN_REQUIRE_PYTHON_TESTING to DAWN_REQUIRE_PYTHON
    Correct HINTS -> PATHS for protobuf detection in DawnConfig.cmake.in
    Moves top-level CMake files into both {dawn,gtclang}/cmake/ directories, since Python pip needs self-contained directories
    Cleans up dawn/examples/python/run.sh and rename to generate_and_diff
    Cleanup dawn/setup.py to allow for out-of-source (non -e) pip install dawn/dawn4py, full wheel support, cleanup sub-project CMakeCache.txt files, use self.spawn correct, etc.
    Corrects placement of SIR and IIR python protobuf files in the pip install process (this was not happening before)
    Remove QUIET from GridTools detection. It can happen that find_package(GridTools)` fails if OpenMP is not detected, so this now prints out a reason for GridTools not being detected
    Cleans up dawn/scripts/jenkins/build.sh

PR #680, opened 2020-02-09
Fix SIR output path

jdahm:
## Technical Description

Output SIR to current working directory inside gtclang when using the `-write-sir` flag.


jdahm:
launch jenkins

PR #698, opened 2020-02-12
Implementation of PassRemoveScalars

### LINKED ISSUE #670: Remove scalar local variables
As discussed (see https://stackoverflow.com/c/dawn/questions/36 ) we want to inline scalar local variables.
DoD: write and test a pass to remove local variables flagged as scalars.

Stagno:
## Technical Description

- Implement and test `PassRemoveScalars` to remove local variables flagged as scalars.
**if-statements with scalar conditions are not supported yet**: as the pass would need to do something completely different for them, it is better to run a preprocessing step that transforms the if-else constructs into just statements with ternary operators and then run this pass. Warning issued when encountered.
**Compound statements and increment/decrement ops are not supported**: they will be removed before this pass runs or we will need to deal with them here. 
- Add `-freport-pass-remove-scalars` option to dawn.
- Change some `const char*` to `std::string`

### Resolves / Enhances

resolves #670


Stagno:
launch jenkins

Stagno:
launch jenkins

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -0,0 +1,241 @@
...
+}
+} // namespace
+
+bool PassRemoveScalars::run(
+    const std::shared_ptr<iir::StencilInstantiation>& stencilInstantiation) {
+  // TODO: sir from GTClang currently not supported: it introduces compound assignments (handling
+  // them complicates this pass). For now disabling the pass for cartesian grids.
+  // We should remove compound assignments and increment/decrement ops from SIR/IIR. They are
+  // syntactic sugar, only useful at DSL level.
+  if(stencilInstantiation->getIIR()->getGridType() == ast::GridType::Cartesian) {
```
twicki: what do compound statements have to do with the grid type?

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -0,0 +1,241 @@
...
+    return true;
+  }
+  for(const auto& stencilPtr : stencilInstantiation->getStencils()) {
```
twicki: why two loops?

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -191,6 +192,7 @@ DawnCompiler::lowerToIIR(std::shared_ptr<SIR> const& stencilIR) {
     optimizer.pushBackPass<PassSetSyncStage>();
     // validation checks after parallelisation
     optimizer.pushBackPass<PassLocalVarType>();
+    optimizer.pushBackPass<PassRemoveScalars>();
```
twicki: For now this pass is not required to run. It does not contribute to validity, right? This will only come in if we have the stage-splitter / stage location type

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -292,6 +294,7 @@ DawnCompiler::optimize(std::map<std::string, std::shared_ptr<iir::StencilInstant
     // since this can change the scope of temporaries ...
     optimizer.pushBackPass<PassTemporaryType>();
     optimizer.pushBackPass<PassLocalVarType>();
+    optimizer.pushBackPass<PassRemoveScalars>();
```
twicki: how can we generate new scalars with merging?

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
Stagno: Cartesian tests and clang-gridtools stencils currently contain compound statements, while unstructured tests/examples don't. As this pass is only critical for unstructured, enabling it for cartesian is deferred to a follow up task.

dawn/src/dawn/Compiler/DawnCompiler.cpp
Stagno: Correct. Still worth keeping it for integration tests (e.g. checking outputs are numerically the same even after removal of scalars).

dawn/src/dawn/Compiler/DawnCompiler.cpp
Stagno: If you never assign an expression with dimensions to a temporary, after merging stages it will become a local variable which is flagged as scalar by `PassLocalVarType`.

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
Stagno: Not needed, correcting

Stagno:
launch jenkins

dawn/src/dawn/Compiler/DawnCompiler.cpp
twicki: That is not how this works. 

Create a group for this pass

dawn/src/dawn/Compiler/DawnCompiler.cpp
twicki: got it

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
twicki: Nothing prevents compound statements from being used in unstructured grids.

This just seems to be a workaround to get the tests to pass somehow and seems far from ideal

Stagno: I realize that from what I wrote it seems a workaround. But https://github.com/MeteoSwiss-APN/dawn/blob/b30672a5b26f834e852b4c7121dbe4d071a87510/dawn/src/dawn/Optimizer/PassRemoveScalars.cpp#L108-L111
actually makes compound statements illegal for unstructured. There are currently no use cases for them in our unstructured examples/tests.

dawn/src/dawn/Compiler/DawnCompiler.cpp
Stagno: But `PassLocalVarType` is also in the parallel group, what is the difference? Both will be necessary to have a valid parallel IIR, which is currently not valid because the stage location types are incorrect.
I'd say it makes sense to test them in the group where they will be at the end, not in another one which will be removed. Unless you suggest to create a group that runs before the parallel group.

Stagno:
launch jenkins

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
```diff
@@ -0,0 +1,587 @@
...
+
+std::shared_ptr<iir::Stmt> getNthStmt(std::shared_ptr<iir::StencilInstantiation>& si, int n) {
+  return getDoMethod(si).getAST().getStatements()[n];
+}
+
+bool isVarInDoMethodsAccesses(int varAccessID, const iir::DoMethod& doMethod) {
+  for(const auto& stmt : doMethod.getAST().getStatements()) {
+    const auto& access = stmt->getData<iir::IIRStmtData>().CallerAccesses;
+
+    for(auto& accessPair : access->getWriteAccesses()) {
```
twicki:
```suggestion
 if ( std::any_of(access->getWriteAccesses().begin(), access->getWriteAccesses().end(), [&](int i){return i==varAccessID;}) )
```

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
```diff
@@ -0,0 +1,587 @@
...
+      if(varAccessID == accessPair.first) { // first = AccessID
+        return true;
+      }
+    }
+    for(const auto& accessPair : access->getReadAccesses()) {
```
twicki: any of again

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
```diff
@@ -0,0 +1,587 @@
...
+  /// f_c = varA;
+
+  auto stencil = b.build(
+      "generated",
+      b.stencil(b.multistage(
+          dawn::iir::LoopOrderKind::Forward,
+          b.stage(b.doMethod(dawn::sir::Interval::Start, dawn::sir::Interval::End,
+                             b.declareVar(varA), b.stmt(b.assignExpr(b.at(f_c), b.at(varA))))))));
+
+  // Setup variable's metadata before running pass
```
twicki: why is this step necessary to be taken manually here?

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
```diff
@@ -0,0 +1,587 @@
...
+
+iir::DoMethod& getDoMethod(std::shared_ptr<iir::StencilInstantiation>& si) {
+  auto& iir = si->getIIR();
+  auto& stencil = iir->getChild(0);
+  auto& ms = stencil->getChild(0);
+  auto& stage = ms->getChild(0);
+  return *stage->getChild(0);
+}
+
+std::shared_ptr<iir::Stmt> getNthStmt(std::shared_ptr<iir::StencilInstantiation>& si, int n) {
```
twicki: should something like this not live in unittest utils?

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
ofuhrer: I agree that we should avoid if-statements like this in the code base as much as possible. I do feel like if the effort is reasonable we should take care to support all use-cases with new features.

havogt: This if is a temporary solution until we do the cleanup of removing the unnecessary compound assignments from SIR.

dawn/src/dawn/Compiler/DawnCompiler.cpp
Stagno: No, I'm not saying that. When we discussed where to put `PassLocalVarType` you made me notice that, at the moment, it's not required for stage splitting so it doesn't need to be put before `PassStageSplitter`, so we put it at the end of the group. I'd say the same goes for `PassRemoveScalars`.
As we cannot implement everything in one big step (flag variables + remove scalars + fix stage location type), it seems reasonable to incrementally add passes to the group in order to get them covered by integration tests.

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
Stagno: My intention was to introduce it incrementally, first unstructured then cartesian.
But I understand this is not what everybody wants, so the PR will have to wait.

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
```diff
@@ -0,0 +1,587 @@
...
+      }
+    }
+  }
+  return false;
+}
+
+TEST(TestRemoveScalars, test_unstructured_scalar_01) {
+  using namespace dawn::iir;
+
+  UnstructuredIIRBuilder b;
```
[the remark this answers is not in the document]
Stagno: Current state is that local variable types are derived info in the sense that they are not serialized, but they are a necessary input for this pass.

dawn/src/dawn/AST/ASTUtil.cpp
```diff
@@ -244,7 +244,7 @@ enum class OpKind {
 /// @brief String to OpKind
 ///
 /// Note that we do not evaluate bit-wise operations and the modulo operator
-OpKind toOpKind(const char* op) {
+OpKind toOpKind(const std::string op) {
```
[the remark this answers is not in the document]
Stagno: fixing

dawn/src/dawn/AST/ASTUtil.cpp
```diff
@@ -273,7 +273,7 @@ static bool evalImpl(ResultType& result, ValueTypes... operands) {
 //
 // Unary operations
 //
-bool evalUnary(const char* opStr, double& result, double operand) {
+bool evalUnary(const std::string opStr, double& result, double operand) {
```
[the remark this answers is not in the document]
Stagno: fixing

dawn/src/dawn/AST/ASTUtil.cpp
```diff
@@ -288,7 +288,7 @@ bool evalUnary(const char* opStr, double& result, double operand) {
 //
 // Binary operations
 //
-bool evalBinary(const char* opStr, double& result, double op1, double op2) {
+bool evalBinary(const std::string opStr, double& result, double op1, double op2) {
```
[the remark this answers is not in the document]
Stagno: fixing

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -310,6 +313,7 @@ DawnCompiler::optimize(std::map<std::string, std::shared_ptr<iir::StencilInstant
     // safe idea
     optimizer.pushBackPass<PassTemporaryType>();
     optimizer.pushBackPass<PassLocalVarType>();
+    optimizer.pushBackPass<PassRemoveScalars>();
```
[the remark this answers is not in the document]
Stagno: I think you are right, because when merging temporary fields we never decrease their scope. Therefore we never need to demote them to local variables. This also means that `PassTemporaryType` and `PassLocalVarType` are not needed, correct?

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -327,6 +331,7 @@ DawnCompiler::optimize(std::map<std::string, std::shared_ptr<iir::StencilInstant
       // since this can change the scope of temporaries ...
       optimizer.pushBackPass<PassTemporaryType>();
       optimizer.pushBackPass<PassLocalVarType>();
+      optimizer.pushBackPass<PassRemoveScalars>();
```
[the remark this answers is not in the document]
Stagno: `PassTemporaryType` in this case will only promote, never demote, right?
Then yes, both `PassLocalVarType` and 'PassRemoveScalars` are not needed.

dawn/src/dawn/IIR/AccessComputation.cpp
```diff
@@ -450,15 +450,20 @@ class AccessMapper : public iir::ASTVisitor {
...
                      ArrayRef<std::shared_ptr<iir::Stmt>> stmts) {
   for(const auto& stmt : stmts) {
-    DAWN_ASSERT(instantiation);
-    AccessMapper mapper(instantiation->getMetaData(), stmt, nullptr);
+    AccessMapper mapper(metadata, stmt, nullptr);
     stmt->accept(mapper);
   }
 }
 
+void computeAccesses(iir::StencilInstantiation* instantiation,
```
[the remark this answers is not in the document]
Stagno: nope, removing

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include "dawn/IIR/StencilInstantiation.h"
+#include "dawn/IIR/StencilMetaInformation.h"
+
+#include <tuple>
+#include <unordered_map>
+
+namespace dawn {
+namespace {
+
+// TODO: replace with a more general expr.getDimensions() = Scalar | OnEdges | ...
```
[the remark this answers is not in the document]
Stagno: For now `ExprScalarChecker` is sufficient. I added the todo anticipating that, when fixing the stage location type, we could end up having such function, which can be reused here. Changing the comment to clarify.

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -0,0 +1,239 @@
...
+/// f_c = !_ifCond_0 ? varA * varB : f_c;
+///
+/// than this pass (PassRemoveScalars) will remove the scalars _ifCond_0, varA, varB, ...
+/// another pass could also remove the ternary operators if the conditions are literals.
+void removeScalarsFromBlockStmt(
+    iir::BlockStmt& blockStmt,
+    std::unordered_map<int, std::shared_ptr<const iir::Expr>>& scalarToLastRhsMap,
+    const iir::StencilMetaInformation& metadata) {
+
+  for(auto stmtIt = blockStmt.getStatements().begin(); stmtIt != blockStmt.getStatements().end();) {
```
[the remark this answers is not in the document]
Stagno: There's no increment expression

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -0,0 +1,239 @@
...
+    if(idToLocalVariableData.at(*scalarAccessID).isScalar()) {
+      return scalarAccessID;
+    }
+  }
+  return std::nullopt;
+}
+
+// Returns the rhs `Expr` of a variable declaration / assignment.
+std::shared_ptr<iir::Expr> getRhsOfAssignment(const std::shared_ptr<iir::Stmt> stmt) {
+  if(const auto& varDeclStmt = std::dynamic_pointer_cast<iir::VarDeclStmt>(stmt)) {
```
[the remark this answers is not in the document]
Stagno: Only for an if that deals with a single variable, e.g.
```
if(varA > 0.0) {
   varA = 0.0;
}
```
But in general you can have multiple variables, fields in there.
See the example of https://github.com/MeteoSwiss-APN/dawn/issues/722

dawn/src/dawn/Compiler/DawnCompiler.cpp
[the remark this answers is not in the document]
twicki: agreed

dawn/src/dawn/Compiler/DawnCompiler.cpp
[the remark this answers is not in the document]
twicki: correct

dawn/src/dawn/Compiler/DawnCompiler.cpp
twicki: sorry. I still do not follow.

In the parallel group there should only be the passes that are *absolutely* required to generate valid, parallel IR.
So if both of these passes are only required for optimization they should appear only in a group below (or be their own group) and if they are necessary up here, they can stay

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
```diff
@@ -0,0 +1,587 @@
...
+#include "dawn/Optimizer/PassRemoveScalars.h"
+#include "dawn/Unittest/IIRBuilder.h"
+
+#include <gtest/gtest.h>
+
+using namespace dawn;
+
+namespace {
+
+std::shared_ptr<iir::Expr> getRhsOfAssignment(const std::shared_ptr<iir::Stmt> stmt) {
```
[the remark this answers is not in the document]
Stagno: I don't know, isn't exposing the private API just for testing a bad practice also? https://enterprisecraftsmanship.com/posts/unit-testing-private-methods/

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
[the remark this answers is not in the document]
twicki: sorry

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
```diff
@@ -0,0 +1,587 @@
...
+      const auto& assignmentExpr =
+          std::dynamic_pointer_cast<iir::AssignmentExpr>(exprStmt->getExpr());
+      return assignmentExpr->getRight();
+    }
+  }
+
+  return nullptr;
+}
+
+iir::DoMethod& getDoMethod(std::shared_ptr<iir::StencilInstantiation>& si) {
```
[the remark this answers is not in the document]
Stagno: changing

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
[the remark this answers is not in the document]
twicki: I only found an ok solution to this. Possible solutions I see are

- refactor tests to not require this
- add it to a header to use
- include a cpp file
- copy past code

If i missed a better option please let me know. I kind of ordered them how i feel best to worst. What is your suggestion?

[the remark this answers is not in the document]
Stagno: First option is ideal, going for it. 
I took the chance to refactor the way the ast is checked after running the pass, by using the API of `UnittestStmtSimplifier.h` (moved from gtclang) to create the target and comparing asts ignoring data (extended the `equals()` method of stmts/exprs). This is the commit https://github.com/MeteoSwiss-APN/dawn/pull/698/commits/6af292cb16a9b41ce7b3b026b6aa5770bb0e287f if you want to check it individually.

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
Stagno: Can extend `IIRBuilder::localvar()` to allow setting the type, so this won't be needed.

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
Stagno: Moving this to utils

dawn/src/dawn/Compiler/DawnCompiler.cpp
Stagno: They are not an optimization, they are required by `PassStageSplitter`. But currently `PassStageSplitter` does not the correct thing, as it does not set the correct location types of stages and thus it does not use the output of these passes.

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
Stagno: Having discussed this with the others, we believe that for now we can live with this temporary solution. Once a decision is taken we will change this.

Stagno:
launch jenkins

Stagno:
launch jenkins

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
Stagno: @twicki you can continue with the review, so that we can try to merge by tomorrow. I'll create the follow up task to extend the implementation to cartesian.

twicki: why can we not move forward with fixing this dependency first and merge this once that is done? I see no strong reason to merge this is that form, or am I missing something?

dawn/src/dawn/Compiler/DawnCompiler.cpp
twicki: ok

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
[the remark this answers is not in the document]
twicki: ok

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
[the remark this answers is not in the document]
twicki: understood

dawn/test/unit-test/dawn/Optimizer/Passes/TestPassRemoveScalars.cpp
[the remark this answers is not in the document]
Stagno: What we understood is that existing tests from sir/DSL would be changed into serialized iir, not that we would disallow using the IIRBuilder for unit testing passes.
Constructing one of the input IIRs, in this case, is trivially done with the IIRBuilder in less than 10 lines of code.

Concerning local variable types: I think they better fit as derived info because they are always consumed by the passes that run immediately after `PassLocalVarType` and only by them. Same pattern as for the stage graph.
If we have a reason to change that, however, doing it is outside the scope of this PR.

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
Stagno: Fixing the dependency would mean writing new code either:

- in this pass to treat compound statements, increment, decrement ops as special cases, but such code would need to be deleted if we decide to remove them from iir, or
- in another pass to remove c.s., ++, --, but such pass won't be needed if we decide to remove them from sir.

The temporary solution is to not waste time in writing code that might be deleted later and to allow us to proceed in the path of fixing stage location types, which is critical for unstructured, without waiting for a decision on c.s., ++, --.

Stagno:
launch jenkins

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -195,11 +182,42 @@ void removeScalarsFromDoMethod(iir::DoMethod& doMethod, iir::StencilMetaInformat
...
+  return false;
+}
+
 bool PassRemoveScalars::run(
     const std::shared_ptr<iir::StencilInstantiation>& stencilInstantiation) {
-  // TODO: add support for cartesian
-  if(stencilInstantiation->getIIR()->getGridType() == ast::GridType::Cartesian) {
-    return true;
+  // Check if we have unsupported statements. If we do, warn the user and skip the pass execution.
+  for(const auto& stmt : iterateIIROverStmt(*stencilInstantiation->getIIR())) {
```
twicki: This looks excellent

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
twicki: This is now resolved, we just do not execute this if we find unsupported statements

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

PR #716, opened 2020-02-14
Fix for GCC on MacOS

jdahm:
## Technical Description

Adds correct include path for GCC on MacOS. Code is only changed inside an __APPLE__ macro check.

### DIFF
--- a/gtclang/src/gtclang/Driver/CompilerInstance.cpp
+++ b/gtclang/src/gtclang/Driver/CompilerInstance.cpp
@@ -91,8 +91,8 @@ clang::CompilerInstance* createCompilerInstance(llvm::SmallVectorImpl<const char
 
 #ifdef __APPLE__
   // Set the root where system headers are located.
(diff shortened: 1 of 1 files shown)


gtclang/src/gtclang/Driver/CompilerInstance.cpp
```diff
@@ -93,6 +93,8 @@ clang::CompilerInstance* createCompilerInstance(llvm::SmallVectorImpl<const char
   // Set the root where system headers are located.
   ccArgs.push_back("-internal-isystem");
   ccArgs.push_back(GTCLANG_CLANG_RESSOURCE_INCLUDE_PATH "/../../../../include/c++/v1/");
+  ccArgs.push_back("-internal-isystem");
```
twicki: this arg is pushed back twice. Is that intended?

jdahm: Yes. It's like `-I` but internally, so the path follows that argument each time.

PR #719, opened 2020-02-17
Adding a Utility to Load NetCDF Files into Atlas

### LINKED ISSUE #696: Create a utility to load a ICON mesh into Atlas
Currently, we are testing our stencils using either Atlas or mylib with the mesh generators therein. To ensure that we are conforming to the ICON grid, a utility is to be created to load a ICON grid into Atlas.
Coordinate with Hannes.

DoD: write the utility and run the ICON Laplacian on the loaded grid.

mroethlin:
## Technical Description

This PR introduces a simple utility to load netcdf files as offered by the DWD web service into Atlas

### Resolves / Enhances

Resolves #696 

### Testing

There is no CI testing for this. The `README` in the `prototype` folder was updated to explain how a very simple test program can be compiled and run that uses the utility to read a netcdf file and dump it as `gmsh`file.


dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+static void AllocNbhTable(atlas::mesh::HybridElements::Connectivity& connectivity, int numElements,
+                          int nbhPerElem) {
+  std::vector<int> init(numElements * nbhPerElem, connectivity.missing_value());
+  connectivity.add(numElements, nbhPerElem, init.data());
+}
+
+static void AllocNbhTable(atlas::mesh::Nodes::Connectivity& connectivity, int numElements,
+                          int nbhPerElem) {
+  std::vector<int> init(numElements * nbhPerElem, connectivity.missing_value());
+  connectivity.add(numElements, nbhPerElem, init.data());
```
havogt: This should create a dangling ptr to `init.data()`. Or do I miss something?

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+    const int cellsPerEdge = 2;
+    const int cellsPerNode = 6; // maximum is 6, some with 5 exist
+    const int edgesPerNode = 6; // maximum is 6, some with 5 exist
+    const int edgesPerCell = 3;
+
+    // Allocate neighbor tables
+    // ------------------------
+
+    // Edges
+    AllocNbhTable(mesh.edges().cell_connectivity(), mesh.edges().size(), cellsPerEdge);
```
havogt: I'd move the allocs just before each call to Addneighborlist to make it easier to read.

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+                                 cellToVertex[2 * ncells + cellIdx] - 1};
+    node_connectivity.set(cellIdx, tri_nodes);
+    glb_idx_cell[cellIdx] = cellIdx;
+    cells_part(cellIdx) = myPart;
+  }
+
+  return true;
+}
+
+static bool AddNeighborList(const netCDF::NcFile& dataFile, atlas::Mesh& mesh,
```
havogt: Remove code duplication by introducing a template (or can it be solved with runtime polymorphism even?)

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+  auto flags = atlas::array::make_view<int, 1>(nodes.flags());
+
+  auto radToLat = [](double rad) { return rad / (0.5 * M_PI) * 90; };
+  auto radToLon = [](double rad) { return rad / (M_PI)*180; };
+
+  for(int nodeIdx = 0; nodeIdx < numNodes; nodeIdx++) {
+    xy(nodeIdx, atlas::LON) = radToLon(lon[nodeIdx]);
+    xy(nodeIdx, atlas::LAT) = radToLat(lat[nodeIdx]);
+
+    lonlat(nodeIdx, atlas::LON) = xy(nodeIdx, atlas::LON);
```
havogt: `lonlat` and `xy` are now set to the same value. Please comment why this makes sense, if it makes sense.

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+#include <atlas/mesh/actions/BuildEdges.h>
+#include <atlas/mesh/actions/BuildPeriodicBoundaries.h>
+#include <atlas/meshgenerator/detail/MeshGeneratorImpl.h>
+#include <atlas/output/Gmsh.h>
+#include <atlas/util/CoordinateEnums.h>
+
+static const int myPart = 0;
+
+template <typename loadT>
+static std::vector<loadT> LoadField(const netCDF::NcFile& dataFile, const std::string& name) {
```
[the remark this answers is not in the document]
twicki: I just want to pitch https://llvm.org/docs/CodingStandards.html#anonymous-namespaces here

dawn/prototype/AtlasFromNetcdf.cpp
mroethlin: unfortunately runtime polymorphism is not possible because the connectivities use a Interface<Impl> pattern, solved with template.

dawn/prototype/AtlasFromNetcdf.cpp
mroethlin: On first sight it seemed that this is the default behavior of the `RegularMeshGenerator.cc` of atlas. Studying the code again this is in fact not the case, but the x,y index on the regular lattice is assigned. There is no such lattice for the ICON netcdf files, and to compute x,y values any of the map projections would need to be used. Hence, I chose to simply not assign `xy` and documented this behavior.

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+  netCDF::NcVar data = dataFile.getVar(name.c_str());
+  if(data.isNull()) {
+    return {};
+  }
+  assert(data.getDimCount() == 1);
+  int size = data.getDim(0).getSize();
```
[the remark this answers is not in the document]
mroethlin: It returns `size_t`. Changed this here, but I'm converting implicitly to `int` later because those values are used as loop boundaries and I have been burned by stuff like `for (size_t i = 0; i < v.size() - 1; i++) { }`  too often.

dawn/prototype/AtlasFromNetcdf.cpp
mroethlin: The naming in atlas is confusing here. `Add` copies the data in init: https://github.com/ecmwf/atlas/blob/develop/src/atlas/mesh/Connectivity.cc#L139

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+#include <atlas/mesh/HybridElements.h>
+#include <atlas/mesh/Nodes.h>
+#include <atlas/mesh/actions/BuildCellCentres.h>
+#include <atlas/mesh/actions/BuildEdges.h>
+#include <atlas/mesh/actions/BuildPeriodicBoundaries.h>
+#include <atlas/meshgenerator/detail/MeshGeneratorImpl.h>
+#include <atlas/output/Gmsh.h>
+#include <atlas/util/CoordinateEnums.h>
+
+static const int myPart = 0;
```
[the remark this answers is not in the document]
mroethlin: we do not support partitioning yet, so the `part` fields of both nodes and cells are all set to the same dummy identifier. I wanted to express that this is the same everywhere without passing this around as an argument. Not the most elegant solution but I can't think of any better options

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,298 @@
...
+    atlas::mesh::Nodes::Topology::reset(flags(nodeIdx));
+  }
+
+  return true;
+}
+
+static bool CellsFromNetCDF(const netCDF::NcFile& dataFile, atlas::Mesh& mesh) {
+  auto [cellToVertex, vertexPerCell, ncells] = Load2DField<int>(dataFile, "vertex_of_cell");
+  if(vertexPerCell != 3) {
+    std::cout << "not a triangle mesh\n";
```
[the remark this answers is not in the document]
mroethlin: I will leave this as is for now, if this becomes part of dawn or rather the dawn integration testing we will need to adapt to dawn exception handling (or lack thereof).

dawn/prototype/AtlasFromNetcdf.cpp
havogt: No, need to change, but it seems that just calling resize would also fill with `missing_value()`?

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,275 @@
...
+      yOfX[innerIdx] = xToY[innerIdx * numY + elemIdx] - 1;
+    }
+    connectivity.set(elemIdx, yOfX);
+  }
+  return true;
+}
+
+// Pre-allocate neighborhood table, snippet taken from
+// https://github.com/ecmwf/atlas/blob/98ff1f20dc883ba2dfd7658196622b9bc6d6fceb/src/atlas/mesh/actions/BuildEdges.cc#L73
+void AllocNbhTable(atlas::mesh::HybridElements::Connectivity& connectivity, int numElements,
```
havogt: I missed it in the first review, but templating here also would remove the code duplication.

dawn/prototype/AtlasFromNetcdf.cpp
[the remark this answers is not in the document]
havogt: I'd propose to add a comment saying that, and maybe also replace `my` with something else, and `Part` -> `Partition` to make it obvious what it is. The `my` sounds like a placeholder as in "myLib".

dawn/prototype/AtlasFromNetcdf.cpp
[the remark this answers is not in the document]
havogt: I wouldn't do that. That would be a narrowing conversion which has other problems, even harder to spot.

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,275 @@
...
+    std::cout << "lat / long not of consistent sizes!\n";
+    return false;
+  }
+
+  int numNodes = lat.size();
+
+  // define nodes and associated properties for Atlas meshs
+  mesh.nodes().resize(numNodes);
+  atlas::mesh::Nodes& nodes = mesh.nodes();
+  auto xy = atlas::array::make_view<double, 2>(nodes.xy());
```
havogt: unused after your last changes

dawn/prototype/AtlasFromNetcdf.cpp
mroethlin: yes, however, the `add` functions also resizes some internal structures and fields. I'll leave as is because I'm afraid to break something down the line.

dawn/prototype/AtlasFromNetcdf.cpp
```diff
@@ -0,0 +1,273 @@
...
+    const int cellsPerEdge = 2;
+    const int cellsPerNode = 6; // maximum is 6, some with 5 exist
+    const int edgesPerNode = 6; // maximum is 6, some with 5 exist
+    const int edgesPerCell = 3;
+
+    // Allocate & fill neighbor tables from file
+    // ------------------------------
+
+    // Edges
+    AllocNbhTable<atlas::mesh::HybridElements::Connectivity>(mesh.edges().cell_connectivity(),
```
havogt: Why do you need to explicitly specify the template argument? What type is `mesh.edges().cell_connectivity()`?

mroethlin: I realize I don't, I was simply not used to doing it this way, thanks.

mroethlin:
Continued on external repo https://github.com/mroethlin/AtlasUtilities

PR #726, opened 2020-02-20
Fix direct reference to std::filesystem

jdahm:


jdahm:
launch jenkins

twicki:
launch jenkins

PR #729, opened 2020-02-21
Add default logger

jdahm:
## Technical Description

Add default logger. Explains cases and adds logging where we inexplicably did not have it before.


dawn/src/dawn/Support/Logging.cpp
```diff
@@ -29,7 +33,8 @@ internal::LoggerProxy::~LoggerProxy() {
 
 Logger* Logger::instance_ = nullptr;
 
-Logger::Logger() : logger_(nullptr) {}
+Logger::Logger() { registerLogger(new DefaultLogger); }
```
jdahm: What do people think of this small bounded memory leak? We could patch this up by changing the interface from `registerLogger` to replace the logger and call delete. The problem is if registering more than one logger, defaulting to always deleting could cause double deallocation.

Conclusion: It's better to not own the memory, but in that case we need to change more of the interface to get a default logger.

jdahm: My position is that a rewrite of the logging interface should be another task.

jdahm:
launch jenkins

dawn/src/dawn/Support/Logging.cpp
jdahm: I fixed the memory leak for the default logger. Now it only deallocates in the destructor if it's the default logger. If you register another logger, then you're responsible for deallocating.

jdahm:
launch jenkins

Issue #702, opened 2020-02-12
Integration / Unit test audit


eddie-c-davis:
Audit tests to determine redundancy and document in the [Wiki](https://github.com/MeteoSwiss-APN/dawn/wiki/Testing) page. A meeting will be scheduled to discuss this issue.

PR #731, opened 2020-02-21
Fix field acccess interval tests

twicki:
## Technical Description

Rework the testing framework. Changes the field_access_interval test to only use IIR and not go through SIR.

### Resolves / Enhances

Work on https://github.com/MeteoSwiss-APN/dawn/issues/702


twicki:
launch jenkins

PR #739, opened 2020-02-25
Update Deserialization

twicki:
## Technical Description

Update the deserializer to set the ID generator to the right place that Passes do not fail.

### Resolves / Enhances

Broken passes after deserialisation


dawn/src/dawn/Serialization/IIRSerializer.cpp
```diff
@@ -152,6 +154,12 @@ static void computeInitialDerivedInfo(const std::shared_ptr<iir::StencilInstanti
   }
 }
 
+static void checkAndUpdateMaxID(int newID, int& maxID) {
```
jdahm: Do we need a function here? Is this just
```
#include <cmath>
#include <algorithm>
maxID = std::max(maxID, std::abs(newID));
```

twicki: do you want to have it as that one liner everywhere or do you want it wrapped in a function anyway?

jdahm:
Other than the question above, this looks good!

twicki:
launch jenkins

PR #746, opened 2020-02-26
Fix stage reorder tests

eddie-c-davis:
## Technical Description

Refactors the stage reordering pass tests to only use IIR and not go through SIR.

### Resolves / Enhances

Work on #702


eddie-c-davis:
launch jenkins

dawn/test/unit-test/dawn/Optimizer/Passes/samples/ReorderTest02.cpp
```diff
@@ -0,0 +1,17 @@
+#include "gtclang_dsl_defs/gtclang_dsl.hpp"
+using namespace gtclang::dsl;
+
+stencil Test {
+  storage field_a0, field_a1, field_a2;
+  storage field_b0, field_b1, field_b2;
+
+  Do {
+    vertical_region(k_end, k_start) field_b1 = field_b0;
```
twicki: can we add `{}` here? I know that this notation technically works but the other one is cleaner in my eyes

eddie-c-davis: Done!

eddie-c-davis:
launch jenkins

Issue #747, opened 2020-02-26
No backend specific code (data_stores) in generated code API


havogt:
Currently we select the GT backend in a `driver-includes` header. We hard-code the GridTools data_stores which can be passed to generated code. This affects all backends as the data_stores appear in the API of the generated code.

A possible solution would be to base the API of the generated code on the SID concept of GridTools which would allow to pass any multi-dimensional array to the generated code (by providing SID-adapters).

havogt:
This would enable to remove the workaround for backend selection introduced in #748.

PR #748, opened 2020-02-26
Gt4py: workaround backend selection

havogt:
## Technical Description

Enables gt4py (and other consumers of generated GridTools code) to select the backend. This is a hacky, but a proper solution is out-of-scope of this task. See comment in code and issue #747.


havogt:
launch jenkins

PR #756, opened 2020-02-27
Expose computed field extents in generated code

### LINKED ISSUE #710: Expose the computed halo to be consumed by the frontend

havogt:
## Technical Description

Expose computed field extents in generated code as tuple-like objects. Additional named functions for accessing the elements could be added.

### Resolves / Enhances
Fixes #710.


havogt:
launch jenkins

havogt:
launch jenkins

dawn/src/dawn/CodeGen/CodeGen.cpp
```diff
@@ -444,5 +445,42 @@ void CodeGen::generateGlobalIndices(const iir::Stencil& stencil, Structure& sten
   }
 }
 
+namespace {
+std::string extent_to_string(iir::Extents const& extents, ast::GridType const& gridType) {
```
twicki:
```suggestion
std::string extentToString(iir::Extents const& extents, ast::GridType const& gridType) {
```

dawn/src/dawn/CodeGen/CodeGen.cpp
```diff
@@ -444,5 +445,42 @@ void CodeGen::generateGlobalIndices(const iir::Stencil& stencil, Structure& sten
   }
 }
 
+namespace {
```
twicki: https://github.com/MeteoSwiss-APN/dawn/blob/4acd78b812d3d2303a9ac5308bc2ee957013d173/dawn/src/dawn/IIR/Extents.cpp#L371-L389

exists. Why do we need another one?

havogt: I first extended this to also capture my use-case (by adding parameters for which kind of bracket to use etc. But then there is another difference: the generic one doesn't know the GridType and therefore cannot properly cast a no-extent. So in the end I decided to go with the specialized one.

havogt: The other one is actually only in error messages. Not sure if we want to keep it.

havogt:
launch jenkins

PR #762, opened 2020-02-27
Fix temporary merge tests

eddie-c-davis:
## Technical Description

Refactor the temporary field merging tests to only use IIR and not go through SIR.

### Resolves / Enhances

Applies to #702


eddie-c-davis:
launch jenkins

Issue #768, opened 2020-02-27
rework dawn unit test: dawn/CodeGen/Naive


twicki:
should be either iir file or iir-builder. If builder, how do we run passes?

comparison to string is ok as is.

PR #793, opened 2020-02-28
Refactor CodeGen tests using IIRBuilder and add missing CUDA tests

### LINKED ISSUE #767: rework dawn unit test: dawn/CodeGen/Cuda
should be either iir file or  iir-builder. If builder, how do we run passes? 

comparison to string is ok as is.

eddie-c-davis:
## Technical Description

Factor common code tests for both from Naive and CUDA backends into a single class and adds missing tests for CUDA backend.

### Resolves / Enhances

Resolves #767 and #768


eddie-c-davis:
launch jenkins

Issue #592, opened 2020-01-14
Stage location type is not set/updated


Stagno:
- [x] Fix in parallel group #797 
- [x] Fix stage merger #831 
- [x] Check if other passes (PassStencilSplitter?, PassIntervalPartitioning?) need fixing
- [ ] Add checks to PassValidation

twicki:
conclusion:

- We use `optional` to implement the [Blackboard pattern](https://en.wikipedia.org/wiki/Blackboard_(design_pattern)). 
- This could also work for stage location type.
- If we opt to go for this route, IIR-completeness can be checked by checking if all of them are set
- Unclear if the stage-splitter does the correct thing @Stagno ?
- Unclear if this is set ever going from SIR to IIR and if our tests didn't catch it @Stagno ?

Stagno:
I looked at the usages of `Stage::setLocationType()`: it's only used by `IIRBuilder`. This means:
- Currently stage splitting does not fix the location type of the stage (the default value = cells is kept). I.e.: we only added the possibility to set a location type to the stage. The logic to set/update such information is yet to be implemented (that's why I created this task).
- Going from SIR to IIR we don't set the location type (again the default is kept). We don't have tests that check the Stage's location type.

Note on usage of `std::optional` to represent information which could be "not yet computed": even if the name optional suggests another thing, this class can be used to serve exactly the purpose of representing an information which is not there yet. We could still hide it behind an API.

Stagno:
DoD:
- make `Stage::type_` an `optional` (or whatever way to represent an information which could be not yet computed).
- fix stage splitter to correctly set the location type of the stages
- fix any code touching a Stage to keep this information up to date (not sure if there's any)
- write tests to check whether statements inside a Stage match the Stage's location type specification

PR #797, opened 2020-03-02
Split stage by LocationType

havogt:
## Technical Description

Adds passes:
- split each statement into a separate stage
- assign a location type to each stage

### Additional changes:
- Refactored `Stage.split()`.
- Remove `0` values in protobuf enums, because 0 is the implicit default.

### Resolves / Enhances

partially #592; missing: ensure other passes don't mess up with the location type (e.g. stage merger).

### Implementation details:
- Assumes that we don't have non-sense expressions (expressions without effect on global state), e.g. expressions which are not assigned and have no side-effect.


havogt:
launch jenkins

havogt:
launch jenkins

dawn/test/unit-test/dawn-c/TestCompiler.cpp
```diff
@@ -38,7 +38,9 @@ static void freeCharArray(char** array, int size) {
   std::free(array);
 }
 
-TEST(CompilerTest, CompileEmptySIR) {
+// TODO empty is not legal anymore as gridtype is required (the default is illegal)
```
havogt: @review Not sure what to do here. Delete the test? Or change it to a minimal SIR with only required info.

The reason this doesn't work anymore is that we removed default values in protobuf enums.

dawn/test/unit-test/dawn/Optimizer/Passes/samples/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+
+    add_custom_target(${ARG_NAME}
+        COMMAND ${CMAKE_COMMAND} -E env
+        PYTHONPATH=${DAWN4PY_MODULE_DIR}:${PROTOBUF_PYTHON_DIR}
+        ${Python3_EXECUTABLE} ${ARG_SCRIPT}
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    )
+endfunction()
+
+add_dawn4py_target(NAME GenerateSetLocationTypeInput SCRIPT SetLocationType.py)
```
havogt: @ review: this is not part of the all target, it generates the sir input for our tests. we consider this temporary until we have a well-defined way to regenerate input.

dawn/src/dawn/Unittest/CompilerUtil.cpp
```diff
@@ -155,6 +157,9 @@ CompilerUtil::createGroup(PassGroup group, std::unique_ptr<OptimizerContext>& co
     addPass<dawn::PassMultiStageSplitter>(context, passes, mssSplitStrategy);
     addPass<dawn::PassStageSplitter>(context, passes);
     addPass<dawn::PassTemporaryType>(context, passes);
+    addPass<dawn::PassLocalVarType>(context, passes);
+    addPass<dawn::PassRemoveScalars>(context, passes);
+    // TODO: how to add grid-dependent passes?: PassStageSplitAllStatements, PassSetLocationType
```
Stagno: Still not clear what is the intended way to use this method (it's not used yet).
Does it aim at replacing also the code doing the same thing in DawnCompiler.cpp?

havogt:
launch jenkins

Stagno:
launch jenkins

dawn/test/unit-test/dawn-c/TestCompiler.cpp
twicki: Does compiling an SIR with no Stencils break things? That would be my initial approach here

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
```diff
@@ -0,0 +1,47 @@
...
+//===------------------------------------------------------------------------------------------===//
+
+#pragma once
+
+#include "Pass.h"
+
+namespace dawn {
+
+/// @brief
+/// * Input: any IIR where GridType == Unstructured and where statements inside the same stage are
```
twicki: it actually also requires every statement to be in it's own stage, see the assert in line 91

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
```diff
@@ -0,0 +1,47 @@
...
+/// on the same location type (local variables must have a location type and no scalar variables are
+/// allowed).
+/// * Output: same IIR, but each stage will have it's location type correctly set (!= std::nullopt)
+/// to the location type of its statements
+/// @ingroup optimizer
+///
+/// This pass is necessary to generate legal IIR for GridType == Unstructured.
```
twicki: can this not also set it to cells trivially for structured grids?

Then this pass could run for both and we would not have an additional branch of flow?

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -168,7 +171,12 @@ DawnCompiler::lowerToIIR(const std::shared_ptr<SIR>& stencilIR) {
   optimizer.pushBackPass<PassTemporaryType>();
   optimizer.pushBackPass<PassLocalVarType>();
   optimizer.pushBackPass<PassRemoveScalars>();
-  optimizer.pushBackPass<PassStageSplitter>();
+  if(stencilIR->GridType == ast::GridType::Unstructured) {
```
twicki: I know you mention in the description that you don't want to be smart about the splitting in the lowering and I agree very much. 
I have an implementation of the parallel model coming that does exactly this (1 stage per stmt) at the lowering level. But for now I would say that `PassStageSplitAllStatements` would be beneficial for both models as it is a step towards implementing our execution model. (MS loop is still to be done). So I would remove the if here and use it for both

dawn/src/dawn/IIR/Stage.cpp
```diff
@@ -280,47 +282,63 @@ void Stage::appendDoMethod(DoMethodSmartPtr_t& from, DoMethodSmartPtr_t& to,
                            std::make_move_iterator(from->getAST().getStatements().end()));
 }
 
-std::vector<std::unique_ptr<Stage>> Stage::split(std::deque<int>& splitterIndices,
-                                                 std::deque<DependencyGraphAccesses>* graphs) {
+namespace {
```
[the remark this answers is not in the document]
havogt: Our style guide seems to mandate static, it would have been helpful if you would have just said that, because now I had to search for it...
My first response would have been, why static function over unnamed namespace. ;)

dawn/test/unit-test/dawn/Optimizer/TestPassSetStageLocationType.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+TEST_F(TestPassSetStageLocationType, CopyFieldsLocationTypes) {
+  // field(cells) in_cell, out_cell;
+  // field(edges) in_edge, out_edge;
+  // fields(vertices) in_vertex, out_vertex
+  // out_cell = in_cell;
+  // out_edge = in_edge;
+  // out_vertex = in_vertex;
+
+  runPass("input/test_set_stage_location_type_copy_fields.sir");
```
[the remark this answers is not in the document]
havogt: Yes, there are 2 problems:
- The last test cannot be made on IIR as we don't serialize stencil functions.
- We currently don't have the infrastructure to generate this from python, but probably we should have it, by exposing something like `CompilerUtil` to python. We thought this should be discussed with the team and considered it out-of-scope for this PR.

We considered to work around the second problem, but then decided to not do it, because of the blocker for stencil functions.

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
havogt: @Stagno What do you think, should we make this comment more restrictive or delete the assert?
@twicki The reason for phrasing it like this was that strictly, the single statement per stage is not required (except for the assert) as long as each stage has only statements on the same location type.

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
havogt: It could, however I'd prefer clarity over not-branching. Running a path which does essentially nothing seems confusing to me.

dawn/src/dawn/Compiler/DawnCompiler.cpp
havogt: Perfect, we didn't dare to touch it, but since you will anyway do it, we can do it already here. However, if performance changes, I think it's not in the scope of this PR to fix it.

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
Stagno: I'm for deleting the assert

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
Stagno: The location type of a stage is deduced from the lhs of the first (useful) statement. Querying the location type of a cartesian field doesn't make sense and it's also not allowed with our current data structures (`sir::FieldDimensions`).
Changing the deduction to consider cartesian as special case is again branching (hidden inside the pass). Changing the data structures to fit this case seems a bad option.

dawn/test/unit-test/dawn-c/TestCompiler.cpp
Stagno: SIR with no stencils should work. Now the question is: do we still start from an sir string? It seems too verbose to write the string in code, so maybe deserializing sir from file is the way?

havogt: I tried to make it work until I realized the `dawnCompile` expects a byte encoded protobuf message. Not sure if it makes sense to embed byte data.

dawn/src/dawn/IIR/Stage.cpp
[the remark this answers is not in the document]
havogt: done

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
havogt: deleted the assert

dawn/src/dawn/Compiler/DawnCompiler.cpp
havogt: I would say, we merge this PR with this if, then I'll open a new PR, where it is removed and we can investigate if things break. Ok?

havogt: 2 tests in dawn fail actually, so I will not remove the if in this PR.

havogt:
launch jenkins

dawn/test/unit-test/dawn-c/TestCompiler.cpp
twicki: understood. So should we drop it?

dawn/test/unit-test/dawn/Optimizer/TestPassSetStageLocationType.cpp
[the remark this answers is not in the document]
twicki: got it. I can live this this for now. Can we create an issue to remember to change this once we have the stencil function serialisation?

dawn/src/dawn/Compiler/DawnCompiler.cpp
twicki: sounds ok to me

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
twicki: The advantage of running it also for structured would be that you have a clear contract on how the IIR would look after the pass (that every stage has a location type. Then we could consider checking for this property after that (maybe after the parallel group) which would not work when we're not doing it for structured grids.

twicki: and I don't understand why cartesian does not just have cells as the location type? Then deduction works the same way.
Can you elaborate on what would need to change?

jdahm: LocationTypes seem like a fundamental feature that should be in both structured and unstructured grids. I was concerned from the get-go when this wasn’t in the structured grids. As you say it could just be set to cells.

Stagno: @twicki we already discussed it (also with @cosunae ) when we were working on https://github.com/MeteoSwiss-APN/dawn/commit/ccfc9818965cdbe540fd4a092a69089e10ecb8db and we needed to refactor the field dimensions. The decision was to split the horizontal part in cartesian (IJ masks) and unstructured (dense location type + sparse). 
To change it back, we would need to change the protobuf (and consequently all serialized sir/iir files), the field dimensions classes and their APIs (and their usages). Also, local variable types (OnCells, OnEdges, ..., OnIJ) were modeled following this decision and would imply more changes.

Not taking into consideration the amount of work needed to change, my opinion is that we shouldn't share a concept which is not needed for one of the gridtypes. 
Also, I'd say having few gridtype-specific passes might be inevitable (I'm thinking about, for example, optimizing away a sparse field by inlining it into a reduction where it's used), while still remains that the vast majority of passes are shared.
I agree that branching is not ideal, but I'm confident that, once we'll have enough integration/end-to-end test also for unstructured, it won't cause us problems.

dawn/test/unit-test/dawn/Optimizer/TestPassSetStageLocationType.cpp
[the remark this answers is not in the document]
havogt: Created https://github.com/MeteoSwiss-APN/dawn/issues/845

dawn/test/unit-test/dawn-c/TestCompiler.cpp
havogt: I deleted it.

havogt:
launch jenkins

dawn/src/dawn/Optimizer/PassSetStageLocationType.h
twicki: I see, thanks for the explanation

PR #805, opened 2020-03-03
Rework From SIR Tests

### LINKED ISSUE #778: rework dawn unit test: dawn/Optimizer/TestsFromSIR
These are the prime example of integration tests. Move them there and assert on properties of the iir after some passes (default group) are ran

mroethlin:
## Technical Description

This PR moved the "From SIR" tests formerly located in the unit tests to the integration test folder (since they are integration tests). Instead of running the whole set of passes, now only a minimal set of passes is run to ensure compliance with the references tested against. 

## NOTE

I had to run an additional group before the StageMergerPass (StageReordering) in order to avoid segfaults. This may indicate a bug in the StageMergerPass or an unsersolved dependency between those two passes. 

### Resolves / Enhances

Fixes #778

### Dependencies

This PR is independent


mroethlin:
launch jenkins

twicki:
super interesting that you had to run something. I could run parallel and stage stage merging on all of clang-gritools. Unclear to me if the compiler-util is up to date @eddie-c-davis?

eddie-c-davis:
Looks like a bug in CompilerUtil, the SetDependencyGraph pass is missing. I will get a fix submitted.

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #814, opened 2020-03-04
Correctly install IIR and SIR generated proto headers

jdahm:
## Technical Description

The proto header files were still not being correctly installed. This PR fixes that.


jdahm:
launch jenkins

PR #816, opened 2020-03-06
Rework optimizer passes

### LINKED ISSUE #769: rework dawn unit test: dawn/Optimizer/Passes
Generally this folder is not reflecting the proper structure. unit-test should be equivalent to root directory and files should have the same name with a prefix `Test` to signal which file they test. Change the structure to reflect this.

In this folder there are also two tests that are not testing passes: TestComputeMaxExtent & TestFieldAccessIntervals. These should be moved to the right places

eddie-c-davis:
## Technical Description

Restructures Optimizer pass test to match source directory, combines stage extent tests, and moves field access unit tests into the IIR module.

### Resolves / Enhances

Resolves #769


eddie-c-davis:
launch jenkins

PR #817, opened 2020-03-07
Add -default-none option to gtclang

jdahm:
## Technical Description

Add back the `-default-none` option to gtclang.


dawn/src/dawn/Optimizer/Options.inc
```diff
@@ -47,6 +47,8 @@ OPT(int, BlockSizeI, 0, "block-size-i", "", "i block size for tiled computations
 OPT(int, BlockSizeJ, 0, "block-size-j", "", "j block size for tiled computations", "", true, false)
 OPT(int, BlockSizeK, 0, "block-size-k", "", "k block size for tiled computations", "", true, false)
 
+OPT(bool, DefaultNone, false, "default-none", "",
```
twicki: is default none a good name? would disable-optimisation be better?

jdahm: Sure. Renamed!

jdahm:
launch jenkins

PR #820, opened 2020-03-10
Quick fix for nested ifs

Stagno:
## Technical Description

PassLocalVarType wasn't looking inside the conditions of nested if statements to register variable dependencies in order to guess their types. This fixes it.

### Testing

As I'm not yet sure how to transform the unit tests of this pass to start from IIR and have a way to regenerate the input IIR, the 4 new tests of this PR follow the same pattern used before. Once the procedure is clarified, they will be all converted at once.


Stagno:
launch jenkins

PR #834, opened 2020-03-11
Fix Dawn4Py SIR serialization utils to work with GTPy

eddie-c-davis:
## Technical Description

This PR resolves the following GT4Py errors in the SIR serialization step in the Dawn4Py integration that occur in the FV3 stencils:

```python
  File "/home/eddied/Work/dawn/dawn/src/dawn4py/serialization/utils.py", line 357, in make_stmt
    raise SIRError("cannot create Stmt from type {}".format(type(stmt)))
dawn4py.serialization.error.SIRError: cannot create Stmt from type <class 'SIR.statements_pb2.Field'>
```

... and ...

```python
File "/home/eddied/Work/dawn/dawn/src/dawn4py/serialization/utils.py", line 722, in make_literal_access_expr
  builtin_type.type_id = type
  AttributeError: 'BuiltinType' object attribute 'type_id' is read-only
```

### DIFF
--- a/dawn/src/dawn4py/serialization/utils.py
+++ b/dawn/src/dawn4py/serialization/utils.py
@@ -365,7 +365,7 @@ def make_block_stmt(statements: List[StmtType]) -> BlockStmt:
     """
     stmt = BlockStmt()
     if isinstance(statements, Iterable):
-        stmt.statements.extend([make_stmt(s) for s in statements])
+        stmt.statements.extend([make_stmt(s) for s in statements if not isinstance(s, Field)])
     else:
         stmt.statements.extend([make_stmt(statements)])
     return stmt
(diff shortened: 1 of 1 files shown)


eddie-c-davis:
launch jenkins

dawn/src/dawn4py/serialization/utils.py
```diff
@@ -718,7 +718,8 @@ def make_literal_access_expr(value: str, type: BuiltinType.TypeID) -> LiteralAcc
     :param type:    Builtin type id of the literal.
     """
     builtin_type = BuiltinType()
-    builtin_type.type_id = type
+    if builtin_type.type_id != type:
```
jdahm: Is this change necessary?

eddie-c-davis: Yes, it throws the second error in the description otherwise.

jdahm: Ah ok, an error gets raised in the = operation otherwise. Thanks for clarifying!

PR #848, opened 2020-03-14
Fix gtclang dawn call

jdahm:
## Technical Description

Calls dawn components separately from gtclang, in order to recognize the `-disable-optimization` gtclang flag, since calling separately does not add default pass groups when none are present.

Also deletes a bit of unnecessary code.


jdahm:
launch jenkins

jdahm:
I checked and `gtclang --disable-optimization --verbose <input>` shows that only the lowering group, in addition to the couple passes in `restoreIIR` get run.

PR #849, opened 2020-03-14
Fix globals serialization

### LINKED ISSUE #724: Global variables not properly deserialized from IIR
This issue occurs when deserializing the IIR from the ```conditional_stencil``` test in ```gtclang/integration-tests```. The specified globals are:

```c++
globals {
  int var1 = 1;
  bool var2;
};
```

These are serialized to IIR as:
```javascript
"globalVariableToValue": {
   "var1": {
    "type": "Integer",
    "value": 1,
    "valueIsSet": true
   },
   "var2": {
    "type": "Boolean",
    "value": 0,
    "valueIsSet": false
   }
  ```

However, the code generator produces the following ```globals``` struct:
```c++
struct globals {
  double var2;
  int var1;
  globals() : var2(0.000000) {}
};
```

Instead, this should be:
```c++
struct globals {
  bool var2;
  int var1;
  globals() : var1(1) {}
};
```

1. The type for ```var2``` is ```Double``` but should be ```Boolean```.
2. The value for ```var1``` is undefined but should be 1.

eddie-c-davis:
## Technical Description

This PR resolves the bug in the IIRSerializer class that caused global variables to be improperly serialized.

### Resolves / Enhances

Resolves #724

### DIFF
--- a/dawn/src/dawn/Serialization/IIRSerializer.cpp
+++ b/dawn/src/dawn/Serialization/IIRSerializer.cpp
@@ -622,25 +622,25 @@ void IIRSerializer::deserializeIIR(std::shared_ptr<iir::StencilInstantiation>& t
     switch(GlobalToValue.second.type()) {
     case proto::iir::GlobalValueAndType_TypeKind_Boolean:
       if(GlobalToValue.second.valueisset()) {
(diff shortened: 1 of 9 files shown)


eddie-c-davis:
launch jenkins

PR #860, opened 2020-03-17
Change global variable map from unordered to ordered

eddie-c-davis:
## Technical Description

This PR changes the SIR global variables map from unordered to ordered to ensure that the same backend code is always generated.

### DIFF
--- a/dawn/src/dawn/SIR/SIR.h
+++ b/dawn/src/dawn/SIR/SIR.h
@@ -559,6 +559,7 @@ public:
   Global(Kind type) : Value(type) {}
 };
 
+// Using ordered map to guarantee the same backend code will be generated
 using GlobalVariableMap = std::map<std::string, Global>;
 
 } // namespace sir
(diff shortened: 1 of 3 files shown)


eddie-c-davis:
launch jenkins

dawn/src/dawn/SIR/SIR.h
```diff
@@ -559,7 +559,7 @@ class Global : public Value, NonCopyable {
   Global(Kind type) : Value(type) {}
 };
 
-using GlobalVariableMap = std::unordered_map<std::string, Global>;
+using GlobalVariableMap = std::map<std::string, Global>;
```
jdahm: Do you think it's appropriate to add a comment here that this should really be an unordered_map but we're changing it to ordered to fix the codegen tests? If so, please add!

eddie-c-davis: Sure, I can add a comment.

eddie-c-davis: Added!

eddie-c-davis:
launch jenkins

PR #862, opened 2020-03-18
Try-catch correctly with dawn-opt

jdahm:
## Technical Description

Correctly try-catch with dawn-opt. Involves removing incorrect `dawn_unreachable(...)` calls and replacing these with throws. The serialization functions should no longer use `dawn_unreachable(...)` in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,3 @@ _local_/*
 
 # Ignore python editable installs
 *.egg-info/
-
-# Python venvs
-.venv/
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ _local_/*
 
 # Ignore python editable installs
 *.egg-info/
+
+# Python venvs
+.venv/
--- a/dawn/src/dawn/Compiler/DawnCompiler.cpp
+++ b/dawn/src/dawn/Compiler/DawnCompiler.cpp
@@ -195,10 +195,6 @@ DawnCompiler::lowerToIIR(const std::shared_ptr<SIR>& stencilIR) {
       throw std::runtime_error("An error occurred.");
 
     DAWN_LOG(INFO) << "Done with parallelization passes for `" << instantiation->getName() << "`";
-
-    if(options_.DumpStencilInstantiation) {
-      instantiation->dump();
-    }
   }
 
   return optimizer.getStencilInstantiationMap();
--- a/dawn/src/dawn/Serialization/IIRSerializer.cpp
+++ b/dawn/src/dawn/Serialization/IIRSerializer.cpp
@@ -29,6 +29,7 @@
 #include <google/protobuf/util/json_util.h>
 #include <memory>
 #include <optional>
+#include <stdexcept>
 
 namespace dawn {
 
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -24,3 +24,6 @@ _local_/*
 
 # Ignore python editable installs
 *.egg-info/
+
+# Python venvs
+.venv/
```
twicki: why would you set up your venv hidden in the root directory?

dawn/src/dawn/Serialization/IIRSerializer.cpp
```diff
@@ -74,8 +75,6 @@ static void setCache(proto::iir::Cache* protoCache, const iir::Cache& cache) {
   case iir::Cache::IOPolicy::unknown:
     protoCache->set_policy(proto::iir::Cache_CachePolicy_CP_Unknown);
     break;
-  default:
-    dawn_unreachable("unknown cache policy");
```
twicki: why are some of them getting removed while others are replaced?

jdahm: Because there's no default option needed here. All cases are accounted for in the switch. The ones that I kept the default on are almost all proto:: enumerations that have a hidden `invalid` option too (at least clang told me so).

.gitignore
jdahm: I thought it may be useful to allow for a venv for dawn (that's how I develop for dawn). If you don't agree or think it should be in a personal global .gitignore file, then I can remove from here.

twicki: I mean I also use a venv for dawn. But i feel like the hidden folder in the source directory is an odd choice and would therefore not put it here

dawn/src/dawn/Serialization/IIRSerializer.cpp
twicki: having this default future proofs this though (adding a new option and forgetting it here would prompt an error instead of silently ignoring it, right?

.gitignore
jdahm: I agree now that this should be in a users global .gitignore, so I removed it!

dawn/src/dawn/Serialization/IIRSerializer.cpp
jdahm: Decided to leave this as is, so there will be a compiler warning/error if new enum values are added and new cases are not added here.

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -195,10 +195,6 @@ DawnCompiler::lowerToIIR(const std::shared_ptr<SIR>& stencilIR) {
       throw std::runtime_error("An error occurred.");
 
     DAWN_LOG(INFO) << "Done with parallelization passes for `" << instantiation->getName() << "`";
-
-    if(options_.DumpStencilInstantiation) {
```
jdahm: This was being printed twice to screen. Oops!

jdahm:
launch jenkins

PR #863, opened 2020-03-18
Fixing flag usage in gtclang

twicki:
## Technical Description

This ensures the use of `-no-opt` and `-default-opt` do the correct thing. 
The new interface does:
- run the default group if nothing is specified
- run only the selected passes if they are specified
- run nothing if `-no-opt` is specified
- run default plus passes if both are specified
- asserts if `no-opt` and optimization is specified

### Resolves / Enhances

Issues with current implementation of the flag check while maintaining backwards compatibility.

## Testing

This requires the default flag to be set in one of Clang-GT's tests, so testing is here:
http://localhost:8080/job/dawn_gtclang_clang-gridtools_PR/415/

### DIFF
--- a/dawn/test/integration-test/dawn4py-tests/data/copy_stencil_ref.cpp
+++ b/dawn/test/integration-test/dawn4py-tests/data/copy_stencil_ref.cpp
@@ -38,6 +38,10 @@
 #include <driver-includes/gridtools_includes.hpp>
 using namespace gridtools::dawn;
 
+//---- Includes ----
+#include "driver-includes/gridtools_includes.hpp"
+using namespace gridtools::dawn;
+
 //---- Globals ----
 
 //---- Stencils ----
--- a/gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp
+++ b/gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp
@@ -142,55 +142,69 @@ void GTClangASTConsumer::HandleTranslationUnit(clang::ASTContext& ASTContext) {
     return;
 
   // Compile the SIR using Dawn
-  std::list<dawn::PassGroup> PassGroups;
+  std::list<dawn::PassGroup> passGroup;
 
   if(context_->getOptions().SSA)
-    PassGroups.push_back(dawn::PassGroup::SSA);
+    passGroup.push_back(dawn::PassGroup::SSA);
 
   if(context_->getOptions().PrintStencilGraph)
-    PassGroups.push_back(dawn::PassGroup::PrintStencilGraph);
+    passGroup.push_back(dawn::PassGroup::PrintStencilGraph);
 
-  if(!context_->getOptions().DisableOptimization)
-    PassGroups.push_back(dawn::PassGroup::SetStageName);
+  if(context_->getOptions().SetStageName || context_->getOptions().DefaultOptimization)
+    passGroup.push_back(dawn::PassGroup::SetStageName);
 
-  if(!context_->getOptions().DisableOptimization)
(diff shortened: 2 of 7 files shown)


twicki:
launch jenkins

twicki:
launch jenkins

gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp
```diff
@@ -142,55 +142,69 @@ void GTClangASTConsumer::HandleTranslationUnit(clang::ASTContext& ASTContext) {
     return;
 
   // Compile the SIR using Dawn
-  std::list<dawn::PassGroup> PassGroups;
```
jdahm: I made this `PassGroups` because the convention elsewhere in GTClang was to begin variable names with a capital letter. I prefer the dawn approach because it distinguishes between variables and types

twicki: https://github.com/MeteoSwiss-APN/dawn/blob/80a61cb5c1488bf604175098e19f345292c493a8/gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp#L118

https://github.com/MeteoSwiss-APN/dawn/blob/80a61cb5c1488bf604175098e19f345292c493a8/gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp#L131

just a couple examples out of this file.

gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp
```diff
@@ -142,55 +142,69 @@ void GTClangASTConsumer::HandleTranslationUnit(clang::ASTContext& ASTContext) {
...
+  if(!context_->getOptions().DisableOptimization && passGroup.size() == 0) {
+    passGroup.push_back(dawn::PassGroup::SetStageName);
+    passGroup.push_back(dawn::PassGroup::StageReordering);
+    passGroup.push_back(dawn::PassGroup::StageMerger);
+    passGroup.push_back(dawn::PassGroup::SetCaches);
+    passGroup.push_back(dawn::PassGroup::SetBlockSize);
+  }
+
+  if(context_->getOptions().DisableOptimization && passGroup.size() > 0) {
+    DAWN_ASSERT_MSG(false,
```
jdahm: Can we `throw std::invalid_argument("Inconsistent arguments: no optimisation as well as optimization specified here")` instead of the assertion?

twicki: since this would be the first instance of this in the code I would say it does not belong into this PR.

And the catching mechanism for throwing here is not set up yet

gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp
jdahm: Oh, hmm then it must be inconsistent anyway. Good to make it more like dawn!

gtclang/src/gtclang/Frontend/GTClangASTConsumer.cpp
jdahm: Voted to delay making these exceptions to a future PR.

dawn/test/integration-test/dawn4py-tests/data/copy_stencil_ref.cpp
```diff
@@ -38,6 +38,10 @@
 #include <driver-includes/gridtools_includes.hpp>
 using namespace gridtools::dawn;
 
+//---- Includes ----
```
eddie-c-davis: Are we okay with the duplicated includes here?

PR #866, opened 2020-03-18
Fix global iteration space bug in CUDA backend

### LINKED ISSUE #865: Fix global iteration space bug in CUDA backend
The current implementation of global indices in the Dawn CUDA backend does not copy the iteration space indices and global offsets to device memory before launching the kernel.

eddie-c-davis:
## Technical Description

This PR modifies the CudaCodeGen backend to copy global indices and offsets to device memory before the kernel call and frees it afterward.

### Resolves / Enhances

Resolves #865

### DIFF
--- a/dawn/src/dawn/CodeGen/Cuda/MSCodeGen.cpp
+++ b/dawn/src/dawn/CodeGen/Cuda/MSCodeGen.cpp
@@ -701,6 +701,17 @@ void MSCodeGen::generateCudaKernelCode() {
   });
 
   if(iterationSpaceSet_) {
+    std::string iterators = "IJ";
+    for(const auto& stage : iterateIIROver<iir::Stage>(*(stencilInstantiation_->getIIR()))) {
+      for(auto [index, interval] : enumerate(stage->getIterationSpace())) {
+        if(interval.has_value()) {
+          ss_ << "__constant__ int stage" << stage->getStageID() << "Global" << iterators.at(index)
+              << "Indices_[2];\n";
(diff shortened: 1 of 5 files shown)


eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

dawn/src/dawn/CodeGen/Cuda/MSCodeGen.cpp
```diff
@@ -1059,15 +1055,13 @@ void MSCodeGen::generateCudaKernelCode() {
           std::string iterators = "IJ";
           for(const auto& stage : iterateIIROver<iir::Stage>(*(stencilInstantiation_->getIIR()))) {
             std::string prefix = "stage" + std::to_string(stage->getStageID()) + "Global";
-            int index = 0;
-            for(const auto& interval : stage->getIterationSpace()) {
+            for(auto [index, interval] : enumerate(stage->getIterationSpace())) {
```
twicki: I am torn if this is more or less readable than the standard `for(int i=0; i< size;++i)`
it is of course fancier and it's nice that c++ offers a way to get indices from range base loops. Is there an an advantage to this or do you think it's just the nicer way of doing things?

eddie-c-davis: Johann has encouraged me to use these modern iterators, I personally have no strong preference. I am happy to change to standard C++ loops if you like.

PR #876, opened 2020-03-25
Move and rename option for PassStencilSplitter

jdahm:
## Technical Description

Move and rename option for the Stencil Splitter (group). This does NOT re-enable the pass.

### DIFF
--- a/dawn/src/dawn/Optimizer/PassOptions.inc
+++ b/dawn/src/dawn/Optimizer/PassOptions.inc
@@ -61,6 +61,9 @@ OPT(bool, SetBlockSize, false, "set-block-size", "",
 OPT(bool, DataLocalityMetric, false, "data-locality-metric", "",
     "Run data-locality-metric pass group", "", false, true)
 
+OPT(bool, StencilSplitter, false, "stencil-splitter", "",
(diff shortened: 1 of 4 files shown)


jdahm:
launch jenkins

dawn/src/dawn/Optimizer/PassOptions.inc
```diff
@@ -61,6 +61,9 @@ OPT(bool, SetBlockSize, false, "set-block-size", "",
 OPT(bool, DataLocalityMetric, false, "data-locality-metric", "",
     "Run data-locality-metric pass group", "", false, true)
 
+OPT(bool, StencilSplitter, false, "stencil-splitter", "",
```
twicki: if we keep the option around,should we not at least have the pass present in the gtclangastconsumer?

jdahm: Sure. I wasn't sure if we should since it was disabled. This PR was just going to make it consistent with the others and move it into PassOptions.inc. We can definitely extend the PR and re-enable the pass.

jdahm:
Discussed this and decided to delay the work for now until this pass can be refactored.

PR #892, opened 2020-03-26
Update loop orders in IIR reference files

eddie-c-davis:
## Technical Description

Updates IIR reference files in the `GTClang::Integration` tests to reflect the change that the default multistage loop order is specified by the user.


eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

PR #896, opened 2020-03-27
Recompute horizontal extents

### LINKED ISSUE #893: Horizontal extents not serialized
Horizontal extents are computed in the `PassComputeStageExtents` pass and stored in `Stage.DerivedInfo.extents_`. The `derivedinfo` object is not serialized in the `IIRSerializer` so the information is lost.

eddie-c-davis:
## Technical Description

This PR moves the logic to compute the horizontal extents in each stencil instantiation from the `PassComputeStageExtents` pass to a `StencilInstantiation::computeDerivedInfo` method that is called from the `IIRDeserializer`. The `PassComputeStageExtents` is removed, and the `computeDerivedInfo' method is instead called from the passes that formerly depended on it.

### Resolves / Enhances

Resolves #893


eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

eddie-c-davis:
launch jenkins

dawn/src/dawn/Optimizer/PassComputeStageExtents.cpp
```diff
@@ -27,71 +27,7 @@ PassComputeStageExtents::PassComputeStageExtents(OptimizerContext& context)
...
-        }
-      }
-    }
-  }
-
-  for(const auto& MS : iterateIIROver<iir::MultiStage>(*(stencilInstantiation->getIIR()))) {
-    MS->update(iir::NodeUpdateType::levelAndTreeAbove);
-  }
-
+  stencilInstantiation->computeStageExtents();
```
twicki: I think this is a dangerous thing to do. 

I assume the thought process is to not change any of the dependency set up, right?

But now we have a pass that can be called outside the pass framework which I don't really like.

I would prefer if the `Serializer` has a dependency on the Optimizer or if the IIR has an option `recomputeAllDerivedInfo` that you call when it is needed and remove the Pass. Option 2 could work with the StencilInstantiaion having the `computeStageExtents` call (or you could even push it down to the stage level, so every stage computes for itself).  Which is then initially called where we have the pass now and than in the recompute phase

dawn/test/unit-test/dawn/CodeGen/TestCodeGen.h
```diff
@@ -12,6 +12,8 @@
 //
 //===------------------------------------------------------------------------------------------===//
 
+//#include "dawn/Optimizer/OptimizerContext.h"
```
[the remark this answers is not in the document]
eddie-c-davis: Removed.

dawn/test/unit-test/dawn/CodeGen/TestCodeGen.h
```diff
@@ -22,13 +24,14 @@
 
 namespace {
 
+using namespace dawn;
```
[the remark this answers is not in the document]
eddie-c-davis: Changed

dawn/src/dawn/Optimizer/PassComputeStageExtents.cpp
eddie-c-davis: Went with Option 2:
- Renamed `computeStageExtents` to `computeDerivedInfo`
- Removed `PassComputeStageExtents` and replaced all occurences

eddie-c-davis:
launch jenkins

dawn/src/dawn/Optimizer/PassSetBoundaryCondition.cpp
```diff
@@ -133,6 +128,10 @@ bool PassSetBoundaryCondition::run(
     }
     return true;
   }
+
+  iir::StencilMetaInformation& metadata = stencilInstantiation->getMetaData();
```
twicki: this should not be necessary. It was a requirement to have happened before

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -277,9 +275,7 @@ DawnCompiler::optimize(const std::map<std::string, std::shared_ptr<iir::StencilI
       optimizer.pushBackPass<PassTemporaryType>();
       optimizer.pushBackPass<PassLocalVarType>();
       optimizer.pushBackPass<PassRemoveScalars>();
-      // modify stages and their extents ...
-      optimizer.pushBackPass<PassComputeStageExtents>();
```
twicki: should we make sure that this function is now called after `PassRemoveScalars` and `PassFixVersionedInputFields`?

eddie-c-davis: It is called in `PassSetSyncStage.run`. What type of check did you have in mind?

dawn/src/dawn/Optimizer/PassSetBoundaryCondition.cpp
eddie-c-davis: Afraid I do not understand, a test that no longer exists should not be a dependency, correct?

PR #897, opened 2020-03-30
Generalize Weights

### LINKED ISSUE #870: Generalize Weights
Currently, the weights consumed in the reduction are compile time literals. There are use cases in the ICON code that require weights to be more general. For this task, lift the requirements on the weights being compile time literals and allow them to be any arithmetic function of `FieldAccessExpr`. 

DoD: introduce a test using more general weights and ensure it passes

mroethlin:
## Technical Description

Until now, weights (in the reduce over neighbors concept) were restricted to be compile time literals. This PR lifts this restriction somewhat and allows the weights to be any arithmetic combination of `FieldAccessExpr`s into dense fields. Accesses into sparse fields or even more complicated constructs like allowing weights to be the result of `ReduceOverNeighborExpr`s are still prohibited. The reason for this is the current interface to the meshing libraries, which would need significant refactoring to enable these constructs. A new validation pass ensuring that the weights conform to these restrictions was introduced.

### Resolves / Enhances

Fixes #870 

### Testing

There are number of new tests:
- Unit tests to ensure that the new validation pass is working correctly in `/dawn/test/unit-test/dawn/Validator/TestUnstructuredDimensionChecker.cpp`
- `dawn/test/unit-test/dawn/IIR/TestIIRSerializer.cpp` was extended to ensure that the generalized weights are (de-)serialized correctly
- The unstructured integration tests were extended with an example close to the ICON diamond FD stencil, using generalized weights

### Dependencies

This PR is independent


mroethlin:
launch jenkins

dawn/src/dawn/Optimizer/PassValidation.cpp
```diff
@@ -52,6 +53,11 @@ bool PassValidation::run(const std::shared_ptr<iir::StencilInstantiation>& insta
                     ("Stage location type consistency check failed at line " +
                      std::to_string(stageConsistencyErrorLocation.Line) + " " + description)
                         .c_str());
+    auto [weightsValid, weightValidErrorLocation] = WeightChecker::CheckWeights(*iir, metadata);
+    DAWN_ASSERT_MSG(weightsValid,
+                    ("Found invalid weights at line " +
+                     std::to_string(stageConsistencyErrorLocation.Line) + " " + description)
```
Stagno:
```suggestion
                     std::to_string(weightValidErrorLocation.Line) + " " + description)
```

dawn/src/dawn/Validator/WeightChecker.cpp
```diff
@@ -0,0 +1,135 @@
...
+    weightsValid_ = false;
+    return;
+  } else {
+    for(const auto& s : expr->getChildren()) {
+      s->accept(*this);
+    }
+  }
+}
+void WeightChecker::WeightCheckerImpl::visit(
+    const std::shared_ptr<dawn::ast::StencilFunArgExpr>& expr) {
```
Stagno: You can skip this, it has no children. It's also only used for structured.

dawn/src/dawn/Validator/WeightChecker.cpp
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+}
+void WeightChecker::WeightCheckerImpl::visit(
+    const std::shared_ptr<dawn::ast::StencilFunCallExpr>& expr) {
+  if(parentIsWeight_) {
+    weightsValid_ = false;
+    return;
+  } else {
+    for(const auto& s : expr->getChildren()) {
```
Stagno: This will only visit the arguments, not the body of a function. I guess you are looking for a nested reduction here, so probably you should also forward the visit to the body.

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #904, opened 2020-04-01
Remove Unnecessary Debug Output

mroethlin:
## Technical Description

Removes accidentally left in Debug printing in python interface


mroethlin:
diff too large

PR #906, opened 2020-04-03
Introduce the ForLoop Concept

### LINKED ISSUE #869: Sparse Dimensions on LHS
Currently, sparse dimensions are only allowed to be consumed in a reduction, but not written to. That is, initialization of sparse dimensions was only possible in driver code. For this task, enable sparse dimensions to be written to. For testing, consider the dot products in the diamond example or the orientation fields in the Laplacian example.

DoD: tests passing

mroethlin:
## Technical Description

This PR introduces a somewhat restricted loop concept throughout the tool chain. This includes SIR, IIR, Codegen, Dawn4Py and (De-)Serialization components. A design document can be found [here](https://docs.google.com/document/d/110IBTnnnnif0M_89KqO7OO_mRJ_iUtOy5lyd7r85cqU/edit#), wherein **Proposal 2** was implemented. Some Notes:
* The loop class is general in the sense that it accepts a block statement and a loop descriptor object, of which the loop bounds are derived. Inly one such loop descriptor was implemented, which is the neighbor chain. More general loop descriptors can be implemented later on. 
* Since the loop index is a neighbor chain, the current loop only really makes sense if there is a sparse dimension with the same dimensions as the chain on the left hand side, hence, the loop with a neighbor chain loop descriptors main purpose is to initialize sparse dimensions
* The contents of the block statement (the body) of the loop are not restricted, they can be very general and were tested to accept even nested reduce over neighbor concepts
* The `UnstructuredDimensionChecker` had to be heavily adapted to correctly check the contents of the loop body

There are two other small changes introduced with this PR which seemed natural to include:
* The sparse index is now a scoped variable and does not need to be reset
* The restriction that sparse fields may not be used in nested reductions was lifted

## Resolves Enhances

Fixes #869 

## Testing

* There are a whole bunch of new tests in the unstructured integration-tests, one for each example discussed in the design document + two tests ensuring the correct working of involved right hand sides
* A number of unit tests for the `UnstructuredDimensionChecker` to ensure the loop body is tested correctly were added
* A new test in `TestSirSerializer`
* A new `dawn4py` test (`loop_copy_stencil.py`)


mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno:
launch jenkins

mroethlin:
launch jenkins

dawn/src/dawn/AST/ASTStringifier.cpp
```diff
@@ -56,6 +56,11 @@ class StringVisitor : public ASTVisitor {
     scopeDepth_--;
   }
 
+  void visit(const std::shared_ptr<LoopStmt>& stmt) override {
+    ss_ << "for (" << stmt->getIterationDescr().toString() << ")\n";
+    stmt->getBlockStmt()->accept(*this);
```
twicki, line 61: do we want to increase scope depth here also?

dawn/src/dawn/CodeGen/ASTCodeGenCXX.cpp
```diff
@@ -103,6 +103,8 @@ void ASTCodeGenCXX::visit(const std::shared_ptr<iir::IfStmt>& stmt) {
   }
 }
 
+void ASTCodeGenCXX::visit(const std::shared_ptr<iir::LoopStmt>& stmt) { dawn_unreachable("TODO"); }
```
twicki, line 106: do we want to solve this?

dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
```diff
@@ -13,14 +13,48 @@
...
 #include "dawn/CodeGen/CXXNaive-ico/ASTStencilFunctionParamVisitor.h"
 #include "dawn/CodeGen/CXXUtil.h"
 #include "dawn/IIR/AST.h"
 #include "dawn/IIR/ASTExpr.h"
+#include "dawn/IIR/ASTFwd.h"
 #include "dawn/IIR/StencilFunctionInstantiation.h"
 #include "dawn/SIR/SIR.h"
 #include "dawn/Support/Unreachable.h"
 
+namespace {
```
twicki, line 26: static function instead of anonymous namespace 
https://llvm.org/docs/CodingStandards.html#id57

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -234,15 +234,20 @@ DawnCompiler::optimize(const std::map<std::string, std::shared_ptr<iir::StencilI
...
-      optimizer.pushBackPass<PassSetStageGraph>();
-      optimizer.pushBackPass<PassSetDependencyGraph>();
-      optimizer.pushBackPass<PassStageReordering>(reorderStrategy);
-      // moved stages around ...
-      optimizer.pushBackPass<PassSetSyncStage>();
-      // if we want this info around, we should probably run this also
-      // optimizer.pushBackPass<PassSetStageName>();
-      // validation check
-      optimizer.pushBackPass<PassValidation>();
+      // TODO: disabled for unstructured as it is broken due to offsets/extents handled as
```
twicki, line 237: this might be a real issue. Do we have it written down somewhere?

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -234,15 +234,20 @@ DawnCompiler::optimize(const std::map<std::string, std::shared_ptr<iir::StencilI
...
+      // cartesian
+      if(stencilInstantiationMap.begin()->second->getIIR()->getGridType() !=
```
twicki, line 239: I would throw a runtime error here instead of silently doing nothing. seems scary

dawn/src/dawn/Compiler/DawnCompiler.cpp
```diff
@@ -300,9 +305,14 @@ DawnCompiler::optimize(const std::map<std::string, std::shared_ptr<iir::StencilI
       optimizer.pushBackPass<PassValidation>();
       break;
     case PassGroup::SetBlockSize:
-      optimizer.pushBackPass<PassSetBlockSize>();
-      // validation check
-      optimizer.pushBackPass<PassValidation>();
+      // TODO: disabled for unstructured as it is broken due to offsets/extents handled as
```
twicki, line 308: see above

dawn/src/dawn/SIR/SIR.cpp
```diff
@@ -565,9 +565,20 @@ CompareResult Field::comparison(const Field& rhs) const {
   return StencilFunctionArg::comparison(rhs);
 }
 
+bool UnstructuredFieldDimension::chainIsValid() const {
```
twicki, line 568: I'm generally happy to have these checks around. So far our assumption was garbage in garbage out in SIR. Do we want a validation module moving forward similar to the verifier we have in IIR.

If so we might want to start with this functionality

dawn/src/dawn/Validator/UnstructuredDimensionChecker.h
```diff
@@ -28,11 +29,22 @@ class UnstructuredDimensionChecker {
 
 private:
   class UnstructuredDimensionCheckerImpl : public ast::ASTVisitorForwarding {
+  public:
+    struct UnstructuredDimensionCheckerConfig {
```
twicki, line 33: why is this declaration public?

dawn/test/integration-test/dawn4py-tests/loop_copy_stencil.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+
+##===-----------------------------------------------------------------------------*- Python -*-===##
+# _
```
twicki, line 4: formatting seems odd

dawn/src/dawn4py/serialization/utils.py
```diff
@@ -1,15 +1,15 @@
 # -*- coding: utf-8 -*-
 ##===-----------------------------------------------------------------------------*- Python -*-===##
-##                          _
-##                         | |
-##                       __| | __ ___      ___ ___
-##                      / _` |/ _` \ \ /\ / / '_  |
-##                     | (_| | (_| |\ V  V /| | | |
-##                      \__,_|\__,_| \_/\_/ |_| |_| - Compiler Toolchain
+# _
+# | |
```
twicki, line 4: ??

dawn/src/dawn/AST/ASTStringifier.cpp
mroethlin: the generated code is a complete mess anyways currently, and we rely on clang-format to make it readable. So far we weren't able to figure out how to do this in cmake (running clang-format on all generated files that is) though, so added for consistency's sake.

dawn/src/dawn/CodeGen/ASTCodeGenCXX.cpp
mroethlin: Maybe eventually. Currently, the only loop descriptor that is implemented works for unstructured meshes only. Still, removed this and deactivated the code gen for this concept the same way we did for the reduction (by declaring an empty body in the header).

dawn/src/dawn/Validator/UnstructuredDimensionChecker.h
mroethlin: maybe I'm missing something obvious here, but the `UnstructuredDimensionCheckerConfig` object is passed to the constructor of the `UnstructuredDimensionCheckerImpl`. So a caller wishing to configure the `UnstructuredDimensionCheckerImpl` has to have access to the `UnstructuredDimensionCheckerConfig`. Would you prefer to have this private and make `UnstructuredDimensionChecker` a friend class?

dawn/src/dawn/SIR/SIR.cpp
mroethlin: I was thinking of this more of as a contract to the argument of the constructor. There is also another check that the chain has at least one entry in the constructor. I can see an argument that we should be as permissive as possible w.r.t the SIR, and handle all the validation in appropriate passes. However, if we should decide to go down that route I'd say we move this to a validator pass in a future PR.

mroethlin:
> maybe add a tests case for the IIR serializer tests?

well, both the SIR as well as the IIR serializer hand the actual (de-)serialization off to the `ASTSerializer`, but I can add one if you insist.

dawn/src/dawn/CodeGen/ASTCodeGenCXX.cpp
twicki: sounds good!

dawn/src/dawn/SIR/SIR.cpp
twicki: sounds good. Just wanted to put it on the radar for discussion

dawn/src/dawn/Validator/UnstructuredDimensionChecker.h
twicki: got it, I was confused. This is fine

dawn/src/dawn/AST/ASTStringifier.cpp
twicki: I am ok with that

mroethlin:
launch jenkins

PR #923, opened 2020-04-14
Fix detection of clang libraries for clang-9+

havogt:
## Technical Description

In some installation of recent clang libraries (probably 9 and later), the libraries are shipped as a single library on some platforms.


gtclang/cmake/FindClang.cmake
```diff
@@ -54,39 +54,38 @@ set(CLANG_RESSOURCE_INCLUDE_PATH "${llvm_install_prefix}/lib/clang/${LLVM_VERSIO
...
-    clangTooling
-    clangToolingCore
-    clangAST
-    clangLex
-    clangBasic
-  )
-  if(${LLVM_VERSION} VERSION_GREATER_EQUAL 7.0.0)
-    list(APPEND clang_libnames clangToolingInclusions)
+  if(${LLVM_VERSION} VERSION_GREATER_EQUAL 9.0.0)
+    set(clang_libnames clang)
```
jdahm, line 58: Did you mean to add these as a list?

gtclang/cmake/FindClang.cmake
```diff
@@ -54,39 +54,38 @@ set(CLANG_RESSOURCE_INCLUDE_PATH "${llvm_install_prefix}/lib/clang/${LLVM_VERSIO
...
+      clangSerialization
+      clangStaticAnalyzerCheckers
+      clangStaticAnalyzerCore
+      clangTooling
+      clangToolingCore
+      clangAST
+      clangLex
+      clangBasic
+    )
+    if(${LLVM_VERSION} VERSION_GREATER_EQUAL 7.0.0)
```
jdahm, line 86: Can the conditional at the bottom be combined with the one at the top?

gtclang/cmake/FindClang.cmake
havogt: I did, but as it works like this, I'll just delete the first line.

gtclang/cmake/FindClang.cmake
havogt: I would keep it like this because the first conditional is splitting between the one-lib and multi-lib clang version and the other one is adding an extra library to the multi-lib.

havogt:
launch jenkins

gtclang/cmake/FindClang.cmake
jdahm: Sure. That makes sense.

havogt:
Apparently not all platforms install the single library version.

havogt:
launch jenkins

havogt:
launch jenkins

havogt:
launch jenkins

havogt:
@jdahm Any idea what's the problem in jenkins? It seems to find the clang-cpp library but then doesn't like to link it.

jdahm:
I took a quick look last night and couldn't figure out what is going on. I'll take a look later and let you know.

jdahm:
This problem is that the symlink exists, but the underlying library it's pointing to does not. Could be a broken Ubuntu package. Would be nice to have this, but it needs to go on hold until we update the testing image.

jdahm:
Requires the `libeling-cpp9` package in the test image. I'm updating the images now and will run a test build after that's done.

jdahm:
launch jenkins

Issue #917, opened 2020-04-09
Implement SIR/IIR syntax model


cosunae:
implement the SIR following the discussions of the workshop

PR #925, opened 2020-04-16
Implement pass to remove advanced statements

Stagno:
## Technical Description

- Add `PassSimplifyStatements` to remove compound assignments and increment/decrement ops. And tests.

### Resolves / Enhances

first point of #917

### DIFF
--- a/dawn/src/dawn/Optimizer/PassSimplifyStatements.cpp
+++ b/dawn/src/dawn/Optimizer/PassSimplifyStatements.cpp
@@ -1,0 +1,69 @@
+//===--------------------------------------------------------------------------------*- C++ -*-===//
+//                          _
+//                         | |
+//                       __| | __ ___      ___ ___
+//                      / _` |/ _` \ \ /\ / / '_  |
+//                     | (_| | (_| |\ V  V /| | | |
+//                      \__,_|\__,_| \_/\_/ |_| |_| - Compiler Toolchain
+//
+//
+//  This file is distributed under the MIT License (MIT).
+//  See LICENSE.txt for details.
+//
+//===------------------------------------------------------------------------------------------===//
+
+#include "PassSimplifyStatements.h"
+#include "dawn/AST/ASTExpr.h"
+#include "dawn/IIR/ASTFwd.h"
+#include "dawn/IIR/DoMethod.h"
+#include "dawn/IIR/StencilInstantiation.h"
+#include "dawn/Support/Type.h"
+#include <memory>
+
+namespace dawn {
+
+bool PassSimplifyStatements::run(
+    const std::shared_ptr<iir::StencilInstantiation>& stencilInstantiation) {
+  for(const auto& doMethod : iterateIIROver<iir::DoMethod>(*stencilInstantiation->getIIR())) {
+    for(auto stmtIt = doMethod->getAST().getStatements().begin();
+        stmtIt != doMethod->getAST().getStatements().end(); ++stmtIt) {
+      if(const auto& exprStmt = std::dynamic_pointer_cast<iir::ExprStmt>(*stmtIt)) {
+        auto sourceLoc = exprStmt->getSourceLocation();
+        if(const auto& assignmentExpr =
+               std::dynamic_pointer_cast<iir::AssignmentExpr>(exprStmt->getExpr())) {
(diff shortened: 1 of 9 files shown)


Stagno:
launch jenkins

Stagno:
launch jenkins

Stagno:
launch jenkins

dawn/src/dawn/Optimizer/PassSimplifyStatements.cpp
```diff
@@ -0,0 +1,106 @@
...
+               std::dynamic_pointer_cast<iir::AssignmentExpr>(exprStmt->getExpr())) {
+          if(assignmentExpr->getOp() != "=") {
+            auto binOp = std::make_shared<ast::BinaryOperator>(
+                assignmentExpr->getLeft()->clone(), assignmentExpr->getOp().substr(0, 1),
+                assignmentExpr->getRight(), sourceLoc);
+            auto newAssignmentExpr = std::make_shared<ast::AssignmentExpr>(
+                assignmentExpr->getLeft(), binOp, "=", sourceLoc);
+            exprStmt->getExpr() = newAssignmentExpr;
+          }
+        }
```
BenWeber42, line 79: Seems to ignore compoung assignments in nested block statements (e.g., nested ifs)

dawn/src/dawn/Optimizer/PassSimplifyStatements.cpp
```diff
@@ -0,0 +1,106 @@
...
+    if(unaryOp->getOp() == "++") {
+      binOp = std::make_shared<ast::BinaryOperator>(
+          unaryOp->getOperand()->clone(), "+",
+          std::make_shared<ast::LiteralAccessExpr>("1", BuiltinTypeID::Integer, sourceLoc),
+          sourceLoc);
+    } else if(unaryOp->getOp() == "--") {
+      binOp = std::make_shared<ast::BinaryOperator>(
+          unaryOp->getOperand()->clone(), "-",
+          std::make_shared<ast::LiteralAccessExpr>("1", BuiltinTypeID::Integer, sourceLoc),
+          sourceLoc);
```
BenWeber42, line 44: Doesn't work in this case:

```
if (i == 0 && ++i == 1) { /* ... */ }
```
will become:
```
i = i + 1;
if (i == 0 && i == 1) { /* ... */ }
```

havogt: This is an interesting one! Can we resolve this (without introducing a temporary variable)?

BenWeber42: A valid transformation output would be:
```
if (i == 0) {
  i = i + 1;
  if (i == 1) { /* ... */ }
}
```
But I don't think that such an output would be suitable for a general transformation algorithm.

Note: this isn't equivalent if SIR also has C++'s short-circuit semantics (it would also introduce more ops):
```
i = i + 1;
if (i - 1 == 0 && i == 1) { /* ... */ }
```

Stagno: As this is not an easy problem to solve, let's leave this open for the next syntax workshop.

BenWeber42: I agree. But do we then merge a pass that only works for a subset of SIR and could change semantics if SIR outside that subset is given?

Stagno: @BenWeber42 I wouldn't merge it for now. Let's briefly present the problem at the review meeting. A quick solution could be saying that increment and decrement ops can't be nested in expressions (i.e. only allowed in statements like i++; i--;) from the beginning (frontends/SIR). If this solution is not widely accepted, then this problem will need to be discussed more in detail, because it seems that the only way to solve it is to do the heavy transformation that you showed.

dawn/src/dawn/Optimizer/PassSimplifyStatements.cpp
Stagno: Thx, need to fix this.

Stagno:
launch jenkins

Stagno:
launch jenkins

PR #936, opened 2020-04-21
Replace FileUtil with std::filesystem calls

jdahm:
## Technical Description

Removes FileUtil, which has been superseded by std::filesystem.


jdahm:
launch jenkins

PR #937, opened 2020-04-22
Add runWithoutSync()

havogt:
## Technical Description

For gt4py we need a run-method which doesn't call sync() on data_stores. This is a quickfix.


havogt:
launch jenkins

havogt:
To not break API such a flag would have to go last with a default value. I thought it will look very ugly if you have a lot of fields (uniform arguments) and then an optional bool at the end.

twicki:
Should have been more specific. I meant a dawn flag

havogt:
This wasn't discussed. Actually I have no opinion here, all we do will be a hack...

twicki:
I think it would be a little cleaner since it would make sure that the driver code looks the same (and would fix the comparison to generated code tests :wink:) but you are right, it is a temporary solution so I'm ok with either option

havogt:
Replaced by #967

PR #940, opened 2020-04-24
Declare cuda globals once

### LINKED ISSUE #939: CUDA backend should only declare globals once
The `CudaCodeGen` class currently generates globals (e.g., `globalOffsets`) once per multistage (i.e., kernel) rather than once per stencil.

eddie-c-davis:
## Technical Description

This PR modifies the `MSCodeGen` class to only declare global CUDA variables and functions once per translation unit, rather than once per kernel.

### Resolves / Enhances

Resolves #939


eddie-c-davis:
launch jenkins

PR #955, opened 2020-05-07
Replace logger

jdahm:
## Technical Description

Replace Logger by a similar but improved infrastructure.

A follow-up PR will replace the diagnostics with this.

The Logger design is similar to before, however there are separate loggers `dawn::log::info`, `dawn::log::warning`, and `dawn::log::error` which are globals similar to `std::cout` and `std::cerr`.

Each logger has an `LoggerProxy operator()(string, int)` as before that returns the stream operator

Each logger also acts as a container that stores the messages (a la `DiagnosticsEngine`).

Instead of a virtual LoggerInterface that potentially leaks memory when registering other class instance pointers, this is replaced with a lightweight std::function object `Logger::Formatter`, so GTClang and other frameworks can register their own message formatting.

Loggers also take underlying `std::ostream`s, which can be customized to write to strings, files, etc.

### Resolves / Enhances

Part of #950.

### Testing

Integrates into ctest.

### DIFF
--- a/dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
+++ b/dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
@@ -234,8 +234,8 @@ bool PassRemoveScalars::run(
   // Check if we have unsupported statements. If we do, warn the user and skip the pass execution.
   for(const auto& stmt : iterateIIROverStmt(*stencilInstantiation->getIIR())) {
     if(isStatementUnsupported(stmt, stencilInstantiation->getMetaData())) {
-      DAWN_LOG(INFO) << "Unsupported statement at line " << stmt->getSourceLocation()
-                     << ". Skipping removal of scalar variables.";
+      DAWN_DIAG(INFO, stencilInstantiation->getMetaData().getFileName(), stmt->getSourceLocation())
+          << "Unsupported statement. Skipping removal of scalar variables...";
       return true;
     }
   }
@@ -245,11 +245,9 @@ bool PassRemoveScalars::run(
     auto removedScalars = removeScalarsFromDoMethod(*doMethod, stencilInstantiation->getMetaData());
     if(context_.getOptions().ReportPassRemoveScalars) {
       for(const auto& varName : removedScalars) {
-        std::cout << "PASS: " << getName() << ": " << stencilInstantiation->getName()
-                  << ": DoMethod: " << doMethod->getID() << " removed variable: " << varName
-                  << std::endl;
+        DAWN_LOG(INFO) << stencilInstantiation->getName() << ": DoMethod: " << doMethod->getID()
+                       << " removed variable: " << varName;
       }
-      std::cout.flush();
     }
     // Recompute extents of fields
     doMethod->update(iir::NodeUpdateType::level);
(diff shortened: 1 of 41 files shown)


jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

jdahm:
launch jenkins

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -234,8 +234,8 @@ bool PassRemoveScalars::run(
   // Check if we have unsupported statements. If we do, warn the user and skip the pass execution.
   for(const auto& stmt : iterateIIROverStmt(*stencilInstantiation->getIIR())) {
     if(isStatementUnsupported(stmt, stencilInstantiation->getMetaData())) {
-      DAWN_LOG(INFO) << "Unsupported statement at line " << stmt->getSourceLocation()
-                     << ". Skipping removal of scalar variables.";
+      DAWN_DIAG(INFO, stencilInstantiation->getMetaData().getFileName(), stmt->getSourceLocation())
```
jdahm, line 237: Here is an example of the diagnostic in code.

dawn/src/dawn/Optimizer/PassRemoveScalars.cpp
```diff
@@ -245,11 +245,9 @@ bool PassRemoveScalars::run(
     auto removedScalars = removeScalarsFromDoMethod(*doMethod, stencilInstantiation->getMetaData());
     if(context_.getOptions().ReportPassRemoveScalars) {
       for(const auto& varName : removedScalars) {
-        std::cout << "PASS: " << getName() << ": " << stencilInstantiation->getName()
-                  << ": DoMethod: " << doMethod->getID() << " removed variable: " << varName
-                  << std::endl;
+        DAWN_LOG(INFO) << stencilInstantiation->getName() << ": DoMethod: " << doMethod->getID()
```
jdahm, line 248: Switched to `DAWN_LOG(INFO)` instead.

dawn/src/dawn/Serialization/SIRSerializer.cpp
```diff
@@ -52,10 +53,10 @@ class ProtobufLogger : public NonCopyable {
       DAWN_LOG(WARNING) << "Protobuf: " << message;
       break;
     case google::protobuf::LOGLEVEL_ERROR:
-      DAWN_LOG(ERROR) << "Protobuf: " << message;
+      throw SyntacticError(std::string("[ERROR] Protobuf error: ") + message);
       break;
     case google::protobuf::LOGLEVEL_FATAL:
-      DAWN_LOG(FATAL) << "Protobuf: " << message;
+      throw SyntacticError(std::string("[FATAL] Protobuf error occurred: ") + message);
```
jdahm, line 59: This should match the other error before merging.

jdahm:
launch jenkins

jdahm:
Replace Logger by a similar but improved infrastructure.

A follow-up PR will replace the diagnostics with this.

The Logger design is similar to before, however there are separate loggers dawn::log::info, dawn::log::warning, and dawn::log::error which are globals similar to std::cout and std::cerr.

Each logger has an LoggerProxy operator()(string, int) as before that returns the stream operator

Each logger also acts as a container that stores the messages (a la DiagnosticsEngine).

Instead of a virtual LoggerInterface that potentially leaks memory when registering other class instance pointers, this is replaced with a lightweight std::function object Logger::Formatter, so GTClang and other frameworks can register their own message formatting.

Loggers also take underlying std::ostreams, which can be customized to write to strings, files, etc.

Part of #950.

Integrates into ctest.

PR #956, opened 2020-05-07
Fixup and expand exceptions

jdahm:
## Technical Description

Fixes issue with unsigned in exceptions and adds a few more.


jdahm:
launch jenkins

PR #967, opened 2020-05-13
Run without sync compiler option

eddie-c-davis:
## Technical Description

Modifies the `runWithoutSync` method to be a codegen option rather than generating two versions of the run method in every backend.

### Resolves / Enhances

PR #937


eddie-c-davis:
launch jenkins

PR #969, opened 2020-05-17
add stdc++fs also for GNU CXX compiler

cosunae:
## Technical Description

* Fix adding -lstdc++fs for gcc 8
On ubuntu 18.04, gcc8.4.0, cmake 3.16.2: CMAKE_COMPILER_IS_GNUCC is not set 

* install the interface with setup.py as external sources for dawn4py

### DIFF
--- a/dawn/setup.py
+++ b/dawn/setup.py
@@ -57,13 +57,14 @@ with open(os.path.join(DAWN_DIR, "version.txt"), mode="r") as f:
 shutil.copyfile(os.path.join(DAWN_DIR, "version.txt"), os.path.join(DAWN4PY_DIR, "version.txt"))
 
 # Copy additional C++ headers for the generated code
-target_path = os.path.join(DAWN4PY_DIR, "_external_src", "driver-includes")
-if os.path.exists(target_path):
-    shutil.rmtree(target_path)
-shutil.copytree(
-    os.path.join(DAWN_DIR, "src", "driver-includes"), target_path,
-)
 
+for srcdir in ["driver-includes", "interface"]:
+  target_path = os.path.join(DAWN4PY_DIR, "_external_src", srcdir)
+  if os.path.exists(target_path):
+    shutil.rmtree(target_path)
+  shutil.copytree(
+      os.path.join(DAWN_DIR, "src", srcdir), target_path,
+  )
 
 # Based on:
 #   https://www.benjack.io/2018/02/02/python-cpp-revisited.html
--- a/dawn/src/dawn/CMakeLists.txt
+++ b/dawn/src/dawn/CMakeLists.txt
@@ -24,7 +24,8 @@ install(
 # GCC 8, still a common compiler, requires linking to a separate library for std::filesystem support.
 # Once that is no longer common we can remove this hack
 add_library(cpp_extra_libs INTERFACE)
-if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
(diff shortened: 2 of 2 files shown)


dawn/src/dawn/CMakeLists.txt
```diff
@@ -24,7 +24,8 @@ install(
 # GCC 8, still a common compiler, requires linking to a separate library for std::filesystem support.
 # Once that is no longer common we can remove this hack
 add_library(cpp_extra_libs INTERFACE)
-if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
+if ((CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX) AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
```
havogt, line 27:
```suggestion
if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
```
I think it should be like that (check the c compiler as in the original version doesn't make sense...)

jdahm: Yes, we remove the C compiler. This is just an artifact from before. Good catch!

dawn/setup.py
```diff
@@ -57,13 +57,14 @@
...
 
 # Copy additional C++ headers for the generated code
-target_path = os.path.join(DAWN4PY_DIR, "_external_src", "driver-includes")
-if os.path.exists(target_path):
-    shutil.rmtree(target_path)
-shutil.copytree(
-    os.path.join(DAWN_DIR, "src", "driver-includes"), target_path,
-)
 
+for srcdir in ["driver-includes", "interface"]:
```
jdahm, line 61: 👍

dawn/src/dawn/CMakeLists.txt
jdahm, line 27: I like all of these changes except `CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0`. Here is why:

`libstdc++fs` is necessary for linking to the `Filesystem Technical Specification` [1] that exists in GCC 7 [2]. However, as of GCC 8 [3] this was removed. The relevant part of the CHANGELOG is:

> Runtime Library (libstdc++)
>   - Improved experimental support for C++17, including the following features:
>     - Deduction guides to support class template argument deduction.
>     - std::filesystem implementation.

If this is not the case on a system then it is the fault of the installation, not the compiler itself, and we should get to the bottom of it instead of linking `libstdc++fs` into libraries linked with GCC-8.

I would like to propose merging all but this change for now and investigating further!

[1] https://gcc.gnu.org/onlinedocs/libstdc++/manual/using.html
[2] https://stackoverflow.com/questions/45867379/why-does-gcc-not-seem-to-have-the-filesystem-standard-library
[3] https://gcc.gnu.org/gcc-8/changes.html

jdahm: This appears to be an Ubuntu-specific bug with default library versions. GCC 8 there links to `libstdc++.so.6.0.26`, which does not have the symbols, but later versions seem to include it. The work-around there is to link to `libstdc++fs`, even though it shouldn't be necessary if you read the GCC docs.

Detailed info here: https://bugs.launchpad.net/ubuntu/+source/gcc-8/+bug/1824721

I've changed my mind on this: since most of our Dawn usage is Ubuntu, I think we should merge this.

cosunae:
launch jenkins

cosunae:
thanks @jdahm @havogt for the review

cosunae:
launch jenkins

PR #971, opened 2020-05-18
Fix global precision in codegen

Stagno:
## Technical Description

(problem discovered by Joerg Behrens)
Fixes codegen to output float/double globals in full precision (instead of default precision).

### Example

Globals d17=1.1234567890123457 and d13=1.1234567890123. Both used to get cut to 1.123457

### DIFF
--- a/dawn/src/dawn/SIR/SIR.cpp
+++ b/dawn/src/dawn/SIR/SIR.cpp
@@ -21,6 +21,7 @@
 #include "dawn/Support/Printing.h"
 #include "dawn/Support/StringUtil.h"
 #include "dawn/Support/Unreachable.h"
+#include <iomanip>
 #include <iostream>
 #include <sstream>
 
@@ -722,6 +723,7 @@ BuiltinTypeID sir::Value::typeToBuiltinTypeID(sir::Value::Kind type) {
 }
 
 std::string sir::Value::toString() const {
+  std::ostringstream out;
   DAWN_ASSERT(has_value());
   switch(type_) {
   case Kind::Boolean:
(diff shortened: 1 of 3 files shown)


dawn/src/dawn/SIR/SIR.cpp
```diff
@@ -722,16 +723,20 @@ BuiltinTypeID sir::Value::typeToBuiltinTypeID(sir::Value::Kind type) {
...
+  std::ostringstream out;
   DAWN_ASSERT(has_value());
   switch(type_) {
   case Kind::Boolean:
     return std::get<bool>(*value_) ? "true" : "false";
   case Kind::Integer:
     return std::to_string(std::get<int>(*value_));
   case Kind::Double:
-    return std::to_string(std::get<double>(*value_));
+    out << std::setprecision(std::numeric_limits<double>::digits10 + 1)
```
cosunae, line 734: just curiosity, is the +1 necessary?

Stagno: Apparently yes https://en.cppreference.com/w/cpp/io/manip/setprecision
to convert from scale to precision

Stagno:
launch jenkins

Stagno:
launch jenkins

BenWeber42:
I stumbled on this when I reviewed my github notifications. I know it's been already merged, but I think it's still buggy. We should use `max_digits10` : https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10

`digits10` + 1 gives for **float 7** and **for double 16** ( https://en.cppreference.com/w/cpp/types/numeric_limits/digits10 ).
But from `max_digits10` documentation:

> Unlike most mathematical operations, the conversion of a floating-point value to text and back is exact as long as at least max_digits10 were used (**9 for float**, **17 for double**)

PR #972, opened 2020-05-19
Fixed arrayAddons.hpp for gcc 10.1

BenWeber42:
## Technical Description

Fixes incorrect usage of `size_t` in `arrayAddons.hpp`

### Resolves / Enhances

None.

### Example

Compiliation with gcc 10.1.

### Testing

Was tested locally with gcc 10.1.

### Dependencies

None

### Commets

We could also include `stddef.h` instead and then `size_t` will be defined in the _global_ namespace.

### DIFF
--- a/dawn/src/driver-includes/arrayAddons.hpp
+++ b/dawn/src/driver-includes/arrayAddons.hpp
@@ -14,15 +14,16 @@
 
 #pragma once
 
+#include <cstddef>
 #include <array>
 
 namespace gridtools {
 namespace dawn {
 
-template <size_t N>
+template <std::size_t N>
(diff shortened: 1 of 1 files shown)


BenWeber42:
launch jenkins

BenWeber42:
Yea, I think this one should be pretty simple. If someone prefers just `size_t` over `std::size_t` we can inluce stddef.h instead (cstddef.h wraps `size_t` in `std::`). Otherwise, we can just merge it.

Not sure what's the process here. I can't request additional reviewers. Do I need to use any other trigger phrases than `launch jenkins`?

havogt:
Once green, you can merge.

PR #975, opened 2020-05-22
do not include storage_runtime if no GT

cosunae:
## Technical Description

For the unstructured code generation we do not use gridtools storages, nor the storage_runtime.hpp


cosunae:
launch jenkins

PR #985, opened 2020-05-28
Fix pragma once warning in AST/Offsets

jdahm:
## Technical Description

Using #pragma once uncovered an issue in AST/Offsets.cpp. This PR fixes that and cleans up the file.

### DIFF
--- a/dawn/src/dawn/AST/Offsets.h
+++ b/dawn/src/dawn/AST/Offsets.h
@@ -22,7 +22,8 @@
 #include <memory>
 #include <string>
 
-namespace dawn::ast {
+namespace dawn {
+namespace ast {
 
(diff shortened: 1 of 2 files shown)


jdahm:
launch jenkins

dawn/src/dawn/AST/Offsets.h
```diff
@@ -22,7 +22,8 @@
 #include <memory>
```
eddie-c-davis, line 22: Should the `#pragma once` be here in the header?

jdahm: Looks like the GitHub diff hid it. It's on line 15 above.

PR #986, opened 2020-06-01
Standalone lowering

jdahm:
## Technical Description

Move IIR lowering out of the OptimizerContext class. Removed `KeepVarnames` option in the process. The removed option becomes a constructor argument to `StatementMapper` so most of the behavior remains the same.

### DIFF
--- a/dawn/src/dawn/Optimizer/OptimizerContext.cpp
+++ b/dawn/src/dawn/Optimizer/OptimizerContext.cpp
@@ -658,8 +658,6 @@ std::map<std::string, std::shared_ptr<iir::StencilInstantiation>>
 toStencilInstantiationMap(const SIR& stencilIR) {
   std::map<std::string, std::shared_ptr<iir::StencilInstantiation>> stencilInstantiationMap;
 
-  // FillIIR()
-  // {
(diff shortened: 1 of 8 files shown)


jdahm:
launch jenkins

dawn/src/dawn/Optimizer/OptimizerContext.cpp
```diff
@@ -350,7 +347,7 @@ class StencilDescStatementMapper : public iir::ASTVisitor {
     // This is the first time we encounter this variable. We have to make sure the name is not
     // already used in another scope!
 
-    int AccessID = metadata_.addStmt(context_.getOptions().KeepVarnames, stmt);
+    int AccessID = metadata_.addStmt(false, stmt);
```
eddie-c-davis, line 350: Should this use `StatementMapper.keepVarnames` or is that inaccessible here?

jdahm: Good idea. It was removed because Options were not accessible at that point, but as of #990 that is again the case.

jdahm:
Closing this PR in favor of #990, which includes this.

PR #988, opened 2020-06-02
Small fixes to atlas_interface.hpp

Stagno:
## Technical Description

- Adds missing API to `Field` and `SparseDimension` in atlas_interface.hpp (required by cuda ico backend).
- Adds the `inline` keyword to functions which are defined in the header


Stagno:
launch jenkins

PR #989, opened 2020-06-02
Move some Support files to GTClang

jdahm:
## Technical Description

Moves some files in dawn/Support to gtclang/Support and remove the now-unused Twine.


jdahm:
launch jenkins

jdahm:
Presumably more could now be removed since these are in GTClang itself (that links to LLVM and Clang).

dawn/src/dawn/Support/StringUtil.h
```diff
@@ -129,4 +129,16 @@ extern std::string decimalToOrdinal(int dec);
 
 /// @}
 
+/// @brief Checks if two strings are equal (case insensitive)
+/// @ingroup support
+bool equals_lower(const std::string& a, const std::string& b);
```
mroethlin, line 134: these should probably be camelCase for consistency w/ rest of codebase

jdahm: I thought that we wanted standalone methods to use snake case. @twicki?

jdahm: @twicki and I discussed this and landed on camelCase, so I changed it to use that.

jdahm:
launch jenkins

PR #991, opened 2020-06-05
Fix bug in unstructured Cuda Codegen

mroethlin:
## Technical Description

The unstructured cuda code emitted was guarding that the number of threads is smaller then the number of entities worked on in the horizontal, but not in the vertical. This bug was masked by the fact that tests often spool up thread numbers in powers of 2, which happen to be divisible the BLOCK_SIZE. This PR fixes that issue. 

### Dependencies

This PR is independent


mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #992, opened 2020-06-09
Use max_digits10 for serialization

BenWeber42:
## Technical Description

`max_digits10` guarantees no loss when serializing -> deserializing

> Unlike most mathematical operations, the conversion of a floating-point value to text and back is exact as long as at least max_digits10 were used (9 for float, 17 for double)

from https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10


BenWeber42:
launch jenkins

PR #993, opened 2020-06-09
Changed lambdas to consts in CXXUtil.h

### LINKED ISSUE #973: GCC 10.1 fails due to globals in `CXXUtil.h`
`CXXUtil.h` defines the following globals: https://github.com/MeteoSwiss-APN/dawn/blob/727835e027b8270d667452b857fa4e0af57b2c2e/dawn/src/dawn/CodeGen/CXXUtil.h#L672
```C++
auto c_gt = []() { return Twine("gridtools::"); };
auto c_dgt = []() { return Twine("gridtools::dawn::"); };
auto c_gt_enum = []() { return Twine("gridtools::enumtype::"); };
auto c_gt_intent = []() { return Twine("gridtools::intent::"); };
```

GCC 10.1 fails for those, since this header is included multiple times and thus the global variables `c_gt`, etc will be defined multiple times. This leads to errors during the linking stage.

One way we could resolve this is by using only function declarations (without implementation) in the header like so:
```C++
Twine c_gt();
...
```

And then implement those functions in an appropriate `.cpp` file. However, `CXXUtil.h` is currently _header only_, so we'd probably want to introduce a corresponding `.cpp` file and add it to the cmake files.

I think there might also be ways to make the linker happy while keeping the implementations in the header. Maybe the `inline` or `constexpr` keywords could help. I haven't been able to investigate this a lot.

BenWeber42:
## Technical Description

Previously global variables (`c_gt` etc) were declared in the `CXXUtil.h` header. Since gcc 10.1 this leads to linker errors because multiple definitions of those globals were found (this is correct behavior). Now lambdas are no longer used, since the values are constant anyways.

Unfortunately `constexpr` can't be used, because `std::string` can only become a `constexpr` since C++20. `std::string_view` could be used as `constexpr`, but doesn't support the `+` operator with `const char*` which is used in many files. `const std::string` solves both the linking problem and then `+` operator problem.

### Resolves / Enhances

resolves #973

### Testing

This was tested locally on my machine with gcc 10.1


BenWeber42:
launch jenkins

dawn/src/dawn/CodeGen/CXXUtil.h
```diff
@@ -661,10 +661,10 @@ struct Struct : public Structure {
       : Structure("struct", name, s, templateName) {}
 };
 
-auto c_gt = []() { return std::string("gridtools::"); };
-auto c_dgt = []() { return std::string("gridtools::dawn::"); };
-auto c_gt_enum = []() { return std::string("gridtools::enumtype::"); };
-auto c_gt_intent = []() { return std::string("gridtools::intent::"); };
+const std::string c_gt = "gridtools::";
```
jdahm, line 664: I like the idea of simplifying this, but won't this potentially cause duplicate symbols? Would declaring here and defining in the source be an option?

BenWeber42: There's is no source/cpp file for `CXXUtil.h`. Without `const` it leads to duplicate symbols (-> linker error -> multiple definitions).

We're not sure whether adding a cpp file and adjusting the cmake files makes sense for this small change.

BenWeber42: I looked at this: https://en.cppreference.com/w/cpp/language/storage_duration
And `const` doesn't necessarily implie `static`, but `static` should still be redundant in this case.

`static` implies:

> static - static or thread storage duration and internal linkage. 

`const` implies internal linkage:

> Any of the following names declared at namespace scope have internal linkage:
> ...
>  non-volatile non-template (since C++14) non-inline (since C++17) non-exported (since C++20) **const-qualified variables** (including constexpr) that aren't declared extern and aren't previously declared to have external linkage;
> ...

Compiling with `static const` succeeds with gcc 10.1

BenWeber42:
We should try to merge this. It works with both `const` and `static const`.
Yes, `const` will make it `static` under the hood in that it will get _internal linkage_, i.e. it's only _visible_ within the same translation unit. Please chose whether `const` or `static const` is preferable.

BenWeber42:
launch jenkins

BenWeber42:
launch jenkins

BenWeber42:
Current branch successfully compiles on my local machine with gcc 10.1.0.

PR #1001, opened 2020-06-22
Fix: CUDA tests are disabled if CUDA should be enabled

havogt:
## Technical Description

There is a typo in setting `GTCLANG_BUILD_TESTING_GT_CUDA` and `GTCLANG_BUILD_TESTING_PLAIN_CUDA` which disables CUDA tests by default if `ENABLE_CUDA_IF_FOUND == ON`. Not sure if testing is affected (if the flags were set explicitly tests are ok).


havogt:
launch jenkins

PR #1005, opened 2020-06-26
Add GlobalGpuTriMesh, raw ptr ctor and export bindings (cuda-ico)

Stagno:
## Technical Description

Required for interoperability with ICON Fortran code.
Purpose here is to bring ICON neighbor tables (need a conversion) and ICON fields (already in GPU memory) into dawn's generated code as raw pointers.

Note: removes codegen unit tests of unstructured backends. Reason is that they are cumbersome to maintain for fast-changing backends (they break on almost every change to unstructured) and their requirements are already met by the e2e tests (which do checks on the output).


Stagno:
launch jenkins

Stagno:
launch jenkins

PR #1007, opened 2020-07-02
Fix double logger prefix

jdahm:
## Technical Description

Logger accidentally added the prefix e.g. WARNING twice. This fixes it.


jdahm:
launch jenkins

Issue #720, opened 2020-02-18
Add checks to PassValidation


Stagno:
After that the location type of a local variable has been determined:

- We should check that accesses to that variable are type-compatible with other accesses (fields or variables) in the same expression.

After that scalars have been removed:

- We should check that no scalars exist in the whole IIR.

After that the location type of a stage has been set:

- We should check that it is correct with respect to its content (statements).

PR #1014, opened 2020-07-17
Fix a Bug in the Unstructured Dimensions Checker

mroethlin:
## Technical Description

There is a problem with the unstructured dimensions checker: VarDeclStmts as well as VarAccessExprs were never considered in the visitor, leaving some statements without a location type. This marked actually valid IIR as invalid.

### Resolves / Enhances

Addresses first bullet point of [#720](https://github.com/MeteoSwiss-APN/dawn/issues/720)

### Testing

New test in `TestUnstructuredDimensionsChecker`

### DIFF
--- a/dawn/src/dawn/AST/ASTStmt.h
+++ b/dawn/src/dawn/AST/ASTStmt.h
@@ -111,6 +111,8 @@ public:
     return dynamic_cast<DataType&>(*data_.get());
   }
 
+  bool hasData() const { return data_ != nullptr; }
+
   /// @brief Iterate children (if any)
   virtual StmtRangeType getChildren() { return StmtRangeType(); }
 
--- a/dawn/src/dawn/AST/ASTStmt.h
+++ b/dawn/src/dawn/AST/ASTStmt.h
@@ -111,8 +111,6 @@ public:
     return dynamic_cast<DataType&>(*data_.get());
   }
 
-  bool hasData() const { return data_ != nullptr; }
-
   /// @brief Iterate children (if any)
   virtual StmtRangeType getChildren() { return StmtRangeType(); }
 
--- a/dawn/src/dawn/Validator/UnstructuredDimensionChecker.cpp
+++ b/dawn/src/dawn/Validator/UnstructuredDimensionChecker.cpp
@@ -119,8 +119,20 @@ UnstructuredDimensionChecker::UnstructuredDimensionCheckerImpl::getDimensions() 
 
 void UnstructuredDimensionChecker::UnstructuredDimensionCheckerImpl::visit(
     const std::shared_ptr<iir::VarDeclStmt>& stmt) {
-  auto newAccessID = *stmt->getData<iir::VarDeclStmtData>().AccessID;
-  const auto varDeclInfo = idToLocalVariableData_.at(newAccessID);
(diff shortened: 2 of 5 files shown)


mroethlin:
launch jenkins

dawn/src/dawn/AST/ASTStmt.h
```diff
@@ -111,6 +111,8 @@ class Stmt : public std::enable_shared_from_this<Stmt> {
     return dynamic_cast<DataType&>(*data_.get());
   }
 
+  bool hasData() const { return data_ != nullptr; }
```
Stagno, line 114: A statement has always a data object attached, also in SIR (it was envisioned to move SIR-specific structures such as `sir::VerticalRegion` into the respective stmt's data, but it never became a priority...). Therefore, a call to this will always return true.
It was envisioned to mirror the same pattern for data objects used in stmts also in expressions, but this was, again, never implemented (currently they are created on-the-fly in the first call to `getData()`, a horrible pattern ...).
From the usage below it seems that you introduced these `hasData()` to determine whether you are visiting SIR or IIR. As we start to specialize the unstruct. dim. checker's implementation for IIR-specific checks, I think a cleaner way to do this would be to introduce and propagate a boolean flag `isIIR` that would enable such checks.

dawn/src/dawn/Validator/UnstructuredDimensionChecker.cpp
```diff
@@ -103,15 +106,58 @@ UnstructuredDimensionChecker::UnstructuredDimensionCheckerImpl::UnstructuredDime
...
 
+void UnstructuredDimensionChecker::UnstructuredDimensionCheckerImpl::visit(
+    const std::shared_ptr<iir::VarDeclStmt>& stmt) {
+  if(!dimensionsConsistent_) {
+    return;
+  }
+  // data is not set when starting from SIR
+  if(!stmt->hasData()) {
+    return;
+  }
```
Stagno, line 128: See comment above

dawn/src/dawn/Validator/UnstructuredDimensionChecker.cpp
```diff
@@ -142,6 +188,51 @@ void UnstructuredDimensionChecker::UnstructuredDimensionCheckerImpl::visit(
...
+    curDimensions_ = sir::FieldDimensions(
+        sir::HorizontalFieldDimension{ast::unstructured, ast::LocationType::Edges}, true);
+    break;
+  case iir::LocalVariableType::OnVertices:
+    curDimensions_ = sir::FieldDimensions(
+        sir::HorizontalFieldDimension{ast::unstructured, ast::LocationType::Vertices}, true);
+    break;
+  default:
+    break;
+  }
```
Stagno, line 233: Same code as above. We could use a private method here.

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #1015, opened 2020-07-20
Fix Bugs Related to Empty Regions

### LINKED ISSUE #304: Single empty vertical region compilation results in stack trace
The following code
``` cpp

stencil ssa_issue {
  Do {
    }
  }
};
```
cannot be compiled and results in a stack trace with no additional message. See the [attached file](https://github.com/MeteoSwiss-APN/gtclang/files/3646924/stacktrace.txt) for more details.

### LINKED ISSUE #997: if/else can crash with empty bodies
While testing dusk's implementation of if/else statements, various bugs in danw's handling of SIR's if/else statements were exposed. They occured whenever the _if-body_ or _else-body_ had no statements. Nested if/else seems to make the matter worse.

```python
        # currently broken in dawn
        # elif no else
        if False:
            a = b
        elif True:
            c = a + 1

        # currently broken in dawn
        # elif as well
        if False:
           a = b
        elif True:
           c = a + 1
        else:
           c = a - 1
```

Dusk will create nested if/else to handle python's `elif`.

mroethlin:
## Technical Description

This PR Fixes some problems caused by empty regions and some related problems

### Resolves / Enhances

* Was supposed to fix #304, but it was already fixed beforehand 🤷
* A problem where ` PassFixVersionedInputFields` introduced new stages, for which the stage location type was never deducted. 
* Fixes  #997 which describes a problem that was suspected to be related to an empty `else` body but then wasn't. In this stencil
```
 @stencil
 def getDenseLocationTypeFromAccessIDBUG(
     a: Field[Edge], b: Field[Edge], c: Field[Edge], d: Field[Edge], e: Field[Edge]
 ):
     with levels_downward[-5:] as k:
         a = b / c + 5
         if d:
             a = b
         elif e:
             c = a + 1
```
`c` is versioned to `c0`. The stage location type deduction could then not figure out stages containing c0, since a method retrieving meta data forgot to look in the `variableVersions` map. 
*  Problems like 
```
def temp_fields3(a: Field[Edge], b: Field[Edge]):
    x: Field[Edge]
    with levels_downward as k:
        x = 1
```
wherein x=1 was first demoted to a scalar and subsequently removed, leading to an empty stage. 

### Testing

since most of these bugs emerge from interaction of different passes I decided to introduce two additional dawn4py tests, which compile code equivalent to the two stencils discussed in this comment


mroethlin:
launch jenkins

dawn/src/dawn/Optimizer/TemporaryHandling.cpp
```diff
@@ -152,10 +152,12 @@ void demoteTemporaryFieldToLocalVariable(iir::StencilInstantiation* instantiatio
...
   //
   iir::ExprStmt* exprStmt = dyn_cast<iir::ExprStmt>(oldStatement.get());
-  DAWN_ASSERT_MSG(exprStmt, "first access of field (i.e lifetime.Begin) is not an `ExprStmt`");
+  DAWN_ASSERT_MSG(
+      exprStmt,
+      "first access of temp field to be demoted (i.e lifetime.Begin) is not an `ExprStmt`");
   iir::AssignmentExpr* assignmentExpr = dyn_cast<iir::AssignmentExpr>(exprStmt->getExpr().get());
-  DAWN_ASSERT_MSG(assignmentExpr,
-                  "first access of field (i.e lifetime.Begin) is not an `AssignmentExpr`");
+  DAWN_ASSERT_MSG(assignmentExpr, "first access of of temp field to be demoted field (i.e "
```
Stagno, line 159:
```suggestion
  DAWN_ASSERT_MSG(assignmentExpr, "first access of temp field to be demoted (i.e "
```

mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #1018, opened 2020-08-10
CUDA Ico Backgen: Interface Changes

mroethlin:
## Technical Description

This PR adapts the cuda-ico codegen to accept new entry points. We now have 3 ways to launch cuda kernels

- A collection of `unstructured_interface` fields (ease of testing)
- A bunch of raw pointers on the GPU (test / launch from ICON)
- A bunch of raw pointer on the CPU (allows convenient point of entry when launching the kernel via the cpp_bindgen headers from C or C++)

### Testing

Tested manually using icondusk-e2e

### Dependencies

This PR is independent


mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #1020, opened 2020-09-07
Unstructured Interface now Distinguishes between FORTRAN and C Fields

mroethlin:
## Technical Description

This PR adds additional "hooks" into the dawn unstructured generated code. Depending if one calls the stencil from c or Fortran, the fields are either transposed or not, lifting this responsibility from the driver code. 

### Testing

Tested externally using the icondusk-e2e repo


mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (approved):
Approved. Not 100% sure about the naming `from_c_host` vs `from_fort_host` to imply a shape, but couldn't come up with better names.

PR #1022, opened 2020-09-10
Update field access name

### LINKED ISSUE #591: Issue with Access IDs on FieldAccesExpr
## Introduction

The `FieldAccessExpr` inherits an id from `Expr` and stores its name. In the case of nested stencils, both are subject to change. The `StencilDescStatementMapper` inlines nested stencils and uses the `StatementMapper` to do the renaming needed. The `StatementMapper` leaves the ID and Name stored in the `FieldAccessExpr` alone, and instead stores the new ID at
```
expr->getData<iir::IIRAccessExprData>().AccessID = std::make_optional(AccessID);
``` 
The correct name can then be retrieved by means of the `StencilMetaInformation`, the `FieldAccessExpr::Name`member is not updated and remains stale.

It is currently not clear why this is implemented in this way. It seems more natural to simply update the information in place instead of keeping a stale copy of the old data around. Code which relies on this information is error prone, because the correct information is not in the most obvious place.

## Description

The obvious solution of simply adding the following lines 
```
  expr->setID(AccessID); //update ID
  expr->setName(metadata_.getFieldNameFromAccessID(AccessID)); //update Name
```
to `StatementMapper::visit(const std::shared_ptr<iir::FieldAccessExpr>& expr)` leads to out of range accesses in some maps during the compilation of the CodeGen tests. 

## ToDo

Investigate and implement a way to have the above information stored only once, and always up to date.

Stagno:
## Technical Description

Update `FieldAccessExpr::getName()` to be the correct name (found in metadata).
Adds equality check in `IntegrityChecker`.

### Resolves / Enhances

resolves #591

### DIFF
--- a/dawn/src/dawn/IIR/DoMethod.cpp
+++ b/dawn/src/dawn/IIR/DoMethod.cpp
@@ -55,8 +55,12 @@ public:
       expr->accept(*this);
   }
   void visit(const std::shared_ptr<VarAccessExpr>& expr) override {
-    auto data = expr->getData<iir::IIRAccessExprData>();
-    int accessID = *data.AccessID;
+    int accessID = iir::getAccessID(expr);
+    std::string realName = metadata_.getNameFromAccessID(accessID);
+    expr->setName(realName);
+  }
+  void visit(const std::shared_ptr<FieldAccessExpr>& expr) override {
+    int accessID = iir::getAccessID(expr);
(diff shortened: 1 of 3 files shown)


Stagno:
launch jenkins

mroethlin (approved):
excellent, thanks

PR #1024, opened 2020-09-22
Separate header/interface from implementation

Stagno:
## Technical Description
Add options for C header and f90 interface outputting.
Substitute what was currently achieved via `cpp_bindgen` with the f90 interface.
(also sneaked in fix of `HAS_UNSTRUCTURED_DEPS` recurring problem, removed annoying warnings, added spack stuff to .gitignore and set compiler to error on missing returns)
### Testing

Via icondusk-e2e's https://github.com/dawn-ico/icondusk-e2e/pull/26


dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
```diff
@@ -896,22 +886,119 @@ std::string CudaIcoCodeGen::generateStencilInstantiation(
...
+
+  for(const auto& nameStencilCtxPair : context_) {
+    std::shared_ptr<iir::StencilInstantiation> stencilInstantiation = nameStencilCtxPair.second;
+    generateCHeaderSI(ssSW, stencilInstantiation);
+  }
 
   return ssSW.str();
 }
 
+inline void
```
mroethlin, line 930: why is this `inline`? I'd suggest `static` instead

dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
```diff
@@ -896,22 +886,119 @@ std::string CudaIcoCodeGen::generateStencilInstantiation(
...
+  std::vector<FortranInterfaceAPI> apis = {
+      FortranInterfaceAPI("run_" + stencilInstantiation->getName(),
+                          FortranInterfaceAPI::InterfaceType::DOUBLE),
+      FortranInterfaceAPI("run_" + stencilInstantiation->getName() + "_from_fort_host",
+                          FortranInterfaceAPI::InterfaceType::DOUBLE)};
+  for(auto&& api : apis) {
+    api.addArg("mesh", FortranInterfaceAPI::InterfaceType::OBJ);
+    api.addArg("k_size", FortranInterfaceAPI::InterfaceType::INTEGER);
+    for(auto field : support::orderMap(stencil.getFields())) {
+      int n = 3;
```
mroethlin, line 953: If you merge master you'll find that the field dimensions now offer a function to give you the spatial dimensions of a field. Maybe you can use this to simplify the code at hand

dawn/src/dawn/CodeGen/F90Util.h
```diff
@@ -0,0 +1,177 @@
...
+
+  std::stringstream& ss() { return ss_.get(); }
+
+  std::string str() { return ss().str(); }
+};
+
+class FortranInterfaceAPI {
+public:
+  enum class InterfaceType { INTEGER, FLOAT, DOUBLE, CHAR, OBJ };
+  FortranInterfaceAPI(std::string name, std::optional<InterfaceType> returnType = std::nullopt)
```
mroethlin, line 74: I don't understand why you need the option with no return type. I don't think you use it, and I don't quite get what would be generated if the return type is `nullopt`

mroethlin (changes requested):
Some Questions and Comments.
Other than that, during my testing I noticed that the interfaces are not generated for naive-ico. I think we should at least be able to generate the c-header in this case for testing over at the end to end repo, no?

dawn/src/dawn/CodeGen/CXXNaive-ico/CXXNaiveCodeGen.cpp
```diff
@@ -628,7 +628,10 @@ std::unique_ptr<TranslationUnit> CXXNaiveIcoCodeGen::generateCode() {
   ppDefines.push_back("#define DAWN_GENERATED 1");
   ppDefines.push_back("#undef DAWN_BACKEND_T");
   ppDefines.push_back("#define DAWN_BACKEND_T CXXNAIVEICO");
+  ppDefines.push_back(
+      "#define GRIDTOOLS_DAWN_NO_INCLUDE"); // Required to not include gridtools from math.hpp
```
Stagno, line 632: I get an include error because "math.hpp" includes "storage.hpp", which includes stuff from gridtools (and we don't depend on gridtools in the unstructured tests or in icondusk-e2e) ... apparently `GRIDTOOLS_DAWN_NO_INCLUDE` needs to be set

dawn/src/dawn/CodeGen/F90Util.h
Stagno: You are right, there's currently no use case. An API that wouldn't need a return type is a setter for globals
```
set_globalname(double value) { ... }
```
that would need to be generated into a `subroutine` and not a `function` (Fortran is very strange to me in this matter).
Do we keep it, just in case?

Stagno:
launch jenkins

mroethlin (approved):

Stagno:
launch jenkins

PR #1027, opened 2020-10-01
Add Horizontal Subdomains for Unstructured Domains

mroethlin:
## Technical Description

This PR introduces horizontal subdomains for the unstructured case. The design is quite simple; it simply "hijacks" the first `Interval` of the iteration space for the structured case, but redefines it's semantic. Instead of indices, magic numbers are assigned to the lower and upper level of the Interval, which can then be set at runtime using additionally introduced setters in the generated code. This works because in ICON, subdomains are contiguous _in memory_, and ICON stencils are only ever over a single subdomain, or over subdomains which are adjacent in memory. Consider this picture:

domains

A ICON stencil may for example run over lateral boundary, the nudging zone, and the interior, but not over lateral boundary + halo. This entails that the contract for the `Interval` does not need to be adapted for the unstructured case and remains meaningful. 

### ToDo

- [x] namespace issues
- [x] Introduce checker that asserts that second interval of an unstructured iteration space is always `nullopt`?

### Testing

New tests have been added to 
* dawn4py integration test 
* unstructured integration test 

### Dependencies

This PR is independent

### DIFF
--- a/dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
+++ b/dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
@@ -27,11 +27,11 @@ static std::string nbhChainToVectorString(const std::vector<dawn::ast::LocationT
   auto getLocationTypeString = [](dawn::ast::LocationType type) {
     switch(type) {
     case dawn::ast::LocationType::Cells:
-      return "dawn::LocationType::Cells";
+      return "::dawn::LocationType::Cells";
     case dawn::ast::LocationType::Edges:
-      return "dawn::LocationType::Edges";
+      return "::dawn::LocationType::Edges";
     case dawn::ast::LocationType::Vertices:
-      return "dawn::LocationType::Vertices";
+      return "::dawn::LocationType::Vertices";
     default:
       dawn_unreachable("unknown location type");
       return "";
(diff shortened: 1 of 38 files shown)


mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (changes requested):
Enumerating in the generated code all the possible markers (`UnstructuredIterationSpace`) conflicts with the frontend design choice to be able to define your own markers. I see two philosophies here, we should maybe have a discussion with the team:
- "LateralBoundary, Nudging, Interior, Halo, End" are all the possible markers that you'll ever need when coding an unstructured model and they are not only specific to ICON (btw `Halo` should be removed once we move entirely halo exchanges from the fortran driver into the DSL world). Dusk should expose to the user only the above markers and map them internally to Dawn's magic numbers. In this case, the user shouldn't be able to set his/her own markers (no definitions.py), as they are already defined as DSL keywords.
- There could be more (less) than just those 5. The DSL user decides.
	  Implementation is more complex. Frontend lets the user define markers, which are string tags to be passed to dawn4py. In order to retain the dawn implementation through `Interval` (which mandates magic integers), a map from string tag to magic integer should be added to the metadata, so that, on both the frontend and backend interfaces, Dawn exposes meaningful string tags instead of integers.

Example dawn4py API:
```
# order imposes magic-numbering (LateralBoundary=1, Nudging1=2, ...)
markers = dawn4py.make_markers("LateralBoundary", "Nudging1", "Nudging2", "Interior", "End") 
myInterval = dawn4py.make_interval(markers, "LateralBoundary", "Nudging2")
```
Example gen. code API:
```
void set_marker_index(LocationType loc, std::string marker, int offset, int index);
```

mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (approved):

PR #1035, opened 2020-10-07
Properly Version Sparse Fields

### LINKED ISSUE #1032: Versioning of Sparse Fields
This dusk code
```
@stencil
def various_expression(
    edge_3d_field1: Field[Edge, K],
    sparse_3d_field1: Field[Edge > Cell > Vertex > Edge, K],
):
    with levels_upward:
        edge_3d_field1 = min_over(Edge > Cell > Vertex > Edge, sparse_3d_field1)
        with sparse[Edge > Cell > Vertex > Edge]:
            sparse_3d_field1 = edge_3d_field1[Edge]
```
leads to two stages
```
sparse_3d_field1_0[<no_horizontal_offset>,0] = sparse_3d_field1[<no_horizontal_offset>,0]
```
and
```
edge_3d_field1[<no_horizontal_offset>,0] = Reduce (min, init = double_type 1.79769313486231571e+308, location = {Edge, Cell, Vertex, Edge}): sparse_3d_field1_0[<has_horizontal_offset>,0];",
for ({Edges, Cell, Vertices, Edges) {
	sparse_3d_field1[<has_horizontal_offset>,0] = edge_3d_field1[<no_horizontal_offset>,0];
}
```

Which trips up the `UnstructuredDimensionsChecker` in the first stage (since assigning to sparse fields is only allowed in `ForLoopStmt`s. Make sure that dawn inserts a proper `ForLoopStmt` in this case.

mroethlin:
## Technical Description

This dusk code
```
@stencil
def various_expression(
    edge_3d_field1: Field[Edge, K],
    sparse_3d_field1: Field[Edge > Cell > Vertex > Edge, K],
):
    with levels_upward:
        edge_3d_field1 = min_over(Edge > Cell > Vertex > Edge, sparse_3d_field1)
        with sparse[Edge > Cell > Vertex > Edge]:
            sparse_3d_field1 = edge_3d_field1[Edge]
```
leads to two stages
```
sparse_3d_field1_0[<no_horizontal_offset>,0] = sparse_3d_field1[<no_horizontal_offset>,0]
```
and
```
edge_3d_field1[<no_horizontal_offset>,0] = Reduce (min, init = double_type 1.79769313486231571e+308, location = {Edge, Cell, Vertex, Edge}): sparse_3d_field1_0[<has_horizontal_offset>,0];",
for ({Edges, Cell, Vertices, Edges) {
	sparse_3d_field1[<has_horizontal_offset>,0] = edge_3d_field1[<no_horizontal_offset>,0];
}
```

Which trips up the `UnstructuredDimensionsChecker` in the first stage (since assigning to sparse fields is only allowed in `ForLoopStmt`s. This PR makes sure that dawn inserts a proper `ForLoopStmt` in this case. 

### Resolves / Enhances

Fixes #1032

### Testing

A new test added to `TestPassFieldVersioning` in the style of `TestPassRemoveScalars`.

### Dependencies

This PR is independent.

### DIFF
--- a/dawn/test/unit-test/dawn/Optimizer/TestPassFieldVersioning.cpp
+++ b/dawn/test/unit-test/dawn/Optimizer/TestPassFieldVersioning.cpp
@@ -11,10 +11,17 @@
 //  See LICENSE.txt for details.
 //
 //===------------------------------------------------------------------------------------------===//
+#include "dawn/IIR/ASTFwd.h"
+#include "dawn/IIR/ASTStmt.h"
 #include "dawn/IIR/IIR.h"
+#include "dawn/IIR/IIRNodeIterator.h"
 #include "dawn/IIR/StencilInstantiation.h"
 #include "dawn/Optimizer/PassFieldVersioning.h"
+#include "dawn/Optimizer/PassFixVersionedInputFields.h"
 #include "dawn/Serialization/IIRSerializer.h"
+#include "dawn/Unittest/IIRBuilder.h"
+#include "dawn/Unittest/UnittestUtils.h"
+#include "dawn/Validator/UnstructuredDimensionChecker.h"
 
 #include <gtest/gtest.h>
 #include <memory>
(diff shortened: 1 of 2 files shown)


mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (commented):
The field versioning and stage split shouldn't have been triggered in the first place, because there's no race condition.
Consider this simplified (omitting k, using Edge > Cell > Edge instead of the longer one) code generation of the stencil:
```
// |Edge > Cell > Edge| = 4
for (int edgeIdx : edges()) {
	edge_3d_field1(edgeIdx) = +inf;
	for(int neigh = 0; neigh < 4; ++neigh) {
		edge_3d_field1(edgeIdx) = std::min(edge_3d_field1(edgeIdx), sparse_3d_field1(edgeIdx, neigh));
	}

	for(int neigh = 0; neigh < 4; ++ neigh) {
		sparse_3d_field1(edgeIdx, neigh) = edge_3d_field1(edgeIdx);
	}
}
```

mroethlin:
> The field versioning and stage split shouldn't have been triggered in the first place, because there's no race condition.
> Consider this simplified (omitting k, using Edge > Cell > Edge instead of the longer one) code generation of the stencil:
> 
> ```
> // |Edge > Cell > Edge| = 4
> for (int edgeIdx : edges()) {
> 	edge_3d_field1(edgeIdx) = +inf;
> 	for(int neigh = 0; neigh < 4; ++neigh) {
> 		edge_3d_field1(edgeIdx) = std::min(edge_3d_field1(edgeIdx), sparse_3d_field1(edgeIdx, neigh));
> 	}
> 
> 	for(int neigh = 0; neigh < 4; ++ neigh) {
> 		sparse_3d_field1(edgeIdx, neigh) = edge_3d_field1(edgeIdx);
> 	}
> }
> ```

you are right about that, it should only have been triggered if the second statement would read
```
        with sparse[Edge > Cell > Vertex > Edge]:
            sparse_3d_field1 = edge_3d_field1[Edge > Cell > Vertex > Edge]
```

unfortunately, as discussed yesterday, simply adding an "is offset true" check to the dependency analysis is not enough since it would lead to additional false positives:

* non-offset (in the gen. code dense access index is pidx)
```
dense_field_edges = reduce(
    Edges > Cells,
    sparse_field_e_c
  )
```
* offset (in the gen. code dense access index is nbhIdx)
```
dense_field_edges = reduce(
  Edges > Verts > Edges,
  reduce(
    Edges > Cells,
    sparse_field_e_c
  )
)
```
so potential race conditions can only occur in the second, but not in the first. This would entail that we also properly set and respect the horizontal offset flag in reductions (which would also need to happen starting from dusk). 

I'd argue that this PR is still worthwhile. Not every fill stage of a sparse field is a "false positive", so this needs to work. This PR does exactly that.

dawn/test/unit-test/dawn/Optimizer/TestPassFieldVersioning.cpp
```diff
@@ -199,4 +206,71 @@ TEST_F(TestPassFieldVersioning, VersioningTest7) {
...
+          b.stage(b.doMethod(
+              dawn::sir::Interval::Start, dawn::sir::Interval::End,
+              b.stmt(b.assignExpr(b.at(dense),
+                                  b.reduceOverNeighborExpr(Op::plus, b.at(sparse), b.lit(0.),
+                                                           {LocType::Edges, LocType::Cells,
+                                                            LocType::Vertices, LocType::Edges}))),
+              b.loopStmtChain(
+                  b.stmt(b.assignExpr(b.at(sparse), b.at(dense, HOffsetType::noOffset, 0),
+                                      Op::assign)),
+                  {LocType::Edges, LocType::Cells, LocType::Vertices, LocType::Edges}))))));
```
Stagno, line 235: This test is currently testing a transformation that shouldn't happen. When field versioning of sparse fields will be fixed this should maybe become a negative test. I suggest commenting it out and have the following positive test instead:
```
dense_field_edges = reduce(
  Edges > Verts > Edges,
  reduce(
    Edges > Cells,
    sparse_field_e_c
  )
)
with sparse[Edges > Cells]:
  sparse_field_e_c = dense_field_edges[Edges]
```
(much can be reused from the code below i think)

Stagno (changes requested):
I'm also now convinced that the fix implemented here is still useful, however i suggest changing the test so that it won't become obsolete soon(-ish).

mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (approved):

Issue #1056, opened 2020-11-05
field versioning bug


cosunae:
With the example below, the field versioning triggers a bug. 
There is a circular dependency with
hC <-, -> hC_t 

which does not trigger a field versioning by the algorithm since the SCC does not contain horizontal extent accesses. 
However the algorithm also checks that the graph contains (at least one) input only nodes and output only nodes. 
https://github.com/MeteoSwiss-APN/dawn/blob/30ba6f1e840471dc927be1774608b1649eafc251/dawn/src/dawn/Optimizer/PassFieldVersioning.cpp#L226

When the algorithm inserts (in the backward traversal of statements) the statement,
```hE = hC[i + 1] * (1 - alpha) + hC[i - 1] * alpha;```

the condition that contains (at least one) input only node does not hold true. The algorithm thinks this is due to the SCC of hC <-, -> hC_t, which is not true, since this SCC was inserted before.

Full example: 

```
#include "gtclang_dsl_defs/gtclang_dsl.hpp"
using namespace gtclang::dsl;
 
stencil lap {
  storage_ij hinitC;
  storage_ij hC;
  storage_ij hC_t;
  storage_ij vinitC;
  storage_ij vC;
  storage_ij vC_t;
  storage_ij uinitC;
  storage_ij uC;
  storage_ij uC_t;
  storage_ij hC_x;
  storage_ij hC_y;
  storage_ij uvC_div;
  storage_ij hE;
  storage_ij vE;
  storage_ij uE;
  storage_ij nx;
  storage_ij ny;
  storage_ij L;
  storage_ij alpha;
  storage_ij boundary_edges;
  storage_ij boundary_cells;
  storage_ij A;
  storage_ij edge_orientation;
  storage_ij dt;
  storage_ij Grav;
  Do {
    vertical_region(k_start, k_end) {
      hinitC = hC;
      uinitC = uC;
      vinitC = vC;
 
      // predict
      hC = hinitC + 0.5 * dt * hC_t;
      uC = uinitC + 0.5 * dt * uC_t;
      vC = vinitC + 0.5 * dt * vC_t;
 
      // lerp cell quantities to edges
      hE = hC[i + 1] * (1 - alpha) + hC[i - 1] * alpha;
      uE = uC[i + 1] * (1 - alpha) + uC[i - 1] * alpha;
      vE = vC[i + 1] * (1 - alpha) + vC[i - 1] * alpha;
 
      // height field gradient
[...]

PR #1057, opened 2020-11-05
Fix fieldversion

cosunae:
## Technical Description

Fixes the field version for #1056 
Until now the field versioning was only detecting and resolving cycles that contain horizontal access extents. 
This PR also applies field versioning to pointwise dependencies.


cosunae:
launch jenkins

cosunae:
launch jenkins

cosunae:
launch jenkins

cosunae:
launch jenkins

PR #1059, opened 2020-11-10
Fix fieldversion2

cosunae:
## Technical Description

Describe here what the PR does

### Resolves / Enhances

Put the link to the issue here
(e.g resolves #XXX)

### Example

If a specific example is being fixed, put it here

### Testing

If anything besides a simple launch needed (perf-test), link it here

### Dependencies

If this is built on top of anything else, put it here

### DIFF
--- a/dawn/src/dawn/IIR/DependencyGraph.h
+++ b/dawn/src/dawn/IIR/DependencyGraph.h
@@ -68,7 +68,9 @@ public:
   };
 
 protected:
+  // map of Value (i.e. normally accessID to Vertex object
   std::unordered_map<int, Vertex> vertices_;
+  // adjacencyList for each vertex where the position within the vector is the vertexID
   std::vector<EdgeList> adjacencyList_;
 
 public:
@@ -96,8 +98,8 @@ public:
   DependencyGraph() = default;
 
   /// @brief Insert a new node
-  Vertex& insertNode(int ID) {
-    auto [iter, inserted] = vertices_.emplace(ID, Vertex{adjacencyList_.size(), ID});
+  Vertex& insertNode(int Value) {
+    auto [iter, inserted] = vertices_.emplace(Value, Vertex{adjacencyList_.size(), Value});
     if(inserted)
       adjacencyList_.push_back(EdgeList());
     return iter->second;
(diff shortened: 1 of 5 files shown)


cosunae:
launch jenkins

cosunae:
launch jenkins

dawn/test/integration-test/dawn4py-tests/data/generate_versioned_field_ref.cpp
```diff
@@ -37,18 +36,14 @@ class generate_versioned_field {
...
 {
     for(int k = 0+0; k <= ( m_k_size == 0 ? 0 : (m_k_size - 1)) + 0+0; ++k) {
       for(auto const& loc : getEdges(LibTag{}, m_mesh)) {
-m_c_0(deref(LibTag{}, loc), (k + 0)) = m_c(deref(LibTag{}, loc), (k + 0));
-      }    }}{
-    for(int k = 0+0; k <= ( m_k_size == 0 ? 0 : (m_k_size - 1)) + 0+0; ++k) {
-      for(auto const& loc : getEdges(LibTag{}, m_mesh)) {
-m_a(deref(LibTag{}, loc), (k + 0)) = ((m_b(deref(LibTag{}, loc), (k + 0)) / m_c_0(deref(LibTag{}, loc), (k + 0))) + (::dawn::float_type) 5);
+m_a(deref(LibTag{}, loc), (k + 0)) = ((m_b(deref(LibTag{}, loc), (k + 0)) / m_c(deref(LibTag{}, loc), (k + 0))) + (::dawn::float_type) 5);
+      }      for(auto const& loc : getEdges(LibTag{}, m_mesh)) {
```
Stagno, line 46: This test can be removed then. It was testing that a versioned field was generated, but it's not generated anymore.

Stagno (changes requested):
Small changes requested, overall it's ok.

Do we understand now how we could get into this unresolvable race condition, though?
https://github.com/MeteoSwiss-APN/dawn/blob/44770e627cfd09ed9ff511345249d6f8a8aef997/dawn/src/dawn/Optimizer/PassFieldVersioning.cpp#L240-L249
I'm still doubtful about that, because dawn will bail out, but a correctly placed kernel-split can always solve a race condition. If it's not possible anymore to reach that condition, then it's dead code.

Stagno:
launch jenkins

PR #1060, opened 2020-11-11
Fix include in IcoChainSizes.cpp

Stagno:
## Technical Description

Missing include to `stdexcept` in IcoChainSizes.cpp making gcc 10.2 fail to compile.
Thanks to @JoergBehrens for reporting this.


Stagno:
launch jenkins

PR #1071, opened 2020-11-30
Cuda-ICO Backend Assumed that every API Field is a Stencil Field

### LINKED ISSUE #1068: Unused API Fields cause CUDA backend to fail
If a API field is unused like so 

```
@stencil
def sparse_reduction(
    a: Field[Edge], b: Field[Edge], UNUSED: Field[Edge],
):
    with levels_downward:
        a = b
```

The cuda backend will fail. This is because the CUDA backend assumes that all API fields are also stencil fields [here](https://github.com/MeteoSwiss-APN/dawn/blob/a861f9eb8e843521508035c614f1ab44abe26fc0/dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp#L566). It does this to figure out if a fields is an out or "inout" field to determine whether it should include the field in the copy back functions. Fixing this is probably as simple as neglecting API fields that aren't stencil fields in the copy back functions.

mroethlin:
## Technical Description

The Cuda ICO backend assumed that every API field is also a stencil field. It did this while trying to figure out which fields are being written to in order to not clutter the functions copying fields back to the host. This is now fixed.

### Resolves / Enhances

Fixes #1068 

### Example

Examples like:
```
@stencil
def vert_offset_write(
    a: Field[Edge,K], b: Field[Edge,K], unused: Field[Edge, K]):
    with levels_downward as k:
        a = b
```
should not break at cuda code gen time anymore.

### Testing

Unfortunately can not be tested in dawn because we do not support compiling of generated cuda code in the current testing. Manually tested in icondusk-e2e.

### Dependencies

This PR is independent.

### DIFF
--- a/dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
+++ b/dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
@@ -505,8 +505,17 @@ void CudaIcoCodeGen::generateCopyMemoryFun(MemberFunction& copyFun,
                                            const iir::Stencil& stencil) const {
 
   const auto& APIFields = stencil.getMetadata().getAPIFields();
+  const auto& fields = stencil.getOrderedFields();
+
+  for(auto field : fields) {
+    auto fname = field.second.Name;
+    copyFun.addArg("::dawn::float_type* " + fname);
+  }
 
   for(auto fieldID : APIFields) {
+    if(fields.count(fieldID) == 0) {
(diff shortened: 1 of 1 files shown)


mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
NOTE: CI Passed

Stagno:
launch jenkins

Stagno (approved):
Looks good to go.

PR #1076, opened 2020-12-01
Ensure Location Type Consistency in IfStmt

### LINKED ISSUE #1064: Type Checking Bug in If Condition
The following code snipped is accepted even though it shouldn't. Produces dangerous code, which, in the best case, would segfault.

```
@stencil
def if_condition(hC_x: Field[Cell],
                boundary_edges: Field[Edge]):
    with levels_downward:
        if boundary_edges:
            hC_x = 0
```

mroethlin:
## Technical Description

So far we failed to ensure location type consistency in between the conditional, then and else branch of of an `IfStmt`. This allowed code like:
```
@stencil
def if_condition(hC_x: Field[Cell],
                boundary_edges: Field[Edge]):
    with levels_downward:
        if boundary_edges:
            hC_x = 0
```
To go through the pipeline, being generated into potentially unsafe code. 

### Resolves / Enhances

Fixes #1064 

### Example

See above

### Testing

Death- and positive tests added to `TestUnstructuredDimensionsChecker` with simple as well as nested if statements 

### Dependencies

This PR is independent.

### DIFF
--- a/dawn/src/dawn/SIR/SIR.cpp
+++ b/dawn/src/dawn/SIR/SIR.cpp
@@ -582,6 +582,7 @@ UnstructuredFieldDimension::UnstructuredFieldDimension(const ast::NeighborChain 
 }
 
 const ast::NeighborChain& UnstructuredFieldDimension::getNeighborChain() const {
+  DAWN_ASSERT(isSparse());
   return neighborChain_;
 }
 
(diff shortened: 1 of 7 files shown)


mroethlin:
launch jenkins

mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (changes requested):
Suggesting a different way to implement the check that covers more cases, see comment.

Stagno:
launch jenkins

Stagno (approved):

Issue #508, opened 2019-12-02
Formalizing math functions in SIR


Stagno:
gtclang DSL exposes basic math functions such as `fabs`, `min`, `ceil`, `exp`, ... to be called from within a stencil. Calls to these are currently represented in the SIR as `FunCallExpr`s with the `callee_` string set to `gridtools::dawn::math::<FUNCTION_NAME>`. Their implementations are backend dependent, e.g.:
https://github.com/MeteoSwiss-APN/dawn/blob/fa22a7df5c34182121517a0a4b7ed6bb4e1bad5c/dawn/src/driver-includes/math.hpp#L131-L136
This is a bit hacky as it is just a string and it is though not formalized in dawn. Currently the frontend is prefixing the function call with the namespace of the implementation (`gridtools::dawn`), which breaks the separation between frontend and backend.
We could make these math functions explicit in the AST (through an enum instead of a string for example), meaning that they would be recognizable from CodeGen and can be translated then to their backend specific implementations.

egparedes:
After discussions on Slack, it's been decided to implement the following set of functions:

- max,min
- pow, sqrt
- exp, log
- sin, cos, tan, arcsin, arccos, arctan
- abs, floor, ceil
- isinf, isnan

egparedes:
The following issues are likely duplicates of this one and could be removed:

https://github.com/MeteoSwiss-APN/dawn/issues/981
https://github.com/MeteoSwiss-APN/dawn/issues/981

PR #1078, opened 2020-12-02
Implement gt4py Math Standard

### LINKED ISSUE #1072: `sin`, `cos` not exposed in math driver include
`sin` and `cos` are missing here: https://github.com/MeteoSwiss-APN/dawn/blob/master/dawn/src/driver-includes/math.hpp

There may be more missing math functions. We never really implemented the _standardized math functions_...

### LINKED ISSUE #1073: `pow` only works if both arguments have the same type
See: https://github.com/MeteoSwiss-APN/dawn/blob/a861f9eb8e843521508035c614f1ab44abe26fc0/dawn/src/driver-includes/math.hpp#L99-L110
Can lead to errors like:

```bash
intro_cxx-naive.cpp:37:89: error: no matching function for call to ‘pow(dawn::float_type, int)’
   37 |  gridtools::dawn::math::pow((::dawn::float_type) 2.0, (int) 1);
```

C++'s `pow` is defined for this case:
https://en.cppreference.com/w/cpp/numeric/math/pow

mroethlin:
## Technical Description

Implements the math functions defined to be "standard" in gt4py land, c.f. [here](https://github.com/GridTools/gt4py/blob/ceb2b02669f9cc535a7d0d48e3810cc85d6a8e4d/src/gt4py/backend/gt_backends.py#L184). 

### Resolves / Enhances

Fixes #1072 
Fixes #1073 

### Testing

Tested manually using dusk

### Dependencies

This PR is independent.

### DIFF
--- a/dawn/src/driver-includes/math.hpp
+++ b/dawn/src/driver-includes/math.hpp
@@ -102,10 +102,10 @@ GT_FUNCTION T fmod(const T x, const T y) {
  * @param x Base value
  * @param y Exponent value
  *
- * @see http://en.cppreference.com/w/cpp/numeric/math/fmod
+ * @see https://en.cppreference.com/w/cpp/numeric/math/pow
  */
-template <typename T>
-GT_FUNCTION T pow(const T x, const T y) {
+template <typename T, typename U>
+GT_FUNCTION T pow(const T x, const U y) {
   return ::pow(x, y);
 }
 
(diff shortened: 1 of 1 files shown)


mroethlin:
launch jenkins

BenWeber42 (approved):
Looks good!

Only thing I noticed is that you used `asin`, `acos` and `atan`. In the original issue #508 we called them `arcsin`, `arccos`, `arctan`. On the other hand, in C++ your naming is used (https://en.cppreference.com/w/cpp/numeric/math).

So I think either is fine. I kinda prefer the shorter version, since it's pretty clear what is meant anyways. But if you think `arcsin` et.al. are better, feel free to change it.

mroethlin:
launch jenkins

PR #1079, opened 2020-12-04
Empty stencils

### LINKED ISSUE #1074: C++ naive codegen doesn't support empty stencils
```python
@stencil
def simple_example_stencil(a: Field[Edge]):
    with levels_upward:
        pass
```

leads to

```bash
intro_cxx-naive.cpp:27:12: error: expected ‘;’ at end of member declaration
   27 |   void run() ) {}
      |            ^
      |             ;
intro_cxx-naive.cpp:27:14: error: expected unqualified-id before ‘)’ token
   27 |   void run() ) {}
      |              ^
```

Relevant generated code:

```C++
  simple_example_stencil(const ::dawn::mesh_t<LibTag> &mesh, int k_size,
                         ::dawn::edge_field_t<LibTag, ::dawn::float_type> &a) {}

  void run() ) {}
};
```

mroethlin:
## Technical Description

The _statements_ in stencil should not affect the API of the stencil. This should even be true if the stencil is completely empty, i.e. consists of the `pass` statement only. So far this was not true:

* The ico-naive emitted invalid c++ code
* The cuda backend did not emit valid API functions 

This PR addresses both of these shortcomings (even though the issue only mentions the naive backend) 

### Resolves / Enhances

Fixes #1074 

### Example

```
@stencil
def empty(a: Field[Edge]):
    with levels_downward:
        pass
```

### Testing

I don't think its possible to test this from within dawn? 

### Dependencies

This PR is independent


mroethlin:
launch jenkins

dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
```diff
@@ -766,12 +766,50 @@ void CudaIcoCodeGen::generateAllAPIRunFunctions(
...
+        apiRunFun->commit();
+      }
+
+      for(const auto& stream : apiRunFunStreams) {
+        ssSW << stream.str() << ";\n";
+      }
+      return;
+    }
+  }
+
```
Stagno, line 807: Looks like it's almost a code clone of the non-empty stencil case, have you considered making the empty/non-empty distinction below? Maybe adding at line 875 something like
```
if(stencils.empty()) {
   apiRunFun->startBody();
   apiRunFun->addStatement("return 0.");
}
```
and letting the rest ( https://github.com/MeteoSwiss-APN/dawn/blob/fda39003acb1287624b12b49dcc9784bc5326f2e/dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp#L876-L900 ) be in the else block.

Stagno (changes requested):
Simplification suggested, otherwise looks fine.
About the testing issue: it could be tested in `icondusk-e2e` starting from an empty dusk stencil and checking that `0` is returned.

dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
mroethlin: I first wanted to avoid that, because the lines you are indicating are already inside a loop over the stencils (which is pointless of course since there is only one). I initially thought that the refactoring due to accesses to `stencil` would be quite involved, but looking at it again it turned out easier than expected.

mroethlin:
launch jenkins

Stagno (approved):

PR #1085, opened 2020-12-14
Fix rank calculation in F90 interface generator

Stagno:
## Technical Description

As the title says, before it was not counting the sparse dimension. Rank computation moved inside `FieldDimensions`.

### Testing
Testing through icondusk-e2e PR: https://github.com/dawn-ico/icondusk-e2e/pull/34


Stagno:
launch jenkins

mroethlin (approved):
good catch!

PR #1086, opened 2020-12-15
Remove Python SIR Printer

mroethlin:
## Technical Description

There is a full SIR printer implemented in python that is not really used (except in a test). This PR removes that printer since it is not needed.


mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (approved):
gtg!

PR #1088, opened 2021-01-05
Make Neighbor Lists Including Center Contain Center First

mroethlin:
## Technical Description

This PR fixes a bug in the unstructured interface where the center element (if included) could reside in any position in the returned neighbor list. However, we rely on the fact that this is always returned first (also in ICON).

### Testing

Tested in `icondusk-e2e`

### Dependencies

This PR is independent.


mroethlin:
launch jenkins

Stagno (approved):
Looks good

mroethlin:
launch jenkins

PR #1091, opened 2021-01-19
Explode globals

mroethlin:
## Technical Description

In order to not be required to expose the `global` struct also in the external API the global variables need to be exploded in all API functions, not just the ones to FORTRAN


mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (approved):

Issue #617, opened 2020-01-20
Refactor the ASTVisitor


mroethlin:
Currently, there are the following interfaces offered for writing an AST Visitor:

`ASTVisitor::visit(const std::shared_ptr<T>& stmt);`
And
`ASTVisitorNonConst::visit(std::shared_ptr<T>& stmt);`

Presumably, the intention was to offer both an interface for queries on the AST, leaving the Stmt's and Expr's constant, and one for transforming the AST. The above interface does not provide that, the `const`in the latter interface refers to the `shared_ptr`, not its content. Thus, refactor the AST Visitors to actually provide the semantics described, e.g. using:

`ASTVisitor::visit(const T& stmt);`
And
`ASTVisitorNonConst::visit(std::shared_ptr<T>& stmt);`

There is yet another Issue with the Forwarding Visitor noticed during the generalization of the weights: The forwarding visitor does not correctly forward into the stencil body of a function. Which, however, is implied by the name. See `WeightChecker` or `PassTemporaryFirstAccess` to see how this can be achieved.

havogt:
Probably `ASTVisitorNonConst::visit(T& stmt);` as well. I guess the visitor newer takes ownership?

PanicSheep:
How about renaming them _NodeVisitor_ and _NodeTransformer_?
The name "transformer" implies that it changes the nodes.
The "visitor" might need comments to state that it doesn't change the nodes.
(I took the idea from python's _ast.NodeVisitor_ and _ast.NodeTransformer_)

PR #1094, opened 2021-01-26
Made const visitor enforce constness of inner type of shared_ptr<T>

muellch:
## Technical Description

As explained in #617, the current visitor interface
```
ASTVisitor::visit(const std::shared_ptr<T>& stmt);
ASTVisitorNonConst::visit(std::shared_ptr<T> stmt);
```
does not respect the constness of the inner type `T` wrapped by the `shared_ptr`.
The proposed solution is to switch to the following interface:
```
ASTVisitor::visit(const T& stmt);
ASTVisitorNonConst::visit(T& stmt);
```
Unfortunately, this would lead to a massive global refactoring, since the visitors are also used to pass around `shared_ptr` from datastructure to datastructure, for example all of the StencilMetaInformation is generated by visiting the AST and generating `shared_ptr`s through the visitor interface.
Instead I propose to change to this interface:
```
ASTVisitor::visit(const std::shared_ptr<const T>& stmt);
ASTVisitorNonConst::visit(const std::shared_ptr<T>& stmt);
```
Note that `ASTVisitorNonConst` now provides the exact same interface as `ASTVisitor` before the changes. Meanwhile `ASTVisitor` now respects the constness of the inner type wrapped by `shared_ptr`.

### Resolves / Enhances

#617

### Example

As a first example, the `ASTStringifier` is now visited by the const visitor. All other visitors in the project now use the `ASTVisitorNonConst` interface.

### Followups

1. Only `ASTStringifier` uses the new `const` visitor interface. Most other visitor usages probably could use the `const` interfaces. However, some work is still to be done here since some methods, which probably could be `const`, do not respect constness, such as `Expr::getData` or `IIR::getAccessID`.
2. Since constness in C++ is a shallow property (i.e. does not propagate downwards through nested datastructures), even with the new const visitor interface some class members of AST classes can still be changed. For example, for `AST::BinaryOperator' the following operation is now prevented by the `const` visitor:
```
void visit(const std::shared_ptr<const BinaryOperator>& expr) override {
    expr->getLeft() = expr->getRight();
```
The following unfortunately still compiles:
```
void visit(const std::shared_ptr<const BinaryOperator>& expr) override {
    *(expr->getLeft()) = *(expr->getRight());
```
One possible way to prevent this behavior is using `std::experimental::propagate_const`. We can evaluate if this is worth it. It has to work with gcc 8 since that is our target compiler.

### DIFF
--- a/dawn/src/dawn/AST/ASTExpr.cpp
+++ b/dawn/src/dawn/AST/ASTExpr.cpp
@@ -397,7 +397,7 @@ void FieldAccessExpr::setPureOffset(const Offsets& offset) {
   argumentOffset_ = Array3i{{0, 0, 0}};
 }
 
-ArrayRef<std::shared_ptr<Expr>> FieldAccessExpr::getChildren() {
+ArrayRef<std::shared_ptr<Expr>> FieldAccessExpr::getChildren() const {
   if(offset_.hasVerticalIndirection()) {
     return ExprRangeType(offset_.getVerticalIndirectionFieldAsExpr());
   }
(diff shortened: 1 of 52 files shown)


muellch:
launch jenkins

mroethlin (approved):
lgtm

PR #1095, opened 2021-01-28
Offset Reduction

mroethlin:
## Technical Description

A rough implementation of the work around described in the [google doc about path dependency](https://docs.google.com/document/d/1nQFnInOmbhQWpHnNrlO4P9jezwIxlsstnACRdWQNbzg/edit). The implementation is complete in the sense that there is SIR, IIR representation of the offset, dawn4py hooks and code generation in both backends. However, no safety is provided:

* There are no facilities to check that the right amount of offsets was passed. Passing too many offsets will lead to ignored values, passing too few will lead to a segfault (in the best case).
* There is no type checking for the sparse field being offset.


mroethlin:
to be reopned from origin branch

Issue #438, opened 2019-10-31
Move statements.proto from SIR to AST


Stagno:
`statements.proto` currently is within SIR directory, but it belongs to AST.
It should be moved and related cmake code should be adapted.

Issue #319, opened 2019-10-09
Remove AST -> SIR dependency


Stagno:
AST component depends on SIR through `sir::Field` and `sir::VerticalRegion`. AST should be upstream w.r.t. SIR.

- `sir::Field` should be moved to AST
- `sir::VerticalRegion` should be moved into the SIR data container of `ast::VerticalRegionDeclStmt`. Member field `sir::VerticalRegion::Ast` should be moved into `ast::VerticalRegionDeclStmt`.

lukasm91:
This would fix remaining odd dependencies. I think it would be a good idea to take this soon.

PR #1096, opened 2021-02-03
SIR <-> AST refactoring part 2

muellch:
## Technical Description

Further disentangled `AST` and `SIR` by moving `statements.proto` and `enums.proto` to the `AST`. Furthermore the `Fielddimension` class was also moved from `SIR` to `AST`.

### Resolves / Enhances

https://github.com/MeteoSwiss-APN/dawn/issues/319
https://github.com/MeteoSwiss-APN/dawn/issues/438

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ _local_/*
 .vscode/
 **/compile_commands.json
 .clangd/
+venv/
 
 # Ignore python editable installs
 *.egg-info/
--- a/dawn/examples/python/copy_stencil.py
+++ b/dawn/examples/python/copy_stencil.py
@@ -33,7 +33,7 @@ import os
 
 import dawn4py
 from dawn4py.serialization import SIR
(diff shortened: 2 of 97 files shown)


muellch:
launch jenkins

muellch:
launch jenkins

.gitignore
```diff
@@ -21,6 +21,7 @@ _local_/*
 .vscode/
 **/compile_commands.json
 .clangd/
+venv/
```
Stagno, line 24:
```suggestion
```
I suggest instead to place your python venv inside the build directory

Stagno (changes requested):
Looks good to me, one tiny change to go.

.gitignore
muellch: I put that there intentionally, because my IDE requires it for python coding.
Is it ok if we leave it in?

PR #1100, opened 2021-02-05
Fixed two bugs in codegen, one related to intervals, one to max_over (#1097 and #1098)

### LINKED ISSUE #1097: Translation of max_over doesn't work
SIRs that contain this:
```
"reductionOverNeighborExpr": {
                                              "op": "max",
                                              ...
```
get wrongly translated to something like:
```
lhs_49 max = <RHS>;
```
instead of
```
lhs_49 = max(lhs_49, <RHS>);
```
in cuda-ico backend.
Same problem if instead of "op": "max" one uses "gridtools::dawn::math::max" (used for max outside of reductions).
This problem makes translation of the `max_over` Dusk syntax incorrect.

### LINKED ISSUE #1098: Cuda-ico vertical interval wrong generation
If one specifies this vertical interval in SIR:
```
"interval": {
                      "specialLowerLevel": "End",
                      "specialUpperLevel": "End",
                      "lowerOffset": -1
                    },
```
(`with domain.upwards[-1:]` in Dusk)
cuda-ico generates the following wrong translation:
```
int klo = kidx * LEVELS_PER_THREAD + 1048575;
int khi = (kidx + 1) * LEVELS_PER_THREAD + 1048575;
```
instead of
```
int klo = kidx * LEVELS_PER_THREAD + (kSize-1);
int khi = (kidx + 1) * LEVELS_PER_THREAD + (kSize-1);
```

muellch:
## Technical Description

Fixes the bugs in the following two tasks:
Fixes #1097 : Now explicitly check if the `lowerLevel` and `upperLevel` of the interval are both `End`.
Fixes #1098 : Now branch with `sum_over` being the special case and `max_over`, `min_over` behaving the same way.

### DIFF
--- a/dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
+++ b/dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
@@ -376,13 +376,20 @@ void ASTStencilBody::visit(const std::shared_ptr<ast::ReductionOverNeighborExpr>
                "lhs, auto "
         << ASTStencilBody::ReductionIndexVarName(reductionDepth_ + 1)
         << ", auto const& weight) mutable {\n";
-    ss_ << "lhs " << expr->getOp() << "= ";
-    ss_ << "weight * ";
   } else {
     ss_ << ", [&, " + ASTStencilBody::ReductionSparseIndexVarName(reductionDepth_) +
                " = int(0)](auto& lhs, auto "
         << ASTStencilBody::ReductionIndexVarName(reductionDepth_ + 1) << ") mutable { ";
+  }
+
+  if(expr->getOp() != "+") {
+    ss_ << "lhs = " << expr->getOp() << "(lhs, ";
+  } else {
     ss_ << "lhs " << expr->getOp() << "= ";
+  }
+
+  if(hasWeights) {
+    ss_ << "weight * ";
   }
 
   auto argName = denseArgName_;
@@ -407,6 +414,9 @@ void ASTStencilBody::visit(const std::shared_ptr<ast::ReductionOverNeighborExpr>
   }
   // "pop" argName
   denseArgName_ = argName;
+  if(expr->getOp() != "+") {
+    ss_ << ")";
+  }
   ss_ << ";\n";
   ss_ << ASTStencilBody::ReductionSparseIndexVarName(reductionDepth_) << "++;\n";
   ss_ << "return lhs;\n";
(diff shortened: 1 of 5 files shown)


muellch:
launch jenkins

dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
```diff
@@ -376,15 +376,22 @@ void ASTStencilBody::visit(const std::shared_ptr<ast::ReductionOverNeighborExpr>
...
         << ", auto const& weight) mutable {\n";
-    ss_ << "lhs " << expr->getOp() << "= ";
-    ss_ << "weight * ";
   } else {
     ss_ << ", [&, " + ASTStencilBody::ReductionSparseIndexVarName(reductionDepth_) +
                " = int(0)](auto& lhs, auto "
         << ASTStencilBody::ReductionIndexVarName(reductionDepth_ + 1) << ") mutable { ";
+  }
+
+  if(expr->getOp() != "+") {
```
mroethlin, line 385: I don't think this is safe. There are other compound assignments (e.g. `-=`)

dawn/src/dawn/CodeGen/Cuda-ico/ASTStencilBody.cpp
```diff
@@ -240,11 +240,18 @@ void ASTStencilBody::visit(const std::shared_ptr<ast::ReductionOverNeighborExpr>
       << "pidx * " << chainToSparseSizeString(expr->getIterSpace()) << " + nbhIter"
       << "];\n";
   ss_ << "if (nbhIdx == DEVICE_MISSING_VALUE) { continue; }";
-  ss_ << lhs_name << " " << expr->getOp() << "=";
+  if(expr->getOp() != "+") {
```
mroethlin, line 243: c.f. above

mroethlin (changes requested):
also consider other compound assignments, then we're good to go

dawn/src/dawn/CodeGen/CXXNaive-ico/ASTStencilBody.cpp
muellch: I thought at the moment there is only `sum_over`, `max_over` and `min_over`?
In that context `ReductionOverNeighborExpr` it should only be these 3.

I guess I could try to make it more future proof?
Would it be good enough if I excluse `+`, `-`, `*`, `/` and `%`?

mroethlin: `sum_over`, `max_over` and `min_over` are the shortcuts available in dusk. there is also a more general `reduce_over` which takes an Operator as third argument. I'm currently not sure we ever formally specified which operators are allowed for the reduction concept. Your selection seems fine to me

muellch:
launch jenkins

muellch:
launch jenkins

muellch:
launch jenkins

muellch:
Alright, should be good now. After dealing with the arithmetic ops I did some cleanup.

mroethlin (approved):
I know we aren't very consistent with this, but dawn is supposed to use `camelCase` internally (but may emit `snake_case` code). Other than that it's good to go!

muellch:
launch jenkins

PR #1113, opened 2021-02-10
Add set_splitter_index and some bindings

Stagno:
## Technical Description

Ship with dawn the API to set markers' indices (and its binding). Also add missing bindings to serialization utilities (so that can be used in any place within the model code).


mroethlin (approved):
lgtm

Stagno:
launch jenkins

PR #1114, opened 2021-02-10
Fix a Bug in the Unstr. Dim Checker w.r.t. Index Fields

### LINKED ISSUE #1106: Type Checking Bug on Sparse Vertical Index Fields
Currently, the type checking on index fields is too strict. Consider:
```
@stencil
def a_stencil(
    a: Field[Cell, K],
    b: Field[Edge, K],        
    test_field: IndexField[Cell > Edge, K]
):
  with domain.upward as k:
    a = sum_over(Cell > Edge, b[test_field])      
```

this is rejected by the dimensions checker because the types of `b` and `test_field` are different. However, this is clearly a sensible example. The type checking should not enforce that  `b` and `test_field`have the same type, but that both types are valid in the current iteration space. 

**NOTE**: This issue was updated because it was initially thought that sparse index fields are not supported at all.

mroethlin:
## Technical Description

During the integration workshop it was noted that there seems to be a problem with dawn when using sparse index fields. Further investigation revealed that, while sparse index fields are fully supported, this was rather a case of too strict dimensions checking. It was required that `a[b] <=> dim(a) == dim(b)`, which is too strict if the index fields are used in reductions. This PR basically relaxes the dimensions checking for index fields to the same rules as for normal fields in reductions, while leaving the `a[b] <=> dim(a) == dim(b)` for point-wise situations. 

### Resolves / Enhances

Fixes #1106 

### Example

We basically want to enable stencils like the following, which is a stripped down version [of a real ICON stencil](https://github.com/dawn-ico/ICON-DSL/pull/21). 

```
@stencil
def a_stencil(
    a: Field[Cell, K],
    b: Field[Edge, K],        
    test_field: IndexField[Cell > Edge, K]
):
  with domain.upward as k:
    a = sum_over(Cell > Edge, b[test_field])      
```

Note that relaxing the type checker in this way also allows stencils like:

```
@stencil
def another_stencil(
    a: Field[Cell, K],
    b: Field[Edge, K],        
    c: IndexField[Cell, K],
):
  with domain.upward as k:
    a = sum_over(Cell > Edge, b[c])
```

this may seem strange at first but could basically be used to achieve things like "for some cells iterate the edges a level above while for others the edges one level below", which, given the general nature of ICON stencils, doesn't seem too far fetched. 

### Testing

* A new `dawn4py` test using the first example
* Two new `UnstructuredDimensionsChecker` tests for both examples above

### Dependencies

This PR is independent.

### DIFF
--- a/dawn/src/dawn/Validator/UnstructuredDimensionChecker.cpp
+++ b/dawn/src/dawn/Validator/UnstructuredDimensionChecker.cpp
@@ -1,5 +1,6 @@
 #include "UnstructuredDimensionChecker.h"
 #include "dawn/AST/ASTStmt.h"
+#include "dawn/AST/FieldDimension.h"
 #include "dawn/AST/IterationSpace.h"
 #include "dawn/AST/LocationType.h"
 #include "dawn/AST/Offsets.h"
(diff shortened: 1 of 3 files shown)


mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (approved):
Looks good

PR #1123, opened 2021-03-10
Merge Reductions

mroethlin:
## Technical Description

This PR introduces a very simple version of a possible optimization pass, the reduction merging pass. It was not implemented as an actual pass, but the actual merging logic was offloaded to the codegen module. While this approach is certainly not optimal, it avoids the burden of having to redesign the `ReductionOverNeighborExpr` node in the IIR.

The merger was kept simple by not performing a dependency analysis of the statements _between_ two merge candidates. That is, merging is performed if and only if two neighbor reductions appear right after each other.

Consider:
```
with domain.upward as k:
     a = sum_over(Cell > Edge, b)
     c = sum_over(Cell > Edge, d)
```
which is merged, whereas
```
with domain.upward as k:
     a = sum_over(Cell > Edge, b)
     e = 1
     c = sum_over(Cell > Edge, d)
```
is **NOT** merged

### Example
simply pass the newly introduced `--merge-reductions` flag to `dawn-codegen`
```
./dawn-codegen --merge-reductions my_sten.iir ...
```

### Testing

* A few tests with some merging and non merging situations were introduced using the `IIRBuilder`
* Verifies on `ICON-DSL` using `probtest` but not `__DSL_VERIFY` (but neither does current dawn master)

### Dependencies

This PR is independent.


muellch (commented):
Awesome

dawn/src/dawn/CodeGen/Cuda-ico/ReductionMerger.h
```diff
@@ -0,0 +1,270 @@
...
+    //           break;
+    //         }
+    //       }
+    //     }
+    //     mergeGroups.push_back(mergeGroup);
+    //     outerIterator = innerIterator;
+    //   } else {
+    //     outerIterator++;
+    //   }
+    // }
```
muellch, line 239: I have no problem with leaving this here, but is it intentional?

muellch (approved):
One small remark aside, looks good

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #1125, opened 2021-03-19
Fix Serialization 

mroethlin:
## Technical Description

The serialization routines had some problems:

* Calling the destructor the second time a serialization routine is called is not safe. A single stencil may serialize multiple fields in the same iteration
* The use of `rdbuf` caused some problems. See also this [stackoverflow question](https://stackoverflow.com/questions/324711/writing-stringstream-contents-into-ofstream) (later answers)

### Testing

Tested using ICON-DSL

### Dependencies

No dependencies

### DIFF
--- a/dawn/src/driver-includes/to_vtk.cpp
+++ b/dawn/src/driver-includes/to_vtk.cpp
@@ -78,8 +78,8 @@ public:
   }
 
   ~StencilFieldsVtkOutput() {
-    fs << cells_ss.rdbuf();
-    fs << points_ss.rdbuf();
+    fs << cells_ss.str();
+    fs << points_ss.str();
     fs.close();
   }
 
@@ -117,14 +117,15 @@ static std::string formatNaNs(const double value) {
 
 namespace {
 StencilFieldsVtkOutput& getStencilFieldsVtkOutput(int num_k, std::string stencil_name, int iter) {
-  if(stencil_to_output_map.count(std::make_pair(stencil_name, iter - 1))) {
-    stencil_to_output_map.erase(std::make_pair(stencil_name, iter - 1));
-  }
   if(stencil_to_output_map.count(std::make_pair(stencil_name, iter)) == 0) {
     stencil_to_output_map.emplace(std::make_pair(stencil_name, iter),
                                   StencilFieldsVtkOutput(num_k, stencil_name, iter));
   }
   return stencil_to_output_map.at(std::make_pair(stencil_name, iter));
+}
+
+void flushAtIter(std::string stencil_name, int iter) {
+  stencil_to_output_map.erase(std::make_pair(stencil_name, iter - 1));
 }
 
 double* fieldFromGpu(const double* field_gpu, const int size) {
@@ -323,5 +324,9 @@ void serialize_dense_edges(int start_idx, int end_idx, int num_k, int dense_stri
 
   dense_edges_to_vtk(start_idx, end_idx, num_k, dense_stride, field, stencil_name, field_name,
                      iter);
+}
+
+void serialize_flush_iter(const char field_name[50], int iter) {
+  flushAtIter(std::string(field_name), iter);
 }
 }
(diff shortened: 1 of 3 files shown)


dawn/src/driver-includes/to_vtk.cpp
```diff
@@ -117,16 +117,17 @@ static std::string formatNaNs(const double value) {
 
 namespace {
 StencilFieldsVtkOutput& getStencilFieldsVtkOutput(int num_k, std::string stencil_name, int iter) {
-  if(stencil_to_output_map.count(std::make_pair(stencil_name, iter - 1))) {
-    stencil_to_output_map.erase(std::make_pair(stencil_name, iter - 1));
-  }
```
Stagno, line 122: > Calling the destructor the second time a serialization routine is called is not safe. A single stencil may serialize multiple fields in the same iteration

Hmm, I don't see how it's possible that the destructor of the same object gets called more than once, as once the entry of `stencil_to_output_map` with some key is deleted (which entails 1 call to the destructor), the next time the condition (of the if) is going to be evaluated to false, as the key is not going to be present anymore.
Moreover this code destroys the object for the same stencil but for the previous iteration (`iter - 1`), for which all fields have already been serialized.
I think the problem then it's somewhere else, but ....

dawn/src/driver-includes/to_vtk.cpp
```diff
@@ -324,4 +325,8 @@ void serialize_dense_edges(int start_idx, int end_idx, int num_k, int dense_stri
   dense_edges_to_vtk(start_idx, end_idx, num_k, dense_stride, field, stencil_name, field_name,
                      iter);
 }
+
+void serialize_flush_iter(const char field_name[50], int iter) {
+  flushAtIter(std::string(field_name), iter);
+}
```
Stagno, line 331:
```suggestion
void serialize_flush_iter(const char stencil_name[50], int iter) {
  flushAtIter(std::string(stencil_name), iter);
}
```

dawn/src/driver-includes/to_vtk.h
```diff
@@ -40,4 +40,5 @@ void serialize_dense_verts(int start_idx, int end_idx, int num_k, int dense_stri
 void serialize_dense_edges(int start_idx, int end_idx, int num_k, int dense_stride,
                            const double* field, const char stencil_name[50],
                            const char field_name[50], int iter);
+void serialize_flush_iter(const char field_name[50], int iter);
```
Stagno, line 43:
```suggestion
void serialize_flush_iter(const char stencil_name[50], int iter);
```

dawn/src/driver-includes/to_vtk.cpp
```diff
@@ -117,16 +117,17 @@ static std::string formatNaNs(const double value) {
...
-  }
   if(stencil_to_output_map.count(std::make_pair(stencil_name, iter)) == 0) {
     stencil_to_output_map.emplace(std::make_pair(stencil_name, iter),
                                   StencilFieldsVtkOutput(num_k, stencil_name, iter));
   }
   return stencil_to_output_map.at(std::make_pair(stencil_name, iter));
 }
 
+void flushAtIter(std::string stencil_name, int iter) {
+  stencil_to_output_map.erase(std::make_pair(stencil_name, iter - 1));
```
Stagno, line 128: ... this solution is cleaner, however it adds implicitly an assumption that was not there before: that the same stencil (same stencil name string) won't be called more than once in the same time iteration. Therefore I suggest to add this assumption somewhere here: https://docs.google.com/document/d/1ac0Z_8K_2mxU_fsR-9I_ih0KWry34bPOeydD8KyDiHM/edit#heading=h.1uyg0j7e7p8t
```suggestion
  stencil_to_output_map.erase(std::make_pair(stencil_name, iter));
```
Also, why flushing the previous iteration here? I think you want to flush the current one in `flushAtIter()`

Stagno (changes requested):
Small changes requested, see comments.

mroethlin:
launch jenkins

mroethlin:
launch jenkins

Stagno (approved):

PR #1129, opened 2021-04-15
Fix optional threshold

Stagno:
## Technical Description

Switch from ternary pattern to if/else when evaluating the optional threshold argument (details: https://github.com/j3-fortran/fortran_proposals/issues/138#issuecomment-573322935 ). Fixes the access to uninitialized memory problem.

### Testing

Manually tested in the integration with ICON


Stagno:
launch jenkins

mroethlin (approved):
lgtm

PR #1132, opened 2021-05-05
Add rank ID to serialized vtk/csv filenames

Stagno:
## Technical Description

Necessary for multi-node output: each mpi work rank in ICON would write independently to its own file.
Example:
```
mo_nh_diffusion_stencil_19_rank0_theta_v_0.csv
mo_nh_diffusion_stencil_19_rank1_theta_v_0.csv
```

Works also with 1 work rank (current setup), the `rank0_` part would be added in any case.
It doesn't necessarily require the code (on the model side) to set the rank ID, if only one work rank is used (as there's a default of 0).

## Testing

In `icon-dsl`.


Stagno:
launch jenkins

mroethlin (approved):
lgtm

PR #1136, opened 2021-05-14
Use numpy isclose-like verification, allow per field tolerances

muellch:
## Technical Description

Now use `numpy.isclose`-inspired verification method using the following formula:
```
abs(dsl[pidx] - fortran[pidx]) <= (abs_tol + rel_tol * abs(fortran[pidx]))
```
The user now can also provide `abs_tol` and `rel_tol` per OUT/INOUT field and not per kernel. There are default values in place for `abs_tol=0.0` and `rel_tol=1e-12`.

If verification fails, more information about the dsl and reference fields is given now: min, max, average.

### Resolves

The relative error formula below could lead to false positives in verification failure:
```
 double error = 0.;
    if(expected == actual) {
    } else if(fabs(expected) < 1e-6 && fabs(actual) < 1e-6) {
      error = fabs(expected - actual);
    } else {
      error = fabs((expected - actual) / expected);
    }
```
This is no longer in dawn.

### DIFF
--- a/dawn/src/driver-includes/cuda_verify.cpp
+++ b/dawn/src/driver-includes/cuda_verify.cpp
@@ -55,14 +55,16 @@ bool verify_field(const int num_el, const double* dsl, const double* actual, std
   maxAbsErr = thrust::reduce(dev_ptr, dev_ptr + num_el, 0., thrust::maximum<double>());
   gpuErrchk(cudaPeekAtLastError());
 
-  std::cout << "[DSL] " << name << " absolute error: " << std::scientific << maxAbsErr << "\n"
+  std::cout << "[DSL] " << name << " maximum absolute error: " << std::scientific << maxAbsErr
+            << "\n"
             << std::flush;
 
   minAbsErr = thrust::reduce(dev_ptr, dev_ptr + num_el, std::numeric_limits<double>::infinity(),
                              thrust::minimum<double>());
   gpuErrchk(cudaPeekAtLastError());
 
-  std::cout << "[DSL] " << name << " absolute error: " << std::scientific << minAbsErr << "\n"
+  std::cout << "[DSL] " << name << " minimum absolute error: " << std::scientific << minAbsErr
+            << "\n"
             << std::flush;
 
   gpuErrchk(cudaFree(gpu_error));
--- a/dawn/src/driver-includes/cuda_verify.cpp
+++ b/dawn/src/driver-includes/cuda_verify.cpp
@@ -1,0 +1,108 @@
+#include "cuda_verify.hpp"
+
+namespace dawn {
+__global__ void isclose_kernel(const int num_el, const double* __restrict__ dsl,
+                               const double* __restrict__ fortran, bool* __restrict__ verified,
+                               const double rel_tol, const double abs_tol) {
+  unsigned int pidx = blockIdx.x * blockDim.x + threadIdx.x;
+  if(pidx >= num_el) {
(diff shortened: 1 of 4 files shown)


muellch:
launch jenkins

dawn/src/driver-includes/cuda_verify.cpp
```diff
@@ -0,0 +1,108 @@
...
+  gpuErrchk(cudaPeekAtLastError());
+
+  dev_ptr = thrust::device_pointer_cast(gpu_error);
+  absErr = thrust::reduce(dev_ptr, dev_ptr + num_el, 0., thrust::maximum<double>());
+  gpuErrchk(cudaPeekAtLastError());
+
+  std::cout << "[DSL] " << name << " absolute error: " << std::scientific << absErr << "\n"
+            << std::flush;
+  gpuErrchk(cudaFree(gpu_error));
+  gpuErrchk(cudaPeekAtLastError());
```
Stagno, line 50: I like the overview of the dataset printed by the code below.
I'd also give an overview of the errors here, i.e. print always both the max (currently printed) and min relative and absolute errors. Not sure about the avg.
Providing more infos here helps in reducing the number of times we need to download the vtk/csv files and visualize them.

Stagno (changes requested):
A small improvement requested, otherwise good to go.

Stagno:
@muellch also need to fix icondusk-e2e which becomes incompatible with the API changes i think: http://jenkins-mch.cscs.ch/job/icondusk-e2e_PR/118/slave=tsa/console

muellch:
launch jenkins

dawn/src/driver-includes/cuda_verify.cpp
```diff
@@ -31,21 +32,39 @@ bool verify_field(const int num_el, const double* dsl, const double* actual, std
...
   compare_kernel<AbsErrTag><<<dG, dB>>>(num_el, dsl, actual, gpu_error);
   gpuErrchk(cudaPeekAtLastError());
 
   dev_ptr = thrust::device_pointer_cast(gpu_error);
-  absErr = thrust::reduce(dev_ptr, dev_ptr + num_el, 0., thrust::maximum<double>());
+  maxAbsErr = thrust::reduce(dev_ptr, dev_ptr + num_el, 0., thrust::maximum<double>());
   gpuErrchk(cudaPeekAtLastError());
 
-  std::cout << "[DSL] " << name << " absolute error: " << std::scientific << absErr << "\n"
+  std::cout << "[DSL] " << name << " absolute error: " << std::scientific << maxAbsErr << "\n"
```
Stagno, line 58: missing "maximum"

dawn/src/driver-includes/cuda_verify.cpp
```diff
@@ -31,21 +32,39 @@ bool verify_field(const int num_el, const double* dsl, const double* actual, std
...
             << std::flush;
+
+  minAbsErr = thrust::reduce(dev_ptr, dev_ptr + num_el, std::numeric_limits<double>::infinity(),
+                             thrust::minimum<double>());
+  gpuErrchk(cudaPeekAtLastError());
+
+  std::cout << "[DSL] " << name << " absolute error: " << std::scientific << minAbsErr << "\n"
```
Stagno, line 65: missing "minimum"

Stagno (changes requested):

muellch:
launch jenkins

PR #1137, opened 2021-05-19
Add switch for Atlas Compatibility

mroethlin:
## Technical Description

In `icondusk-e2e` we test on atlas meshes on the full domain. This means that neighborhood entries may be missing even without the pentagon / hexagon issue (at the border). This PR introduces a switch to emit a guard against such neighborhoods for _all_ chains. 

### Example

`./dawn-codegen --atlas-compatible test_reduce.iir -b cuda-ico -o out.cpp`

### Testing

Tested using `icon-dusk` 

### Dependencies

This PR is independent

### DIFF
--- a/dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
+++ b/dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
@@ -155,7 +155,7 @@ run(const std::map<std::string, std::shared_ptr<iir::StencilInstantiation>>&
       options.OutputCHeader == "" ? std::nullopt : std::make_optional(options.OutputCHeader),
       options.OutputFortranInterface == "" ? std::nullopt
                                            : std::make_optional(options.OutputFortranInterface),
-      options.MergeReductions);
+      options.MergeReductions, options.AtlasCompatible);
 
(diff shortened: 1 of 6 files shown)


dawn/src/dawn/CodeGen/Cuda-ico/CudaIcoCodeGen.cpp
```diff
@@ -1204,6 +1204,7 @@ void CudaIcoCodeGen::generateAllCudaKernels(
     std::stringstream& ssSW,
     const std::shared_ptr<iir::StencilInstantiation>& stencilInstantiation) {
   ASTStencilBody stencilBodyCXXVisitor(stencilInstantiation->getMetaData());
+  stencilBodyCXXVisitor.setGenAtlasCompatCode(codeGenOptions_.AtlasCompatible);
```
cosunae, line 1207: Since we are always setting this, how about adding it to the ctr, instead of having to call a setter ? We might have multiple instances of a ASTStencilBody, and given the compiler flag is like a global option, we might end up with inconsistencies if some are instantiated and call the setter but other dont

cosunae (approved):

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #1140, opened 2021-06-03
Remove Cuda Timers from CudaIco Codegen

mroethlin:
## Technical Description

The cuda timers (using `cudaEventRecord`) interspersed into the generated code cause large overhead. This does not affect timings when timing individual kernels, but skews the wall clock time when looking at the runtime of e.g. a complete submodule. Since we never consume those timings, this PR removes them completely, that is, also from the FORTRAN interface.

### Testing

Nothing much to test, really. The runtime was a return argument, but said argument was always ignored, hence `icon-dsl` builds and runs just fine with these changes (tested).

### Dependencies

This PR is independent.


mroethlin:
launch jenkins

mroethlin:
launch jenkins

muellch (approved):
Looks good!

PR #1141, opened 2021-06-09
Vtk precision fixes

BenWeber42:
## Technical Description

Serializing double precision fields to vtk was lossy due to a number of factors. This and related issues are fixed here:

- Changed vtk data types from `float` to `double`.
- Added `setprecision` to vtk functions
- Fixed `formatNaN` to respect `setprecision` (for more see below)
- Fixed types and `max_digits10` for `setprecision` (for more see below)

#### Regarding `formatNaN`

I considered various different options. Before, the function `formatNaNs` returned a string and the function didn't respect `setprecision`. I wanted that `formatNaNs` respects `setprecision` while also formating NaNs correctly (as `"nan"`). I could have also used a string stream and then set the precision each time before formatting the double into that string stream. However, this means the function won't respect `setprecision` of other streams and the precision isn't really customizable for a user.
Using this wrapper type, I can customize the formatting of NaNs, but also ensure that `setprecision` of the stream is respected.

I'm unsure whether the struct should still be named `formatNaNs`. I considered various alternatives, but couldn't come up with anything that I think is better. So I thought I'll stick with the simple solution for now. Additionally, the struct and the overloading of the `std::ostream <<` operator are only local to the `.cpp` file. So even though the solution is a bit hacky, it's only confined to this file. If in the future we want to use this trick elsewhere, we should probably consider better ways to name the struct and overload the operator (or a different implementation alltogether). However, because it's confined only to this file and it's a simple implementation that works, I thought it's fine like this.

#### Regarding `std::setprecision(std::numeric_limits<double>::max_digits10);`

`long double` is wrong in those cases, because all fields are statically typed as `double`.

I believe `max_digits10` is the correct value we're looking for here. We've had this already before here: https://github.com/MeteoSwiss-APN/dawn/pull/992

There's also:

> `max_digits10` is the number of decimal digits needed to guarantee correct float → text → float round-trip.

from here: https://stackoverflow.com/a/22458961/12958632

### Resolves / Enhances

Resolves lossy serialization of double precision fields to vtk.

### Testing

Tested manually with ICON and ParaView.

### DIFF
--- a/dawn/src/driver-includes/to_vtk.cpp
+++ b/dawn/src/driver-includes/to_vtk.cpp
@@ -52,6 +52,8 @@ public:
 
     fs << "POINTS " << num_verts * num_k << " double\n";
 
+    fs << std::setprecision(std::numeric_limits<double>::max_digits10);
+
     for(int k = 0; k < num_k; k++) {
       for(int nodeIter = 0; nodeIter < num_verts; nodeIter++) {
         double x = vlat[nodeIter];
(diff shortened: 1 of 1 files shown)


Stagno (approved):
Looks good to me. Thx for the fix!

BenWeber42:
launch jenkins

BenWeber42:
Thanks, waiting for CI of `dsl/icon-cscs` to pass with these changes as well...

PR #1147, opened 2021-07-16
Unused api fields bugfix

BenWeber42:
## Technical Description

It fixes a bug where broken code gets generated if an API field is unused.
The code generated wrapper function will only pass used API fields to the run call.
If an API field is unused, it will occur in the signature of the run function, but it will be missing when the function is called in the wrapper function. Thus the call is invalid and the generated code is broken.

### Testing

Tested manually with ICON.


BenWeber42:
launch jenkins

muellch (approved):
lgtm

PR #1150, opened 2021-08-04
Make getShape convenience method in atlas_interface consistent

mroethlin:
## Technical Description

Joerg Behrens reported the following:

in the file `dawn/src/interface/atlas_interface.hpp` you define the class
`SparseDimension`. There, the access `operator()` seems to expect the
following order of dimensions `(elem_dim, sparse_dim, level_dim)`, e.g.:

```
T const& operator()(int elem_idx, int sparse_dim_idx, int level)
```

The `getShape()` method, however, returns the field shape in the class
internal order of dimensions which seems to be (elem_dim, level_dim,
sparse_dim).

Therefore the field shape given by `getShape()` seems to be inconsistent
with the field access or, in other words: The shape does not give the
access bounds of the access operator as would be expected (I am only
guessing the semantics from the method name - which triggers my Fortran
background).

In my opinion a consistent `getShape` method would return something like

```
std::make_tuple(
   sparse_dimension_.shape(0),
   sparse_dimension_.shape(2),
   sparse_dimension_.shape(1));
```

This PR implements the suggested `getShape` method.

### Testing

This method is dead code in the dawn repo, so this remains untested!

### DIFF
--- a/dawn/src/interface/atlas_interface.hpp
+++ b/dawn/src/interface/atlas_interface.hpp
@@ -156,8 +156,8 @@ public:
     return sparse_dimension_.shape(0) * sparse_dimension_.shape(1) * sparse_dimension_.shape(2);
   }
   std::tuple<size_t, size_t, size_t> getShape() const {
-    return std::make_tuple(sparse_dimension_.shape(0), sparse_dimension_.shape(1),
-                           sparse_dimension_.shape(2));
+    return std::make_tuple(sparse_dimension_.shape(0), sparse_dimension_.shape(2),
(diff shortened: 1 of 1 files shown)


BenWeber42 (approved):
I think the other option is to change the access operator. But to be honest, I'm not sure if there is a canonical order for these three dimensions. As long as there isn't one, it's not really clear what should be changed how. So this change makes the two methods consistent with each other and thus I think is a valid improvement.

mroethlin:
launch jenkins

mroethlin:
launch jenkins

PR #1156, opened 2021-10-14
Fix Codegen for Sparse Fields in Nested Reductions

mroethlin:
## Technical Description

There was a spurious error emitted in the codegen, complaining that a sparse field was not contained in a reduction, even though it was. This can for example be triggered by 

```
@stencil
def codegen_bug(inF: Field[Cell, K], coeff: Field[Edge > Vertex, K], outF: Field[Edge, K]):
    tempF: Field[Vertex, K]
    with domain.upward.across[lb+2:halo-1]:       
        outF = sum_over(Edge > Vertex, coeff*sum_over(Vertex > Cell, inF))
```

The problem was that the current state of the codegen stack was not properly communicated "downwards". This PR fixes that issue.

**NOTE** Due to dawn becoming EOL, This is a quick and dirty fix without test

### Example

c.f. above

### Testing

**None**

### Dependencies

This PR is independent.


muellch (approved):
Sorry, did not see we want to merge this.
Looks good!

PR #1158, opened 2022-01-11
Fix Check Guarding Against Empty Region in Codegen

mroethlin:
## Technical Description

The guards in the `cuda-ico` codegen guarding against stencils with empty domains caused problems; they were formulated like so:

```
if (hsizeXXX == 0) {
  return;
}
```

causing subsequent stencils to not be run if `hsizeXXX` was indeed zero. This PR addresses this by conceptually inverting the guard, i.e. only running the stencil if `hsizeXXX == 0`

### Testing

Tested using `icon-cscs`


mroethlin:
launch jenkins

mroethlin:
launch jenkins

muellch (approved):

PR #1162, opened 2022-05-19
Make Verification Functions Run in Same Stream as Kernel

mroethlin:
## Technical Description

The verification functions in our `cuda_verify` files did not respect the stream being set in the setup function of the stencil. This was safe since the verification functions call a `cudaDeviceSynchronize()` (both explicitly as well as implicitly using `cudaMalloc`). However, it is still good practice to run everything in the stream that the user set.


mroethlin:
launch jenkins

muellch:
Looks good!
Did you double check with nsys?

muellch (approved):

PR #1163, opened 2022-06-02
Set protobuf version to be <3.20.0

muellch:
## Technical Description

It seems google broke backwards compatibility with protobuf ~3.21, so we put a maximum version of <3.20.0 to protobuf since dawn is very near to end-of-life


muellch:
launch jenkins

BenWeber42 (approved):

PR #1165, opened 2022-07-28
Json serialisation jenkins test

muellch:
Test what is wrong with jenkins


muellch:
launch jenkins

muellch:
launch jenkins

dawn/src/driver-includes/to_json.cpp
```diff
@@ -0,0 +1,77 @@
...
+                    }
+            };
+    return j;
+}
+
+std::string getEnvVar( std::string const & key )
+{
+    char * val = getenv( key.c_str() );
+    return val == NULL ? std::string("") : std::string(val);
+}
```
muellch, line 77: Use `nullptr` instead of `NULL`. If you are interested why: https://www.geeksforgeeks.org/understanding-nullptr-c/
Instead of `std::string("")` the default constructor of std::string also results in an empty string. So you can use `std::string()`.

dawn/src/driver-includes/to_json.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+using namespace dawn;
+
+MetricsSerialiser::MetricsSerialiser(VerificationMetrics metricsStruct, std::string metricsPath, std::string stencilName,
+                                     std::string fieldIdentifier) {
+        fieldId = fieldIdentifier;
+        path = metricsPath;
+        stencil = stencilName;
+        newJsonMetrics = generateJsonFromStruct(metricsStruct);
+    }
```
muellch, line 11: In C++ we use initializer lists for contructors. In this case this would look like this:
```
MetricsSerialiser::MetricsSerialiser(VerificationMetrics metricsStruct, std::string metricsPath, std::string stencilName,
                                     std::string fieldIdentifier) : fieldId(fieldIdentifier), path(metricsPath), stencil(stencilName), 
                                     newJsonMetrics(generateJsonFromStruct(metricsStruct))
                                     {}
```

muellch (commented):
Some of the new files you are submitting (for example `to_json.cpp` and `to_json.hpp` have indentation of 8 spaces. I think default in dawn is 2.

You can just format the code with clang-format, either in the command line or in your IDE.

There is a `.clang-format` file in the dawn root folder, which encodes the correct formatting for this project.

PR #1167, opened 2022-09-19
Template Serialization and Verification Functions 

mroethlin:
## Technical Description

We also need to verify and serialize int fields now, so let's template the serialization and verification functions


dawn/src/driver-includes/cuda_verify.hpp
```diff
@@ -32,31 +33,166 @@ struct compute_error<AbsErrTag> {
...
-template <>
-struct compute_error<RelErrTag> {
-  static double __device__ impl(const double expected, const double actual) {
+template <typename FieldType>
+struct compute_error<RelErrTag, FieldType> {
+  static double __device__ impl(const FieldType expected, const FieldType actual) {
     double error = 0.;
     if(expected != actual) {
       error = fabs((expected - actual) / expected);
     }
```
muellch, line 42: I think `(expected - actual)/expected` will not yield the desired result if `expected` and `actual` are both integers. In that case integer division applies, and they are only later casted to float because the result is used as an input to `fabs`.

One workaround would be
```
error = fabs(((double)expected - actual) / expected);
```

Maybe there is something more elegant ...

muellch (changes requested):
In general there probably should be two concepts of `is_close` for floating comparison and `is_equal` for boolean and integer comparison. And one should only be allowed to use `is_close` with floating point types and `is_equal` with boolean, integer types.

These might already exist in the std library.

So the clean solution needs a bit more work. For now we can merge this and shape a task to clean up the whole dawn C++ library the generated backend code depends on.

PR #1168, opened 2022-11-11
Fixed multinode verification, fixed hardcoded nproma

muellch:
* For multiple nodes multiple verification files are created. Each process writes into a separate file. All stencils get handed a pointer to a shared object at setup time. At the end of an ICON run that object gets dumped into a file.
* Nproma is no longer hard coded but passed to the stencils during setup time. Specifically `Nproma` is now part of the `GlobalGpuTriMesh` struct.


muellch:
launch jenkins

muellch:
launch jenkins

muellch:
launch jenkins

samkellerhals (approved):
Looks good!

